PR #44, opened 2016-05-25
Update lightningdesignsystem.json

kaelig:
Thanks for setting this up for us! Here are a few tweaks that should make search on www.lightningdesignsystem.com slightly more relevant (if I understand well how it works!).


maxiloc:
I just deployed. It will be available after the crawling.
Thanks for contributing. Btw, congrats, your are the first contributor of this repo 🎉

kaelig:
Merci for the prompt merge, @maxiloc!

PR #45, opened 2016-05-25
Update Infinit configuration.

cadeuh:
Changed the parsed URLs and put some rankings on it.

### DIFF
--- a/infinit.json
+++ b/infinit.json
@@ -1,8 +1,38 @@
 {
   "index_name": "infinit",
   "start_urls": [
-    "https://infinit.sh/documentation/",
-    "https://infinit.sh/documentation/overview"
+    {
+      "url": "https://infinit.sh/documentation/reference",
+      "page_rank": 1
+    },
+    {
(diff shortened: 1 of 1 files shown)


infinit.json
```diff
@@ -1,8 +1,38 @@
...
+    {
+      "url": "https://infinit.sh/documentation/status",
+      "page_rank": -2
+    },
+    {
+      "url": "https://infinit.sh/documentation/changelog",
+      "page_rank": -2
+    },
+    {
+      "url": "https://infinit.sh/documentation/.*",
```
redox: I think you need to put that rule last; otherwise it will bypass the next one. Can you guys confirm @maxiloc @ElPicador ?

maxiloc: Yes you just need to switch the last 2. Because we use the page_rank of the first matching start_url

maxiloc:
I updated the configuration manually, so I am closing this. Thanks for contributing

cadeuh:
Thanks!

PR #48, opened 2016-05-26
Update meteor search to search code blocks too

tmeasday:
Ala https://github.com/algolia/docsearch-configs/commit/777967a67dca7dfa62cd96c13086bf2abe04a7a9

Not sure if I need to change `nb_hits` also..


maxiloc:
It will be updated automatically 😄. I am deploying right now it should be available in a few minutes

PR #52, opened 2016-06-24
Update lightningdesignsystem.json

kaelig:
Our information architecture was just updated so I'm putting those changes in here.

Also experimenting with tweaking the lvl1,2,3... Not sure how they work but it seems we often have doubles because of the current config. So we'll see.


ElPicador:
LGTM

PR #53, opened 2016-06-27
Update config for adonisjs.com

thetutlage:
I have updated the website and it's DOM structure has changed. This new config file contains the correct configuration.

### DIFF
--- a/adonisjs.json
+++ b/adonisjs.json
@@ -2,24 +2,6 @@
   "index_name": "adonisjs",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


adonisjs.json
```diff
@@ -2,24 +2,6 @@
   "index_name": "adonisjs",
   "start_urls": [
     {
-      "url": "http://legacy.adonisjs.com/docs/(?P<version>.*?)/installation",
```
ElPicador: Just to be sure, you don't want to index version 1 & 2 anymore?

thetutlage: Yes, those are legacy versions.

PR #59, opened 2016-07-26
Update scraper config for Pydio Documentation.

cdujeu:
Updating level0 selectors and their default value to reflect Pydio doc.


ElPicador:
LGTM

PR #61, opened 2016-08-01
Tweak config for lightningdesignsystem.com

kaelig:
I'm starting to play around with crawler-specific attributes, let's see what happens :)

### DIFF
--- a/configs/lightningdesignsystem.json
+++ b/configs/lightningdesignsystem.json
@@ -53,7 +53,7 @@
     "[data-docsearch-ignore]"
   ],
   "selectors": {
-    "lvl0": "main h1",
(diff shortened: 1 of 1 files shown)


maxiloc:
Scrappy does not handle this syntax for css selector [data-docsearch-ignore], the easiest would be to add a class instead

maxiloc:
I am going to revert that in the meantime

kaelig:
Oh, thanks!

Is this something you plan on supporting?

maxiloc:
This is tricky as this not handle by our code logic but scrappy and lxml which our crawler is built on top of. In the short term I don't think this is something we can handle.

kaelig:
Ok that makes sense, I'll just switch to class names :)

PR #64, opened 2016-08-04
Fix symphony

ElPicador:


maxiloc:
It's better but it doesn't fix their facet implem. I need to implement a new feature for that

PR #65, opened 2016-08-18
Pantheon configuration tweaks

rachelwhitton:
- Exclude selector for dropdown-menu class (Do not index Contributor CTA / dropdown menu)
- Add selector for alert class (notes)
- Set js_render: true


ElPicador:
LGTM

PR #74, opened 2016-09-19
Config changes for the lightning design system

kaelig:
Thanks to @maxiloc and @Shipow for their help!

### DIFF
--- a/configs/lightningdesignsystem.json
+++ b/configs/lightningdesignsystem.json
@@ -44,6 +44,15 @@
     {
       "url": "https://www.lightningdesignsystem.com/",
       "page_rank": 1
+    },
+    {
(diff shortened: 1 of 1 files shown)


maxiloc:
👍 Deploying now

kaelig:
@maxiloc It doesn't look like class names are being fully picked up by the crawler.

Steps to reproduce:

1. Go to https://www.lightningdesignsystem.com/
2. Search for `slds-image__action`

Expected: results highlight this particular term
Actual: other results are surfaced.

Any idea how we could get those class names indexed correctly?

maxiloc:
`separatorsToIndex` need to be inside `custom_settings`, I missed it when merging. This commit should fix it 5658386.

kaelig:
Thank you so much!

PR #77, opened 2016-09-20
Index class names from the table overview

kaelig:
For some reason the crawler doesn't seem to index contents of overview tables (e.g. in https://www.lightningdesignsystem.com/components/activity-timeline/#overview).

This should fix it.


maxiloc:
Just deployed it, should be reindexed in a few minutes

PR #78, opened 2016-09-20
Lightning Design System:

kaelig:
Stop ignoring the overview, hoping it will improve usability of class name searching.

(class names are now indexed but unfortunately clicking on the result redirects to a component flavor instead of the component overview)

Sorry for the many edits today… this is a lot of trial and error but we're getting there :)

### DIFF
--- a/configs/lightningdesignsystem.json
+++ b/configs/lightningdesignsystem.json
@@ -61,7 +61,6 @@
   "stop_urls": [],
   "selectors_exclude": [
     "span.slds-badge",
-    "#overview",
     ".docsearch-ignore"
   ],
   "selectors": {


maxiloc:
No problem 😄, Deploying now

kaelig:
@maxiloc not sure why this is not working. If you search for `slds-image__actions` there are no results returned. This class name can be found in the table at the bottom of this page: https://www.lightningdesignsystem.com/components/images/#overview

Any idea what this could be?

kaelig:
ping @maxiloc

maxiloc:
Sorry for the delay. It should be good now

kaelig:
Thanks a lot! It does work much better now :)

PR #80, opened 2016-09-29
Stop ignoring component overview headings

kaelig:
Not sure this will have a good effect on results but I'm hoping it will improve the experience when searching for class names.


maxiloc:
Let's try 😄

kaelig:
Much better now! Thanks @maxiloc!

PR #87, opened 2016-10-21
Update Pydio documentation

cdujeu:
Admin doc moves from v6-enterprise to v7-enterprise


ElPicador:
I'll relaunch the indexation right now

PR #89, opened 2016-10-26
update aframe config

ngokevin:
https://aframe.io/docs/

- No need to specify links to old `/guide/`s
- Index tables

### DIFF
--- a/configs/aframe.json
+++ b/configs/aframe.json
@@ -11,9 +11,7 @@
           "master"
         ]
       }
-    },
-    "https://aframe.io/docs/0.2.0/guide/",
-    "https://aframe.io/docs/0.1.0/guide/"
+    }
   ],
   "stop_urls": [],
   "selectors": {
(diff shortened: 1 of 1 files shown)


configs/aframe.json
```diff
@@ -11,9 +11,7 @@
           "master"
         ]
       }
-    },
-    "https://aframe.io/docs/0.2.0/guide/",
-    "https://aframe.io/docs/0.1.0/guide/"
+    }
   ],
```
maxiloc: We need those, because https://aframe.io/docs/0.2.0/ is redirecting to a 404

maxiloc: so nothing is crawled for those version

ngokevin: Ah, I will make sure we have redirects then.

ngokevin: I've deployed the redirects. They should work now. Thanks for the catch!

maxiloc:
Thanks deploying now

PR #90, opened 2016-10-27
Blog posts should be indexed

bcremer:
A lot of you blog posts contain important information for developers. In fact a lot of our documentation starts out as a blog post and is split up and moved into "real" documentation later on.


maxiloc:
Thanks. Deploying now

PR #104, opened 2017-01-10
fix lvl0 selector for batch.com

mouton-rebelle:
Just updated https://batch.com/doc to add lvl0 class to the right elements.

related to #HS:302951754-20561#


ElPicador:
Thanks, I'll deploy it right now

PR #107, opened 2017-01-13
Do not index Buck's permalink markers

sdwilsh:
If you search for `deps` on buckbuild.com, you'll see all of the `#` in the headings for the permalink markers that are visible on hover.  I believe this will remove them.


sdwilsh:
cc @JoelMarcey

maxiloc:
@sdwilsh Thanks. Deploying that right now, it should be live in a few minutes

PR #110, opened 2017-01-19
tags: updated `latest` to `ft` (fast track)

woodyrew:
Also
- Removed unnecessary/duplicated start_urls
- Added `pre` and `table` to


ElPicador:
I'll deploy it right now

PR #111, opened 2017-01-26
Remove old Ember config

acorncom:
We're not using this config on the Ember guides/website any more (we've switched over to https://github.com/algolia/docsearch-configs/blob/master/configs/emberjs_versions.json instead).  Thought it might be nice to clean up your index a bit.

Thanks for all you guys do for us here!


maxiloc:
Thanks for thinking about that and taking the time to submit a PR.
Much apreciated

PR #112, opened 2017-01-27
Updated project url

laffan:
We've changed our repo name!


maxiloc:
Thanks. Deploying that right now

PR #115, opened 2017-02-05
modify url

waruqi:


maxiloc:
@waruqi Thanks. I did a small modifs on the start urls.
Deploying right now. Should be live in a few minutes.

waruqi:
Thanks! 👍

PR #116, opened 2017-02-07
fix english document problem

waruqi:
The english document can not be searched, but the chinese document is ok.

So I attempt to remove the repeat tag to fix this issue.

### DIFF
--- a/configs/xmake.json
+++ b/configs/xmake.json
@@ -8,10 +8,6 @@
     {
       "url": "http://xmake.io/#/home",
       "tags": "en"
-    },
-    {
-      "url": "http://xmake.io/#/",
-      "tags": "en"
     }
   ],
   "stop_content": [


maxiloc:
@waruqi That will not work as expected. Can you give me a bit more details about the issue ?
We need http://xmake.io/#/ to fetch all pages that are not under the http://xmake.io/#/zh

waruqi:
I added some [script](https://github.com/tboox/xmake/blob/master/docs/config.js) to my doc site.

```js
plugins: [
    docsearch({
      apiKey: 'fbba61eefc60a833d8caf1fa72bd8ef8',
      indexName: 'xmake',
      tags: ['en', 'zh']
    })
  ]
```

This script refers to the configurate of [docute/config.js](https://github.com/egoist/docute/blob/master/docs/config.js), because I use docute to build my doc site.

But it's [docsearch/docute.json](https://github.com/algolia/docsearch-configs/blob/master/configs/docute.json) is ok and I can search it's [english documents](https://docute.js.org/#/home).

maxiloc:
@waruqi Sorry, I still not sure to understand the issue.
Can you give me an example:
- what search query you're doing
- which result you have
- which result you would expect

waruqi:
For example:

* what search query you're doing

I attempt to search the keyword `Installation` 

```
http://xmake.io/#/home?keyword=Installation
``` 
* which result you have
```
No search results :(
```
* which result you would expect

### Installation

#### Windows

* Download xmake installer (Releases)
* Run xmake-[version].exe

#### MacOS
...

maxiloc:
On your front-end you are filtering with `tags:english` but the tag is `en` so you need to replace it by `tags:en`

waruqi:
Oh, thanks!

PR #119, opened 2017-02-14
Index new version 0.5.0 for A-Frame

ngokevin:


ElPicador:
Thanks, will deploy right now

PR #122, opened 2017-02-15
Update Waylay crawler URL

gillesdemey:
We've decided to host the documentation on a different subdomain, unfortunately the CNAME record doesn't cooperate with our SNI proxy.


maxiloc:
Thanks deploying right now

PR #123, opened 2017-02-21
[serenytics.com] Update tags in config

achauve:
Developer documentation is in English but is not translated, so use both en and fr tags for it.


achauve:
By the way I'm not sure about the syntax for multiple tags and I didn't find it in the doc, so I guessed... :)

PR #124, opened 2017-02-22
[docs.rightscale.com] Disable https

crunis:
We currently don't support https in our documentation, so we need the results to be http.


maxiloc:
Thanks, let me deploy that right now

PR #128, opened 2017-03-01
Update denali js config

davewasmer:
Fix lvl0 selector. This might take a couple tries as I figure out how the `lvlN` selectors correspond to various parts on the search dropdown UI.


ElPicador:
Thanks, I'll deploy it right now

PR #131, opened 2017-03-01
Add more synonyms

mikaa123:


Haroenv:
You'll also need es2015, es2016, es2017 etc

mikaa123:
good point @Haroenv

PR #133, opened 2017-03-02
Change of URL for Hugo docs concept website.

rdwatters:
I changed the URL of this concept site for better integration with Netlify's subdomain/git integration. As a head's up, if the Hugo team digs the finished product, this would live as part of the docs on their official site (gohugo.io; already using Algolia) and I would therefore cancel this service altogether. Thanks so much --this is incredibly generous. Algolia rocks!


ElPicador:
I'll deploy it right now

PR #138, opened 2017-03-27
Congif update to omit objects from index

stevenbennitt:
I have updated our config file to omit certain pages and selectors from being indexed.


maxiloc:
Deploying right now.

PR #140, opened 2017-03-28
Update search rankings weighting for documention.chartiq.com

RMcNeely:


ElPicador:
I'll deploy right now

PR #145, opened 2017-04-07
Config Fine Tune

stevenbennitt:


configs/bigcommerce.json
```diff
@@ -1,37 +1,625 @@
 {
-  "index_name": "bigcommerce",
+  "index_name": "TESTING_INDEX",
```
ElPicador: Could you put the previous `index_name`?

configs/bigcommerce.json
```diff
@@ -1,37 +1,625 @@
...
   "start_urls": [
-    "https://support.bigcommerce.com/",
-    "https://support.bigcommerce.com/guides",
-    "https://support.bigcommerce.com/university",
-    "https://support.bigcommerce.com/documentation"
-  ],
+	{"url":"https://support.bigcommerce.com/articles/Learning/What-is-SEO","selectors_key": "guides"},
```
ElPicador: Seems big and complicated, can't you do something simpler?
It seems each `selectors_key` is associated with a specific URL pattern.

configs/bigcommerce.json
```diff
@@ -1,37 +1,625 @@
...
+	{"url":"https://support.bigcommerce.com/articles/Public/How-do-my-customers-reset-their-password-for-their-account-on-my-store","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public/How-do-I-change-the-Unable-to-Ship-to-this-Location-message-on-checkout","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public/How-do-I-auto-return-my-customers-to-my-store-after-they-checkout-with-PayPal","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public/How-do-I-control-the-display-order-of-the-products-on-my-category-and-brand-pages","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public/How-Bigcommerce-Calculates-Trailing-12-Month-Sales-Orders-to-Determine-Your-Pricing","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public/How-do-I-access-and-install-the-new-single-click-apps-within-my-Bigcommerce-store-control-panel","selectors_key": "docs"},
+	{"url":"https://support.bigcommerce.com/articles/Public
```
ElPicador: could you indent it correctly?

ElPicador: it's `js_wait`

configs/bigcommerce.json
stevenbennitt: I agree and I would prefer not to do it this way but based on the way the site is set up that craw config is not getting every URL in the site.

configs/bigcommerce.json
stevenbennitt: How critical is that? It would take a while to do is why I'm asking.

stevenbennitt: k, since i've had it wrong Im going to remove it.

ElPicador: can't you use a linter? like `jq`? 
It allows to have better diff when modifying the file

configs/bigcommerce.json
ElPicador: Have you tried it locally? Because I'm not sure that many start_urls will work

stevenbennitt: I have, it should find the URLs on my local copy it crawls the urls without issue.

ElPicador:
I'm sorry but it's still not indented, just use a linter even an online one like http://jsonlint.com/

maxiloc:
Let's close this. Could you send us via email urls that are not crawled and we'll investigate

PR #146, opened 2017-04-07
Scrape the Ember api docs using SSL

acorncom:
We recently updated to requiring SSL for all our pages.  API search stopped working sometime after that, which seems like it might be related 😀

### DIFF
--- a/configs/emberjs_versions.json
+++ b/configs/emberjs_versions.json
@@ -11,7 +11,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


ElPicador:
I'll deploy it right now

mschinis:
thanks @ElPicador. When should we expect the api docs to be reindexed?

ElPicador:
It should be done

PR #150, opened 2017-04-20
Changed datastructures.js.json

deadcoder0904:


ElPicador:
Thanks, I'll deploy it right away

PR #151, opened 2017-04-25
Update search weightings

RMcNeely:
When we last updated our search rankings we ended up with the entire tutorial section in our search results. I'm assuming this is because we weighted the class name and not the individual headers, will this change make it so that our entire tutorial isn't returned as the search result?


maxiloc:
This would fix the issue but not perfectly, I am going to update the config and we'll iterate on it

PR #154, opened 2017-05-15
Update admin doc for Pydio 8

cdujeu:
Hi, we are upgrading our docs for next release. Thx.

### DIFF
--- a/configs/pydio.json
+++ b/configs/pydio.json
@@ -2,7 +2,7 @@
   "index_name": "pydio",
   "start_urls": [
     {
-      "url": "https://pydio.com/en/docs/v7-enterprise",
+      "url": "https://pydio.com/en/docs/v8",
       "tags": [
         "administration-guide"
       ],


maxiloc:
Thanks. Deploying right now.

cdujeu:
salut @maxiloc, small questions : is it possible to know when the indexation is run, and is there a log available somewhere? I miss some results in my search but i'm not sure whether it's a config issue or because indexation is not (yet) up-to-date.
thx in advance

maxiloc:
We have the log on our side, but we don't have a way for you to access it.
Last run was 15 May 2017 11:10:37 GMT

cdujeu:
can you quickly check if you see many pages starting with pydio.com/en/docs/v8/[....] properly parsed?

maxiloc:
I re-run the crawler just to be sure, I see around 40 page which from v8

cdujeu:
.. and i have better results now! Thank you so much :-)

PR #158, opened 2017-06-07
Remove Cypress doc scraper config

bahmutov:
We at Cypress.io are now scraping the docs ourselves, and this config is now unnecessary


ElPicador:
Thanks for letting us know!

PR #159, opened 2017-06-07
update config for a-frame 0.6.0

ngokevin:
This should be released tomorrow morning. I don't know if there's harm in updating the config a day earlier, hopefully scraper can handle 404s.


ElPicador:
I'll deploy it right now

PR #160, opened 2017-06-11
Update uniwebview.json

onevcat:
Domain name changed. lvl5 selector updated.


maxiloc:
Thanks deploying that now

PR #161, opened 2017-06-12
Do not crawl archived path

onevcat:


maxiloc:
It's currently deploying 😄

onevcat:
So quick! Thanks!

PR #164, opened 2017-06-16
Change bucklescript link

chenglou:
BuckleScript recently moved and the links are broken for some reasons. This fixes it

cc @bobzhang


ElPicador:
Deploying right now

PR #167, opened 2017-06-22
update weex config json

Tancy:
Increase the multi-language configuration and the China mirror document site configuration

### DIFF
--- a/configs/weex.json
+++ b/configs/weex.json
@@ -2,13 +2,13 @@
   "index_name": "weex",
   "start_urls": [
     {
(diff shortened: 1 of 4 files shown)


ElPicador:
I do not see any difference between `weex.apache.org` & `weex-project.io`. 
Do you need to index both?

Tancy:
Weex.apache.org in some of China's network case will be unable to visit, if
not a number of index search results link is pointing to weex.apache.org
Rémy-Christophe Schermesser <notifications@github.com>于2017年6月23日
周五上午1:31写道：

PR #171, opened 2017-06-23
facetwp.json - does the "?" before URL variables need to be escaped?

mgibbs189:


ElPicador:
Careful it's not a valid json. 
And no you do not need to escape the `?`

mgibbs189:
@ElPicador Hmm, do you have **any** idea why no content in the `/blog/` section is getting indexed? I've tried everything I can think of so far `¯\_(ツ)_/¯`

PR #172, opened 2017-06-29
Update sequelizejs.json

felixfbecker:
I just tried to integrate the search locally. It worked great for the markdown articles/guides, but did not suggest anything from the [API reference](http://docs.sequelizejs.com/identifiers.html).
Not sure if this fixes it, help appreciated.


maxiloc:
The issue is the stop url that prevent the crawler from going to the ref

PR #173, opened 2017-06-30
Remove CE searches for GitLab

axilleas:
In recent versions of docs.gitlab.com, we hide the CE page
and use only one for all GitLab documentation. We don't want
CE search results to show as of that change.

https://gitlab.com/gitlab-com/gitlab-docs/merge_requests/94


ElPicador:
I'll deploy the config right now

PR #174, opened 2017-06-30
Update Bootstrap config for beta release

mdo:
I don't think I'm doing this 100% right, but here's what's up:

- We'll be shipping a v4 beta soon. That's the only place our instance of DocSearch will live moving forward (the `v4-alpha.getbootstrap.com` site will remain as-is).
- The bulk of the docs content will live at `getbootstrap.com/docs/4.0/`, but not all pages are there (home, examples, and about aren't).
- I don't want the page's first `<h1>`s to show up in the list—I want the _name_ of the page to show up. Dropping folks on the first heading like that skips a few pixels and doesn't give you the page's name. Is that possible?
- We don't need every `<h4>` and `<h5>` in there, so I've pull them out.

Other than that, we'll likely need to rebuild the index, too. I've reordered a good amount since we added DocSearch originally :).

Lemme know what else I can do, and thanks for all your help!

### DIFF
--- a/configs/bootstrap-v4.json
+++ b/configs/bootstrap-v4.json
@@ -1,16 +1,13 @@
 {
   "index_name": "bootstrap-v4",
   "start_urls": [
-    "https://v4-alpha.getbootstrap.com/"
+    "https://getbootstrap.com/"
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".bd-pageheader h1",
     "lvl1": ".bd-content h1",
     "lvl2": ".bd-content h2",
     "lvl3": ".bd-content h3",
-    "lvl4": ".bd-content h4",
-    "lvl5": ".bd-content h5",
     "text": ".bd-content p, .bd-content li"
   },
   "min_indexed_level": 1,


ElPicador:
The crawler needs a lvl0, so could you shift the `lvl` to start at `0` ?

mattonit:
@ElPicador so would it be ok to replace it with the following?
```
      "lvl0": ".bd-content h1",
      "lvl1": ".bd-content h2",
      "lvl2": ".bd-content h3",
```

ElPicador:
@mattonit exactly

mdo:
Thanks, y'all! Let me know if there's anything else I can change here.

ElPicador:
Change the PR :)

mdo:
@ElPicador Hah, well yeah, just had to merge @mattonit's PR :).

ElPicador:
I'll deploy it right away

maxiloc:
@mdo Actually I don't think this change is going to work as you expect. I am going to revert the change.
What you want to achieve can be done by modifying the snippet of code. I am going to bootstrap it now and send it to you as soon as it's ready.

For the url we are going to use v4-alpha.getbootstrap.com for now. When  getbootstrap.com/docs/4.0/ is live, we'll update the config for the prod and create a second one for the alpha.

maxiloc:
@mdo Here is the snippet I would use

```
  apiKey: '48cb48b22351bc71ea5f12f4d1ede198',
  indexName: 'bootstrap-v4',
  inputSelector: '#input',
  handleSelected: function (input, event, suggestion) {
    var url = suggestion.url;
    url = suggestion.isLvl1 ? url.split('#')[0]: url;
    
    // If it's a title we remove the anchor so it does not jump.
    
    window.location.href = url;
  },
  debug: true // Set debug to true if you want to inspect the dropdown
});
</script>
```

Tell me if that what you have in mind.
If it's not, could you give me a concrete example, maybe using a screenshot.

mdo:
That's great!

PR #180, opened 2017-07-14
update glamorous.rocks config

kentcdodds:
In the future, we'll remove the `rc.` prefix (stands for release-candidate). But for now this is how our URL schema works. Thanks!

### DIFF
--- a/configs/glamorous.json
+++ b/configs/glamorous.json
@@ -2,12 +2,14 @@
   "index_name": "glamorous",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


kentcdodds:
Woo! Thanks! How long does it take for this configuration to be updated so I can start relying on it?

Haroenv:
I'm actually not 100% sure, I know it runs automatically every day, but don't know at what time

kentcdodds:
Ok, that's fine. Thanks so much!

PR #181, opened 2017-07-15
[RFC] Fix url structure to new major.minor

mschinis:
We recently updated our URL structure from `major.minor.patch` to `major.minor`. The API docs were also released in production, so I changed the url to `/api` instead of `/api-beta`


ElPicador:
Seems ok to me, I'll deploy it right now

PR #187, opened 2017-07-24
Update deepstreamhub.json start_urls

MuriloFrade:
Hi,
Thanks so much for the great tool! We would like to change the scope of our search.


ElPicador:
Deploying right now

PR #188, opened 2017-07-25
Add Litho API to search config

JoelMarcey:
The API is Javadoc which is frame-based by default. So I made the starting URL the non-frame version of the docs.

My xpath skills are rusty, so I just made the PR as best as I could for now, will wait for comments on the PR to see what needs to be changed.


ElPicador:
Deploying right now

PR #190, opened 2017-07-27
remove rc from glamorous site settings

kentcdodds:
We'll be releasing the site soon!

### DIFF
--- a/configs/glamorous.json
+++ b/configs/glamorous.json
@@ -2,7 +2,7 @@
   "index_name": "glamorous",
   "start_urls": [
     {
-      "url": "https://(?P<language>.*?)rc.glamorous.rocks/",
+      "url": "https://(?P<language>.*?)glamorous.rocks/",
       "variables": {
         "language": [
           "",


Haroenv:
Perfect, will deploy soon

kentcdodds:
Looks like search is busted on [glamorous.rocks](https://glamorous.rocks/) now :-(

Could I get help debugging this?

screen shot 2017-07-29 at 7 15 46 am

Haroenv:
Hmm, seems odd, I’m going to check that right away

Haroenv:
Seems like `rc.` still works, maybe something is stalled on the deployment @kentcdodds

kentcdodds:
How can I help? 😀

Haroenv:
Well I couldn’t test it fully because I don’t have production credentials. What I’m thinking now is that this configuration is actually matching `.glamorous.rocks` instead of `glamorous.rocks`.

kentcdodds:
Who can we get who could help test things out and find out what needs to be changed?

PR #195, opened 2017-08-10
[RFC] Updates to ember docs

mschinis:
Updated some css selectors after we moved some things around.

Can we also have `ember_versions` with the contents of this `ember_versions_new` or shall I create a separate PR for this?

### DIFF
--- a/configs/emberjs_versions_new.json
+++ b/configs/emberjs_versions_new.json
@@ -55,10 +55,10 @@
         "default_value": "API Documentation"
       },
       "lvl1": "article h1",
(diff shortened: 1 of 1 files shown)


mschinis:
@ElPicador If I want `ember_versions` docsearch config updated with the same configuration as `ember_versions_new`, do I need to make another PR for that?

mschinis:
We can delete this one after the transfer of the config is complete. `ember_versions_new` was created at the time when our new api docs were in beta

ElPicador:
Yes you need to make another PR

PR #201, opened 2017-08-28
Update onsen.json

asial-matagawa:
We would like to add https://onsen.io/v2/guide/ into search results as well as https://onsen.io/v2/api/ .
Thank you.


asial-matagawa:
Thank you for reviewing!

PR #202, opened 2017-08-28
use semantic tags in selectors

drstearns:
I exchanged email with Rémy-Christophe Schermesser about this, and he said I should open a pull request to correct it.

The `.content` and `.hero-body` style classes are part of the CSS framework I happen to use, and I may change that framework in the future. A more reliable set of selectors would leverage the `header` and `main` semantic elements, which I use in my template. I'm careful to include indexable content only within those elements. I also added `dt` and `dd` elements, in case I start using those in the future.

If I've misunderstood anything here, or if you have concerns about this change, feel free to add comments and I can add further commits to this pull request.

Thanks!

### DIFF
--- a/configs/drstearns.json
+++ b/configs/drstearns.json
@@ -5,12 +5,12 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".content h1, .hero-body h1",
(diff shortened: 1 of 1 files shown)


ElPicador:
Seems good to me

PR #203, opened 2017-08-30
[rfc] Update ember config

mschinis:
Updates ember configuration based on:
- new api docs


mschinis:
@ElPicador can we get this merged in?

Issue #186, opened 2017-07-21
Update drone_io documentation


bradrydzewski:
hey all, I was hoping for some help updating the indexing for the [drone_io](https://github.com/algolia/docsearch-configs/blob/master/configs/drone_io.json) documentation. We have moved the documentation:

```diff
{
  "index_name": "drone_io",
  "start_urls": [
-   "http://readme.drone.io/"
+   "http://docs.drone.io/sitemap/"
  ],
```

We have also adjusted the structure. I was hoping that perhaps Algolia could crawl the documentation using the sitemap only (at http://docs.drone.io/sitemap/). This would allow us to make structural changes to the main documentation, without having to re-configure the crawling, since the sitemap structure would never change.

Do you think this would be possible?

I would offer to submit a pull request but was unsure if the below notation was correct, and I was having trouble setting up an environment to test myself (I will keep trying, though)

```diff
-    "lvl0": "header nav a.selected",
-    "lvl1": "main h1",
-    "lvl2": "main h2",
+    "lvl0": "body > ul > li > span",
+    "lvl1": "body > ul > li > ul > li > span",
+    "lvl2": "body > ul > li > ul > li > ul > li > a",
```

ElPicador:
Seems good to me :)
Put everything in a PR and we'll deploy it

PR #204, opened 2017-09-03
update drone.io

bradrydzewski:
Updating configuration per https://github.com/algolia/docsearch-configs/issues/186

I have confirmed the selectors work, but I was hoping someone could review and confirm that this configuration will work as expected. The goal is for docsearch to crawl the documentation sitemap:
http://docs.drone.io/sitemap/

The hope is that by crawling the sitemap and then indexing each page the he sitemap, the crawler configuration and selectors would no longer be impacted by structural changes to the our documentation templates, navigation, etc.

So hoping this configuration works the way I think it will. Thanks for the code review and for providing this service!

### DIFF
--- a/configs/drone_io.json
+++ b/configs/drone_io.json
@@ -1,7 +1,7 @@
 {
   "index_name": "drone_io",
   "start_urls": [
-    "http://docs.drone.io/"
+    "http://docs.drone.io/sitemap/"
   ],
   "stop_urls": [
     "/tags/",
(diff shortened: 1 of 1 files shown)


ElPicador:
The `selectors` should be the selectors of the page indexed not the page that links to it.
So you probably do not need to update them.

maxiloc:
Let's merge this. I'll update the config if needed.

PR #206, opened 2017-09-06
Update teamcofh.json

Tonius:


configs/teamcofh.json
```diff
@@ -5,13 +5,11 @@
...
-    "lvl3": ".uk-container h4",
-    "lvl4": ".uk-container h5",
-    "text": ".uk-container p, .uk-container li"
+    "lvl0": "#cofh-docs-content h1",
+    "lvl1": "#cofh-docs-content h2",
+    "lvl2": "#cofh-docs-content h3",
+    "text": "#cofh-docs-content p, #cofh-docs-content li"
   },
-  "min_indexed_level": 1,
+  "min_indexed_level": 0,
```
ElPicador: why did you change this?

Tonius: The search currently seems to only return results for subheaders (h2, h3, etc.). I would like it to return results for pages as a whole, by page title (h1). If I understand correctly, this option configures it so that higher level results (page titles) show up. Please correct me if I'm wrong though.

maxiloc:
I deployed the config, should be live in a few minutes

PR #207, opened 2017-09-07
Update material-ui.json

oliviertassinari:
Work around for #190. It should only match once and not twice.

### DIFF
--- a/configs/material-ui.json
+++ b/configs/material-ui.json
@@ -17,7 +17,7 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "//*[contains(@class, 'algolia-lvl0')]",
+      "selector": ".algolia-drawer .algolia-lvl0",
       "type": "xpath",
       "global": true
     },


oliviertassinari:
It seems that this fix hasn't been deployed. I'm right to think so?

ElPicador:
I've checked, it was deployed

oliviertassinari:
@ElPicador Thanks for checking it out. Then, something is wrong, this change hasn't changed the indexed data. I will have a second look.

ElPicador:
My mistake, I've reviewed this too quickly.
The selector should be a XPATH selector, not CSS.
So either change it to XPATH, or only but a string with the selector

oliviertassinari:
@ElPicador Awesome, I will fix this 👍

oliviertassinari:
CSS path seems to be supported but I won't try the devil `//*[contains(@class, 'algolia-drawer')]//*[contains(@class, 'algolia-lvl0')]` selector should do the job. hold my beer, a PR is coming.

https://github.com/algolia/docsearch-configs/blob/7029f78a9152fd71201294418011ee18f26008ab/configs/flowtype.json#L15

oliviertassinari:
Looks at the docs, I'm getting more confident, let's try the devil.

PR #208, opened 2017-09-08
Updated ckeditor_nightly config

m-turek:
We would like to have each section (API, Framework, Builds, Features, Examples) tagged.
Also introduced separate selectors for API. Hope it's fine


maxiloc:
Everything looks good. I deployed should be live in a few minutes

PR #209, opened 2017-09-11
Updated flexilivre.json

benjaminclot:
Added a stop url pattern


benjaminclot:
Ok, so you can ignore my PR :)
The fact that higher level pages would not be parsed wasn't clear to me. Thanks!

Space69:
All right then.
No problem. 
Feel free to contact us whenever you want and welcome with using docSearch.

PR #210, opened 2017-09-11
Update material-ui.json

oliviertassinari:
Following #207

### DIFF
--- a/configs/material-ui.json
+++ b/configs/material-ui.json
@@ -18,7 +18,6 @@
   "selectors": {
     "lvl0": {
       "selector": ".algolia-drawer .algolia-lvl0",
-      "type": "xpath",
       "global": true
     },
     "lvl1": {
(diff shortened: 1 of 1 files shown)


configs/material-ui.json
```diff
@@ -47,7 +46,5 @@
     }
   },
   "min_indexed_level": 1,
-  "js_render": true,
```
oliviertassinari: Our docs pages are now server side rendered. This change should speed up the crawling.

configs/material-ui.json
```diff
@@ -18,7 +18,6 @@
   "selectors": {
     "lvl0": {
       "selector": ".algolia-drawer .algolia-lvl0",
-      "type": "xpath",
```
oliviertassinari: Don't miss CSS selector with xpath config.

maxiloc:
I updated a bit the config and deploy. Should be live in a few minutes

oliviertassinari:
It's working like a charm 💃 .

PR #211, opened 2017-09-12
Corrected errors in API text selectors. Removed stop_urls

m-turek:
There was an error in API text selectors.

Also I was trying to use `"page_rank": 1` for 

```
    {
      "url": "https://ckeditor5.github.io/docs/nightly/ckeditor5/latest/api/module_core_editor_editorconfig-EditorConfig.html",
      "selectors_key": "api",
      "page_rank": 1,
      "tags": [ "api" ]
    }
```

but it didn't work (results for `EditorConfig` still have `page_rank` == 0). So I changed the config a bit hoping that it may help.

### DIFF
--- a/configs/ckeditor_nightly.json
+++ b/configs/ckeditor_nightly.json
@@ -2,14 +2,14 @@
   "index_name": "ckeditor_nightly",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


s-pace:
Glad to hear this new,

PR #213, opened 2017-09-17
Update amphp config

kelunik:
I hope I did everything right, this should include the sidebar from search results, but still crawl the content linked there. It should also index options now, which are usually in a table.


s-pace:
Glad to hear this new, 

PR #214, opened 2017-09-18
Account for v3.3 and v4.0 docs

acmetech:
make the config consistent with Bootstrap v3:
* add the https url to start_urls
* add min_indexed_level


configs/todc-bootstrap-v3.json
```diff
@@ -1,6 +1,7 @@
 {
-  "index_name": "todc-bootstrap",
+  "index_name": "todc-bootstrap-v3",
```
Space69: This attribute must be the same as the file name. Don't change it please

configs/todc-bootstrap-v3.json
```diff
@@ -1,6 +1,7 @@
...
   "start_urls": [
+    "https://todc.github.io/todc-bootstrap/",
```
Space69: Please give only one protocole. Otherwise you will have duplicated data

acmetech: Okay. FYI - I was following what is in the bootstrap-v3 config.

Space69: Thanks. It is possible to do it as long as you have canonical url

configs/todc-bootstrap-v3.json
acmetech: Here's the text of an email conversation I was had:

My question:

> Is it possible for me to add/update indexes? This site is documentation for a Google styled theme for Bootstrap. The documentation includes the full Bootstrap documentation as well as documentation for any additional/modified components. Bootstrap includes documentation for multiple versions and I'd like to make the versions on the theme site searchable via their own indexes.

Response:

> Hey Tom,

> I would definitely recommend to add it into your configuration file as long as it is within the same website.

Does this mean that I should update my start_urls and selectors within the one configuration file?

```https://todc.github.io/todc-bootstrap/docs/3.3.7/,```
```https://todc.github.io/todc-bootstrap/docs/4.0/```

Thanks

Comment (Space69) [reply configs/todc-bootstrap-v3.json]: Don't worry, I noticed it was you.

As long as the new content is linked from the start url or its sub ones, it will be updated at the next crawl (every 24h)

Comment (acmetech) [reply configs/todc-bootstrap-v3.json]: I'm looking for each version of the docs to be searched separately. I don't want to get hits back that point to the v3 docs when I'm in the v4 docs. Wouldn't this require separate indexes and configurations?

The urls given about are not the way the site is currently structured, but it's what I'm moving towards

Comment (s-pace) [reply configs/todc-bootstrap-v3.json]: Hi,

If the different versions have the same DOM structure, you just have to add a new start url.

Otherwise you will have to use different selector key. I will support you in order to that.

Best,

Comment (acmetech) [reply configs/todc-bootstrap-v3.json]: Hi,

Versions 3.x and 4.x would have different DOM structures. The selector keys in the current config are correct for v3.x. These are the keys for v4.x:

``` 
"selectors": {
    "lvl0": ".bd-toc-item.active > a",
    "lvl1": ".bd-content h1",
    "lvl2": ".bd-content h2",
    "lvl3": ".bd-content h3",
    "text": ".bd-content p, .bd-content li"
  },
```

Do I need to add these selectors to the existing ones?

```
  "selectors": {
    "lvl0": ".bs-docs-header h1, .bd-toc-item.active > a",
    "lvl1": ".bs-docs-container h1, .bd-content h1",
    "lvl2": ".bs-docs-container h2, .bd-content h2",
    "lvl3": ".bs-docs-container h3, .bd-content h3",
    "text": ".bs-docs-container p, .bs-docs-container li, .bd-content p, .bd-content li"
  },
```

Thanks

s-pace: It those selector have no conflict from one version to another and you don't have  any overlap, yes it will works since we do an OR between all of those.

Let me know when I should try it

acmetech: The selectors, currently, do not overlap between v3.x and v4.x. I've uploaded the v4 **content** to https://todc.github.io/todc-bootstrap/docs/4.0/

FYI - The updated CSS hasn't been uploaded and the links will probably not work.

s-pace: Thanks, please let us know as soon as your layout is ready and that your website is stable. We will crawl it once it is in production.

acmetech: Will do. Thanks

acmetech: I believe we're ready to go.

Thanks

s-pace:
This new configuration is live, feel free to ask us anything if you need further details

acmetech:
Awesome, thanks. Does this mean the index is currently up to date?

s-pace:
You are very welcome. Yep it was when i deployed it a day ago. It will be updated every 24h

acmetech:
Thanks. When I do a search for buttons from http://todc.github.io/todc-bootstrap/docs/4.0/getting-started/introduction/ the results link to https://todc.github.io/todc-bootstrap/docs/3.3/css/ instead of http://todc.github.io/todc-bootstrap/docs/4.0/components/buttons/

s-pace:
Given that the configuration provided an URL not available (because of the wrong scheme https instead of http), neither link from v4 was crawled nor scraped.
This is fixed by this commit. It is live now.

A new version of our scrapper will be soon released in order to handle any scheme and avoid this common mistake.

Feel free to ping us if you need any further details or if you have any feedback

acmetech:
Thanks for your help.

PR #216, opened 2017-09-20
update boosted.orange.com v4 documentation path

Lausselloic:
Hi,

We've update our deployment path for boosted.orange.com, documentation is now at boosted.orange.com/docs/4.0

So this pull request update the indexation path


s-pace:
Glad to hear this new, 

Issue #221, opened 2017-09-27
public "metadata" in the docsearch configs?


phrawzty:
It would be useful for tracking and reporting purposes to have some fields in the configs that aren't necessarily used by the scraper. Immediately I'd like to add:
* `name`: Canonical, human-readable label (such as company or project name) for a given config.
* `human_url`: A URL that a human could go to and get the documentation site.
I'm not married to those key names - suggestions welcome. 😄

maxiloc:
This is already doable 😄, it's just a matter of adding them.

For name => it's what we try to do with index_name but sometimes it can be quite tricky.

phrawzty:
Following the conversation on Slack, I further propose an object `tags` which can contain whatever human-readable elements that may be interesting. Example:
```json
"tags": {
  "name": "Super PHP Framework 2000",
  "url": "https://human.readab.le/entrypoint#to-the-docs",
  "categories": [
    "framework",
    "php",
    "ecommerce"
  ]
}
```

phrawzty:
Following the conversation on Slack, I further propose an object `tags` which can contain whatever human-readable elements that may be interesting. Example:
```json
{
	"tags": {
		"name": "Super PHP Framework 2000",
		"url": "https://human.readab.le/entrypoint#to-the-docs",
		"categories": [
			"framework",
			"php",
			"ecommerce"
		]
	}
}
```

phrawzty:
Concept validated. Now we figure out how to automate and scale. ;)

PR #224, opened 2017-09-27
[tags] laravel

phrawzty:
First pass at a `tags` object in a Docsearch config (see #221). I selected Laravel more or less randomly. :) If we like how this looks, we can start rolling the format out to existing configs, and introducing it to new configs as a best practice as well.

### DIFF
--- a/configs/laravel.json
+++ b/configs/laravel.json
@@ -1,4 +1,13 @@
 {
+  "tags": {
+    "name": "Laravel",
+    "url": "https://laravel.com/docs/",
+    "categories": [
+      "ecommerce",
+      "framework",
+      "php"
+    ]
+  },
   "index_name": "laravel",
   "start_urls": [
     {


maxiloc:
I am good with this version. We need to think about how we do this at scale. We are working hard at removing manual operations needed to operate docsearch.
Right now we are starting to think how we can make the config creation automatic, so we need to find a way to make that automatic as much as possible as well

dzello:
What if we automate adding the URL they put in this field? I think @maxiloc has a good point about wanting to avoid too much manual work in the configs. Also there are other places we could keep metadata for feeding into our reporting stack.

screenshot 2017-09-27 15 11 13

phrawzty:
Agreed fully re: automation & scale. That URL field is likely the best place to start. Regarding the tags, if we don't want to manage it ourselves, we can let the user decide from a likely list, I suppose?

phrawzty:
As per further discussion on Slack, we'll move _some_ or _all_ of this to a private repo. To be continued.

PR #227, opened 2017-10-01
New domain name

oliviertassinari:
So we get a cleaner URL 💅


s-pace:
Will be live within 1h

s-pace:
Nice, it will be live within 1 hour

PR #228, opened 2017-10-02
Fix Chinese and Japanese + Add Korean and French :)

Atinux:


Atinux:
It is now @Haroenv 

You were too fast :-)

s-pace:
It will be live within 1h maximum.

Thanks for the update!

inouetakuya:
Thank you so much !!!

PR #232, opened 2017-10-09
Removing old releases from indexes.

laffan:
Anything with a url that begins `http://docs.polysync.io/releases` should be ignored.


s-pace:
LGTM

s-pace:
Do you want https://docs.polysync.io/release-notes/* to be scraped?

PR #235, opened 2017-10-11
Update testcafe lvl0 selector

VasilyStrelyaev:
We've updated TestCafe documentation site at https://devexpress.github.io/testcafe/ and here's the new selector.


s-pace:
Perfect

PR #238, opened 2017-10-19
Change level 1 and 2 selectors for Express Gateway

XVincentX:
This PR will modify the selectors for level 0 and 1 indexes. Hopefully these should match the way we organize the contents on our website.

image

Level 0: *Policies Reference*
Level 1: *Customization*

### DIFF
--- a/configs/express-gateway.json
+++ b/configs/express-gateway.json
@@ -8,8 +8,14 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": "section h1",
-    "lvl1": "section h2",
+    "lvl0": {
+      "selector": "//nav[@class='sidebar-docs']/ul/li/a[../input[contains(@class,'expanded')]] | //section[@class='doc-wrapper']/h1",
+      "type": "xpath"
+    }
+    "lvl1": {
+      "selector": "//nav[@class='sidebar-docs']/ul/li/ul/li/a[../input[contains(@class,'expanded')]] | //section[@class='doc-wrapper']/h2",
+      "type": "xpath"
+    }
     "lvl2": "section h3",
     "lvl3": "section h4",
     "lvl4": "section h5",
@@ -17,3 +23,4 @@
   },
   "nb_hits": 1679
 }
+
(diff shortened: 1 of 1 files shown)


XVincentX:
@s-pace As you've been the one adding this file — could you please review this and merge it if it makes sense to you?

Thanks!

s-pace:
Hi Vincenzio,

Please try your selectors on your web page here: https://www.express-gateway.io/docs/configuration/gateway.config.yml/

You will see that it doesn't have the expected behaviour.
1st: lvl0 matches 2 node whereas it should only match one element since your web page is related to one level 0 only (Configuration)
2nd: lvl1 and lvl2 selector match the same element while it shouldn't be so

In order to gather record by same hierarchical path, the easier solution will be to change the DOM generation as exposed in the mail I sent you.

If you need more explanations, ping me.

We need consistency regarding the way you expose level within your webpage. One rule can't fit every different cases you have.

XVincentX:
Hey @s-pace,
thanks a lot for the review.

So, I wrote the selector in a way it could match two elements on purpose, on both cases, as it fits the way (probably wrong, as you point) we're crafting the html page.

If I understand correctly you don't want to handle the case when you get two results (I thought that you could just grab the first one in that case). Is that right?

If so, I'll try to rewrite the page so all the hierarchy is embedded in the page.

Thank you.

s-pace:
Glad to help,

The issue is the logic introduced regarding the "double selector" is pretty intricate:
If the first match of the selector matches anything, it informs us that the current web page is a special one and thus, other selectors should behave differently accordingly

In order to have this behaviour with XPATH/CSS selector, it will expose way to much verbose selector that will not be maintainable.

The simpler, the better this is why having a consistency policy regarding the way you expose hierarchy will be better for DocSearch and even regarding UX purpose. 

Feel free to ping us if you need clarification or tips

XVincentX:
@s-pace 

Is Docsearch Crawler able to get things even if hidden?

I am in the process of putting better hierarchy informations in the pages, but it's not an information I'd like to display — but rather just put it in the DOM so you can grab it.

Would `display: none` work?

s-pace:
@XVincentX Yep it should work 👍

XVincentX:
Cool, thanks. A PR is on the way!

PR #240, opened 2017-10-23
Add version 10.5

efarem:
Add docs version 10.5


s-pace:
LGTM

s-pace:
I update the number of hints and remove version that are down.
Let me know if it suits you.
Greetings

PR #241, opened 2017-10-24
Update AtlassianPS config

lipkau:


s-pace:
LGTM, let's try it

s-pace:
I update the number of hints and trigger the deploy.
I checked it locally and it looks relevant.
You could check by yourself if it suits you.
Let me know.
Congrats Oliver!

lipkau:
Thank you for the support!
Wouldn't have done it by myself.

I really like the outcome:
image

s-pace:
As for the missing lvl2=h1 About ConfluencePS,
This is expected given that another lvl2 is closer through the flow to the record. "SHORT DESCRIPTION" is the previous one closest to the record considered by "many wiki" along the flow:
image


The h1 About ConfluencePS is a lvl2 for the next record which is **about_ConfluencePS**:
image

The h1 About ConfluencePS is considered as any lvl2 and isn't above SHORT DESCRIPTION which as the same level and thus will be process as the same kind no matter the way you display it.

lipkau:
Ah. I missed that. Will fix it and send a minor change as PR.
Damn those autogenerated files!

PR #247, opened 2017-10-31
change tinymce configs min_indexed_level

fyrkant:
Hi again!

I figured out that our H1s weren't being indexed, making the search on the tinymce docs working pretty bad and I think this is the reason. Perhaps there is some reason that this was set to `2` before?

### DIFF
--- a/configs/tinymce-v3.json
+++ b/configs/tinymce-v3.json
@@ -18,6 +18,6 @@
     "lvl4": ".content h4",
     "text": ".content p"
   },
-  "min_indexed_level": 2,
+  "min_indexed_level": 1,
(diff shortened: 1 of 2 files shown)


s-pace:
Hi Mattias,

This attribute is set mostly because you have sometimes a potential record without any content nor it had another level set.

You can have a look and I can merge it in order you test it and let me know if it suits you.

fyrkant:
I would really appreciate that, thank you!

s-pace:
Jobs pending. It will be available within 1 hour.

fyrkant:
Seems like it's already been updated and works great! Thank you so much.

PR #249, opened 2017-11-09
Update configuration for meta box

rilwis:


s-pace:
Good change, always a good practise to have a default_value set. It is a good fallback.

rilwis:
Thanks for your comment. I’ve just added `default_value`.

PR #257, opened 2017-11-14
remove API V1 links from being scrapped.

lr-pascal:


s-pace:
@lr-pascal 
Could you please add a specific class such as "docSearch-lvl0" to the matching element "#content h1, ul label.md-nav__link--active"
It will allow use to avoid browser rendering in order to have a good collection by lvl 0.
We don't want to have such client rendering anymore. 
Feel free to ask me further details.
Cheers

lr-pascal:
@s-pace I'm definitely open to changing this, can I get more details? My understanding is that you want me to have lvl0 set to a specific class that's always going to be in the title of my pages?

s-pace:
Actually we would like to gather by highest level.

In fact those level match nav a.selected.

Since it is only style oriented, we can't target them without client side rendering.

s-pace:
I double check. You are doing full browser rendering regarding the content.

This small change will not be useful.

PR #264, opened 2017-11-22
Create metro.json

Haroenv:
so the site isn't live yet, but this should be the config since the site is the same as Jest

site: http://facebook.github.io/metro or http://facebook.github.io/metro-bundler

cc @cpojer

### DIFF
--- a/configs/metro.json
+++ b/configs/metro.json
@@ -1,0 +1,27 @@
+{
+  "index_name": "jest",
+  "start_urls": [
+    {
+      "url": "http://facebook.github.io/metro-bundler/docs/",
+    }
+  ],
+  "selectors_exclude": [
+    ".hash-link",
+    ".edit-page-link"
+  ],
+  "selectors": {
+    "lvl0": {
+      "selector": "//nav//*[contains(@class,'navListItemActive')]/../../h3",
+      "type": "xpath",
+      "global": true,
+      "default_value": "Documentation"
+    },
+    "lvl1": ".post header h1",
+    "lvl2": ".post article h2",
+    "lvl3": ".post article h3",
+    "lvl4": ".post article h4",
+    "lvl5": ".post article h5",
+    "text": ".post article p, .post article li"
+  },
+  "min_indexed_level": 2,
+}
--- a/configs/metro.json
+++ b/configs/metro.json
@@ -2,7 +2,20 @@
   "index_name": "metro",
   "start_urls": [
     {
-      "url": "http://facebook.github.io/metro-bundler/docs/",
+      "url": "http://facebook.github.io/metro-bundler/docs/(?P<lang>.*?)/getting-started.html",
+      "variables": {
+        "lang": [
+          "en"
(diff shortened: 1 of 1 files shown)


cpojer:
That's awesome! Can I send a PR for this repo when we rename the website from metro-bundler to metro? :)

Haroenv:
Yep, you can send a PR to this repo with any changes in the configuration. I put metro for now, but if metro-bundler will launch earlier, that can be changed

cpojer:
Oh, there will only be one. It'll start as metro-bundler and then we'll rename that to metro. Unfortunately the renaming takes a while internally.. :(

configs/metro.json
```diff
@@ -0,0 +1,27 @@
+{
+  "index_name": "jest",
```
s-pace: Typo

configs/metro.json
```diff
@@ -0,0 +1,27 @@
...
+  "start_urls": [
+    {
+      "url": "http://facebook.github.io/metro-bundler/docs/",
```
s-pace: They plan to change this

s-pace:
Hi again @cpojer,
@Haroenv has shown you the right process. Please do so as soon as you are ready from your site. I will double check it/ tailored/ deployed/provide you the snippet as soon as you are live.
Thank you guys

configs/metro.json
Haroenv: good call, let's use the one that's online first

cpojer:
Site is live here: https://facebook.github.io/metro-bundler/ it only contains the scaffolding for the docs and we'll fill it out later.

s-pace:
@cpojer 

Congratulations, your search is now ready!
I've successfully configured the underlying crawler and it will now run every 24h.

You're now a few steps away from having it working on your website:


  apiKey: '0cf53e80b4f3b371219cdcaa6b62e08b',
  indexName: 'metro',
});
</script>

- Add a search input if you don't have any yet, and update the inputSelector value in the code snippet to a CSS selector that targets your input field.
- Optionally customize the look and feel by following the DocSearch documentation (https://community.algolia.com/docsearch/documentation/)
- You can also check your configuration in the github repo (https://github.com/algolia/docsearch-configs/blob/master/configs/metro.json). Feel free to open pull requests!

Feel free to get back to us if you have any issues or questions regarding the integration.
We'd also be happy to get your feedback and thoughts about DocSearch - so we can continue to improve it.

Have a nice day :)

cpojer:
Thank you very much.

s-pace:
You are welcome, If you need to add an new language, you will only need to add it into the lang array.
These "language" variables will also behave as arguments for faceting. Thus you can filter which kind of results you want to request. This will be made from youir snippet by adding the following input: 
`algoliaOptions: { facetFilters: ["lang:en"] } `

Haroenv:
you can also do that on jest by the way, which already has different languages, but the search isn't enabled for the others yet @cpojer

PR #265, opened 2017-11-23
[rvm] remove infinity symbol and add min_indexed_level

phrawzty:


phrawzty:
👍 from @s-pace who was right beside me, heh. 😄

PR #278, opened 2017-12-04
Update reason-react urls

chenglou:
https://reasonml.github.io/reason-react/docs/ is actually 404.
I'm not sure why https://reasonml.github.io/reason-react/docs/en/installation.html is listed as a start url. Maybe that's a default (thanks though =))

### DIFF
--- a/configs/react-reason.json
+++ b/configs/react-reason.json
@@ -1,8 +1,7 @@
 {
   "index_name": "react-reason",
   "start_urls": [
-    "https://reasonml.github.io/reason-react/docs/",
-    "https://reasonml.github.io/reason-react/docs/en/installation.html"
+    "https://reasonml.github.io/reason-react/"
   ],
   "stop_urls": [],
   "selectors": {


s-pace:
LGTM, let's have a look

s-pace:
Hi @chenglou ,

The current state of your configuration scrap and crawl only your english version, available under https://reasonml.github.io/reason-react/docs/en/*.

I revert your change because they will introduce noise from other languages. We can support languages by using variables within the start urls. You can [have a look to the following changes](https://github.com/algolia/docsearch-configs/commit/dfef794bf22f10cc539ce2163b033a5c4549a3e2#diff-194ce1c8b7c8cd06fb0e614d21f6bd64).
You will only need to add some new "lang" values.

Those variables will also behave as arguments for faceting. Thus you can filter which kind of results you want to request. This will be made from your snippet by adding the following input: 
algoliaOptions: { facetFilters: ["lang:en"] } 

Since it requires you to change the snippet on your web site, I would rather inform you before than launching this live.

FYI: We were dealing on this point with @JoelMarcey (for the Docusaurus project), @cpojer for metro and @rickyvetter for this precise project.

Let us know if you need our support guys.

Cheers

PR #279, opened 2017-12-04
Update config for bucklescript.github.io

chenglou:
We've launched a new doc site that's based on the same doc site generator as [reason-react](https://github.com/chenglou/docsearch-configs/blob/fb21662f71a0dac7c909349181a5a8702a066f3a/configs/react-reason.json), so I think most of the configs transfer over.

Also incorporated changes from #278


s-pace:
same as #278, I have updated it for the English version.

PR #289, opened 2018-01-03
Update improbable.json

lottec:
Add new version of docs, and index code tabs

### DIFF
--- a/configs/improbable.json
+++ b/configs/improbable.json
@@ -6,6 +6,7 @@
       "variables": {
         "version": [
           "12.0",
+          "11.2",
           "11.1",
           "11.0",
           "10.5",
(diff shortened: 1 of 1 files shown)


Susiew78:
It looks good to me

configs/improbable.json
```diff
@@ -6,6 +6,7 @@
       "variables": {
         "version": [
           "12.0",
+          "11.2",
```
Susiew78: Needs fixed

configs/improbable.json
```diff
@@ -32,10 +33,10 @@
     "lvl2": "#docPage h3",
     "lvl3": "#docPage h4",
     "lvl4": "#docPage h5",
-    "text": "#docPage p, #docPage li"
+    "text": "#docPage p, #docPage li, #docPage code"
```
Susiew78: Were are documents

s-pace: We don't recommend to do that since it will introduce noise and duplicate within your index:
As for:
```
<p>Check this succeeded by running <code>spatial update</code> in your terminal. You should get the output:</p>
```
It will create one extra records for `<code>spatial update</code>` since the whole one is already a record given that it matches `#docPage p`

s-pace:
Closing since no follow up

PR #290, opened 2018-01-08
Update CKEditor configs

m-turek:


s-pace:
Indexing th and td element may introduce a lot of noise, let's have a look.

s-pace:
Hi @m-turek, I have removed the `td` and `th` selectors since It will introduce too much noise for example for (this page)[https://docs.ckeditor.com/ckeditor5/latest/framework/guides/architecture/intro.html], it will create 3 different records for each row with the same level.
You can add dedicated nested classes if you want to have a better relevance.

m-turek:
Thank you @s-pace 
We need to be able to search the contents of [this table](https://docs.ckeditor.com/ckeditor5/latest/builds/guides/migrate.html#Configuration-options-compatibility-table)
Should I add a custom class to the `<table>` element and put that class into docsearch config?

s-pace:
You can had a special classes to each element from the 1st column like `docSearch-lvl3` and `docSearch-txt` to each element of the second one. Hence, each element of the row will be related.

m-turek:
So, do I need to add anything to docsearch config?
Sorry, I'm not sure if I understood. I should add `docSearch-lvl3` class to every table cell in first column, and `docSearch-txt` to each cell in second column? (that doesn't sound good to me because there are a lot of cells :D)
Currently the contents of the table never appear in search results

screen shot 2018-01-09 at 19 47 35

s-pace:
You will need to firstly do the change on your web site and then update your configuration accordingly.
Last but not least, the result will be changed at the next crawl/scrap which is at maximum every 24h.

PR #296, opened 2018-01-22
Update pantheon.json

rachelwhitton:


configs/pantheon.json
```diff
@@ -16,13 +16,14 @@
     "lvl2": "#doc h3",
     "lvl3": "#doc h4",
     "lvl4": "#doc h5",
-    "text": "#doc p, #doc li"
+    "text": "#doc p, #doc li, #doc ol, #doc code, #doc alert"
```
s-pace: We don't recommend to extract those elements since they will include noise and duplicates (`ol` encompasses `li`)
`code` element mostly introduce noise. 
What about adding a more specific classe?

rachelwhitton: Thanks @s-pace! I updated the PR to add only our notes and warnings to the text selector.

s-pace: 👍

PR #300, opened 2018-01-28
Update bootstrap_italia.json

francescozaia:
We're removing the docs version number from the URL


s-pace:
Congrats for the new version!

PR #301, opened 2018-01-29
Update codefresh.json

vmechkauskas-codefresh:
-- updated selectors and min_indexed_level


s-pace:
👋 @vmechkauskas-codefresh,

Your selector aren't working. I have reverted your configuration in order to avoid wrong crawl.

Please ensure that those selectors are working in the first place. You can use `Ctrl + F` within the `Elements` windows from your inspector in order to check it.

Cheers

PR #308, opened 2018-02-06
Always search and return results from 'current'

2m:
This aligns the configuration to the akka_io config. We are no longer sending version parameter to algolia when searching.

### DIFF
--- a/configs/akka-http.json
+++ b/configs/akka-http.json
@@ -2,42 +2,30 @@
   "index_name": "akka-http",
   "start_urls": [
     {
-      "url": "https://doc.akka.io/docs/akka-http/(?P<version>.*?)/(?P<language>.*?).html",
+      "url": "https://doc.akka.io/docs/akka-http/current/(?P<language>.*?).html",
       "variables": {
         "language": [
           "scala",
           "java"
-        ],
-        "version": {
-          "url": "https://doc.akka.io/docs/akka-http/",
-          "js": "var versions = []; for(var i = 0 ; i < document.getElementsByTagName(\"a\").length; i++) { versions.push(document.getElementsByTagName(\"a\")[i].attributes[0].value); }; versions = versions.filter(function(e) { return !e.startsWith('?') && e.endsWith(\"/\") && !e.startsWith('rp') && (parseInt(e, 10) > 1 || e.startsWith('current') || e.startsWith('snapshot')); }).map(function(e) { return e.replace(/\\//g, ''); }); return JSON.stringify(versions);"
-        }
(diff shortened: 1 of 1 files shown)


s-pace:
Hi, those changes will not have any impact since you are using the language as an parameter `?language=java`. We are note currently handling such filtering from parameters.

Could it be possible to add such value within the url itself (/java/) or as a DOM element (meta or not) in the pages?

2m:
The issue we currently have is the duplicate search results that are returned. For example try searching `extract` here: https://doc.akka.io/docs/akka-http/current/

We noticed that the duplication is not present in `akka` therefore with this PR I tried to align akka-http configuration to the akka configuration which is working fine.

Are you saying that we should remove language parameter as well for this to work? We could do that as well, as the language choice is now completely handled client side (https://github.com/akka/akka-paradox/pull/23).

s-pace:
I am saying that we cannot differentiate at the crawling time whether or not the page is fro scala or java.

The duplicate issue you have pointed isn't because of this fact but because we have indexed records without `text` attribute. The following commit has fixed that.

Since the language is set locally and that the content is common, we will crawl the website once without acknowledging which language it presents. It should works this way. Please remove the `facetFilters` input from your snippet since our index is common.

Another solution would have been to put the language within the url itself : like https://doc.akka.io/docs/scala/akka-http/current/ or to add a meta element that will be collected in order to categorised the content although it should be fine this way.

LMK. We must do the same for `akka_io` which isn't updated any more (same issue from parameter)

2m:
Our old documentation tool generated different URLs for scala and java documentation. Since then we were slowly moving towards merging these two documentation trees and now the content is not duplicated anymore.

We will remove the `facetFilters` with the next release of akka and akka-http as we will update the documentation then.

Is that fine, that we will still send the `facetFilters` parameter for some time until the next release?

s-pace:
@2m Alright,
I have found a way to bypass it: 9f6d6b3.

I have updated it and deployed it. It is live, let me know what you think of it.

Cheers

2m:
I have just checked [here](https://doc.akka.io/docs/akka-http/current/) and it does not return any results if the facetFilters (`language=scala`) are provided.

s-pace:
Please double check, It works right now (old crawling in queue)

2m:
Perfect! Thank you very much!

s-pace:
You are welcome.

Should I do the same for `akka_io`?

2m:
Yes, that would be great!

s-pace:
Done

PR #318, opened 2018-02-19
pre-commit: https

asottile:


s-pace:
👍@asottile,
Deployed, will be effective at the next crawling

PR #326, opened 2018-03-02
commercetools: make non-jekyll content from CMS and code examples indexed, too.

nkuehn:
* site contains pages with content from other sources, too so the #content selector works better
 * content in figures is a paragraph
 * more restrictive exclusion


s-pace:
👍 
`stop_urls` are regex so I will update it to :)

PR #328, opened 2018-03-05
Add version 3.0 to Laravel-Excel index

patrickbrouwers:
We just added `3.0` to the versions of our documentation. If I understand the config correctly, the version should be added here to make indexable.

Thanks!


s-pace:
👏 

Please think about adding a [sitemap](https://www.sitemaps.org/)

patrickbrouwers:
Thanks. Yes, we still need to update some SEO stuff :)

PR #329, opened 2018-03-07
Change GitLab tag from ee to gitlab

axilleas:
# Pull request motivation(s)

We need to display GITLAB instead of EE in our custom search results, so we need to change the tag https://gitlab.com/gitlab-com/gitlab-docs/merge_requests/203.


s-pace:
👍 
Live

PR #331, opened 2018-03-07
GitLab docs: set "only_content_level": true

ramosmd:
# Pull request motivation(s)

For https://gitlab.com/gitlab-com/gitlab-docs/issues/134

@s-pace does this solve the problem with the description? https://gitlab.com/gitlab-com/gitlab-docs/issues/134

Results display headers only, with no content accompanying the hit.

Have a content description accompanying every hit that only shows headers

cc/ @axilleas

### DIFF
--- a/configs/gitlab.json
+++ b/configs/gitlab.json
@@ -78,6 +78,7 @@
   ],
   "min_indexed_level": 1,
   "scrap_start_urls": false,
+  "only_content_level": true,
   "conversation_id": [
     "434475832"
   ],


ramosmd:
@s-pace considering that we have a  meta description tag `<meta name="description" content="content">`, how can we change the index to include the description in the search results when available?

s-pace:
This will fix such issue indeed.

`content` here is only a synonym of the `text` attribute.

👏

ramosmd:
Thanks @s-pace! But the DocSearch stopped working 😢 https://docs.gitlab.com/

s-pace:
My bad, I missed a step.

Fixed

ramosmd:
Oh, great, thanks! ❤️

PR #333, opened 2018-03-08
Update gatsbyjs.json

KyleAMathews:
# Pull request motivation(s)


s-pace:
Already live 👏 

Regarding the content and the huge project you are doing, we will feature it too since it is helpfull for the community.

Cheers

PR #338, opened 2018-03-10
Update jovo.json with blog

jankoenig:
# Pull request motivation
Added search globally to our site: https://www.jovo.tech/ and wanted to make sure it also finds the tutorials we offer, which are located in our blog.

Added the blog url to the start_urls.

### DIFF
--- a/configs/jovo.json
+++ b/configs/jovo.json
@@ -6,8 +6,12 @@
       "page_rank": -1
     },
     {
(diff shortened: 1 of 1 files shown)


s-pace:
Usually we do not provide DocSearch for content which isn't technical documentation but since you have many tutorials from your blog part, we will consider it as technical documentation and we will feature those parts.

jankoenig:
Great, thank you!

PR #344, opened 2018-03-15
Added reference / and referenz as they should be indexed

dflorey:
Adjusted the Ultradox configuration to include /reference/ and /referenz/ in the index by removing them from the stop_urls section.

### DIFF
--- a/configs/ultradox.json
+++ b/configs/ultradox.json
@@ -13,7 +13,8 @@
     "https://help.ultradox.com/en/guides/overview.html"
   ],
   "stop_urls": [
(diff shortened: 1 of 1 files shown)


dflorey:
You are right! I've excluded the generated API docs under /reference/api, this should index the manually written pages only

s-pace:
Let's have a look

s-pace:
Better this way @dflorey . Changes already live, let us know if it suits you well

dflorey:
Renders perfectly now!
Thanks a million!

PR #345, opened 2018-03-15
Revert "Update index selectors for Express-Gateway"

s-pace:
Reverts algolia/docsearch-configs#343
@XVincentX 
Please revers the order of your elements within your page. It should be:
```html
<p class="algoliaLv0">...</p>
<p class="algoliaLv1">...</p>
<p class="algoliaLv2">...</p>
```

### DIFF
--- a/configs/express-gateway.json
+++ b/configs/express-gateway.json
@@ -8,10 +8,10 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".algoliaLv0",
-    "lvl1": ".algoliaLv1",
-    "lvl2": ".algoliaLv2",
-    "lvl3": "section h1",
+    "lvl0": "section h1",
+    "lvl1": "section h2",
+    "lvl2": "section h3",
+    "lvl3": "section h4",
     "lvl4": "section h5",
     "text": "section p, section li"
   },


XVincentX:
So just to make sure — `lv0` is the most specific one and `lvn` the less specific one?

s-pace:
👋 @XVincentX,

It is the other way around. lvl0 is the least specific while lvl6 is the most. It is likewise <h1/> markups.

XVincentX:
@s-pace I've checked the selectors on the page and they are indeed behaving in this way. Can you point out where's that wrong?

XVincentX:
@s-pace Ping!

s-pace:
@XVincentX 

Sorry, it was buried by the notifications.
image
[example here](https://www.express-gateway.io/docs/configuration/system.config.yml/plugins/#markdown)

You have put the elements in the wrong order. We expect something like:
```html
<p class="algoliaLv0">...</p>
<p class="algoliaLv1">...</p>
<p class="algoliaLv2">...</p>
```

` lvl0 is the least specific while lvl6 is the most."

I spoted some pages were it was the other way arround

XVincentX:
@s-pace Is the order of the element really important?

The reason why this is happening is because of the (poor) algorithm that's scanning our web pages. The order might be wrong, but the levels are assigned correctly.

Is your crawler dependent on the element order?

s-pace:
Yes the order matter this is why I revert it. [reference here](https://community.algolia.com/docsearch/documentation/docsearch/how-it-works/#building-your-index)

XVincentX:
Oh gosh ok — will fix that.

s-pace:
Could you also explain me why you said that our algorithm was *poor* ?

s-pace:
Alright then. Thank you for your feedback @XVincentX 

I will try to tailor your next PR and let you know if everything is going fine :D

XVincentX:
@s-pace I've modified the page template so the hidden elements will always be in order.

s-pace:
👍 Deployed and live ✅  @XVincentX

XVincentX:
@s-pace Do you know when I can test the index?

s-pace:
On your website :). It is live already

XVincentX:
Whoooa! I just tested it and the experience is *WAY* better than the previous one!

Thank you so much!

s-pace:
Prego 🇮🇹

PR #346, opened 2018-03-16
Reconfigure Apollo to work across all docs repos.

abernix:
I'd appreciate some input on this!

Most all of the Apollo docs repositories now use the same format and I'm
hoping this will allow us to re-enable search on all of them, with a simple
configuration that will automatically expand as we deploy more properties.

A partial list of docs which would use this new configuration can be found at [this URL](https://github.com/meteor/hexo-theme-meteor#known-deployments) (paying attention to only observe the Apollo GraphQL links, not the Meteor links.

It's possible that some of these haven't been updated yet, but they will be soon!

### DIFF
--- a/configs/apollodata.json
+++ b/configs/apollodata.json
@@ -19,8 +19,7 @@
       "text": ".content-wrapper p, .content-wrapper li, .header-wrapper div.subtitle-page"
     }
   },
-  "only_content_level": true,
-  "min_indexed_level": 1,
+  "min_indexed_level": 2,
   "conversation_id": [
     "355172644"
   ],
--- a/configs/apollodata.json
+++ b/configs/apollodata.json
@@ -24,5 +24,5 @@
   "conversation_id": [
     "355172644"
   ],
-  "nb_hits": 2420
+  "nb_hits": 3623
 }
--- a/configs/apollodata.json
+++ b/configs/apollodata.json
@@ -4,7 +4,7 @@
     "https://www.apollographql.com/docs/"
   ],
   "stop_urls": [
-    "#"
+    "index\\.html"
   ],
   "selectors": {
     "default": {
--- a/configs/apollodata.json
+++ b/configs/apollodata.json
@@ -23,5 +23,5 @@
   "conversation_id": [
     "355172644"
   ],
-  "nb_hits": 3623
+  "nb_hits": 3263
 }
--- a/configs/apollodata.json
+++ b/configs/apollodata.json
@@ -4,17 +4,23 @@
     "https://www.apollographql.com/docs/"
   ],
   "stop_urls": [
-    "index\\.html"
+    "#"
   ],
   "selectors": {
-    "lvl0": ".content h1",
-    "lvl1": ".content h2",
-    "lvl2": ".content h3",
-    "lvl3": ".content h4",
-    "lvl4": ".content h5",
-    "text": ".content p, .content li"
+    "default": {
+      "lvl0": {
+        "selector": "div.sidebar span.title-sidebar",
+        "default_value": "Documentation"
+      },
+      "lvl1": ".header-wrapper h1",
+      "lvl2": ".content-wrapper h2",
+      "lvl3": ".content-wrapper h3",
+      "lvl4": ".content-wrapper h4",
+      "text": ".content-wrapper p, .content-wrapper li, .header-wrapper div.subtitle-page"
+    }
   },
-  "min_indexed_level": 2,
+  "only_content_level": true,
+  "min_indexed_level": 1,
   "conversation_id": [
     "355172644"
   ],


configs/apollodata.json
```diff
@@ -4,17 +4,23 @@
     "https://www.apollographql.com/docs/"
   ],
   "stop_urls": [
-    "index\\.html"
+    "#"
```
abernix: I decided to use this based on a similar configuration I saw on the `meteor` configuration, but I'm not positive if this is correct or not (or even necessary!).

configs/apollodata.json
```diff
@@ -4,17 +4,23 @@
...
+      },
+      "lvl1": ".header-wrapper h1",
+      "lvl2": ".content-wrapper h2",
+      "lvl3": ".content-wrapper h3",
+      "lvl4": ".content-wrapper h4",
+      "text": ".content-wrapper p, .content-wrapper li, .header-wrapper div.subtitle-page"
+    }
   },
-  "min_indexed_level": 2,
+  "only_content_level": true,
```
abernix: Similar to my previous comment, I'm not sure if this is necessary and can't seem to find the documentation for it, but this might not be necessary (or might be covered by `min_indexed_level` below?)  Some explanation of what this does would be appreciated!

configs/apollodata.json
```diff
@@ -4,17 +4,23 @@
...
   "selectors": {
-    "lvl0": ".content h1",
-    "lvl1": ".content h2",
-    "lvl2": ".content h3",
-    "lvl3": ".content h4",
-    "lvl4": ".content h5",
-    "text": ".content p, .content li"
+    "default": {
+      "lvl0": {
+        "selector": "div.sidebar span.title-sidebar",
```
abernix: The goal here is to have each docs repository automatically appear in a group based on its title.  The title is available on the sidebar, inside the `title-sidebar` `span` tag.

configs/apollodata.json
s-pace: `index\\.html` was there because you might have pages available under two deferent URLs. One with the trailing `index.html`, the other one without.
What are you aiming with this `#`?

configs/apollodata.json
s-pace: Documentation regarding this parameter are available here https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/#only_content_level-optional though

If you use this parameter, you don't need to set `min_indexed_level`

abernix: Does `only_content_level` === `min_indexed_level: 1` (specifically "1")?

configs/apollodata.json
abernix: I copied this from https://github.com/algolia/docsearch-configs/blob/master/configs/meteor.json#L27, which is a _similar_ theme and configuration.  Happy to adjust this accordingly, as you think is appropriate.

With what you're saying though, having `index\\.html` seems to make sense for both this configuration and `meteor`?

configs/apollodata.json
s-pace: Nope, it override completely the behaviour of it. It tells the crawler to only populate to algolia records built with a content not empty. This content is the text matched by the `text` selector

configs/apollodata.json
s-pace: This "#" has no use anymore. Must be because of the way of rendering. [Removed here](https://github.com/algolia/docsearch-configs/commit/c4685752b67f86f16f48a9126766d30b752c1fe6).

Yep it makes sense if you can't avoid `index.html` to be available (200).

s-pace:
Let's try it

abernix:
Yes, I was able to successfully run the `docsearch-scraper`!  It was actually very easy and painless process.  Thanks for that!

s-pace:
👍 @abernix 

Everything is live.

You may want to update both the scrapper and you includes:
```html
```

abernix:
Thanks, @s-pace.  Looks good!

One last question, hopefully: I'd like to give a higher ranking to items which are matching at the same `lvl0` as the page the search is currently being executed from.  So, for example, if you search for "Graph" on this [GraphQL Tools](https://www.apollographql.com/docs/graphql-tools/) page, your first results should be from the same "GraphQL Tools" section first, _followed by_ all the other results as usual.  Likewise, on [Apollo Server](https://www.apollographql.com/docs/apollo-server/), the "Apollo Server" results should show first.

I thought this might be possible with an `optionalFilters` inside `algoliaOptions`, but I haven't been able to figure it out:

```js
docsearch({
  // other options indexName, apiKey, etc....
  algoliaOptions: {
    optionalFilters: ["hierarchy.lvl0:GraphQL Tools"]
  }
});
optionalFilters 
```

...or perhaps an easier way?  Here's an image demonstrating what I'm asking, from the https://www.apollographql.com/docs/graphql-tools/ page.

<a href="https://cl.ly/3C432s1k3p2e" target="_blank"></a>

s-pace:
👋 @abernix 

I saw your post but it slips my mind. Good you have tagged me.

Regarding your use case and that you want to use the Serach UI context, I would recommend you to leverage [the transformData method](https://community.algolia.com/docsearch/documentation/docsearch-autocomplete/configuring-the-search/#transformdata). It will reorder your `hits` and put first the ones from your webpage. This method will leverage  `window.location`

Let me know if you need more details.

Cheers

PR #348, opened 2018-03-20
Update cockroachlabs config

j-low:
__[description]:__
* restrict `<h1>` target to `.post-header`; restrict lower hierarchy selectors to target content in `.post-content` only
* ignore URLs with `see-also` anchor

# Pull request motivation(s)
We're seeing some unhelpful information being returned in hit content.  It seems to stem from two main sources: 

1. Content from `<li>` tags in the header container being indexed
2. Content from `<li>` tags in `See Also` section at end of page being indexed

See above.

indexing_issues

We would like to have more descriptive content summaries available in our search results. 

Bug

### DIFF
--- a/configs/cockroachlabs.json
+++ b/configs/cockroachlabs.json
@@ -20,12 +20,12 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": "#main-content .content-col h1",
-    "lvl1": "#main-content .content-col h2",
-    "lvl2": "#main-content .content-col h3",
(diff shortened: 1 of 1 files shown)


s-pace:
FYI, something that you may need at some point [selectors_exclude](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/#selectors_exclude-optional)

s-pace:
Live and deployed

j-low:
@s-pace: Sounds good.  I'll try that route if this latest fix doesn't quite get us there.

PR #349, opened 2018-03-21
Update cockroachlabs.json to use "selectors_exclude" key

j-low:
__[description]:__
* update config to use "selectors_exclude" key to prevent indexing of "See Also" and "What's Next" sections at end of pages

# Pull request motivation(s)
Content in "See Also" and "What's Next" section is being indexed and returned in search results.  We wish to prevent this from happening.

See above.

See above.

Bug

### DIFF
--- a/configs/cockroachlabs.json
+++ b/configs/cockroachlabs.json
@@ -17,17 +17,15 @@
   ],
   "sitemap_urls": [
     "https://www.cockroachlabs.com/sitemap.xml"
-  ],
-  "stop_urls": [
-    "#see-also"
-  ],
+  ]
   "selectors": {
     "lvl0": "#main-content .post-header h1",
     "lvl1": "#main-content .post-content h2",
     "lvl2": "#main-content .post-content h3",
     "lvl3": "#main-content .post-content h4",
     "lvl4": "#main-content .post-content h5",
-    "text": "#main-content .post-content p, #main-content .post-content li"
+    "text": "#main-content .post-content p, #main-content .post-content li",
+    "selectors_exclude": "#see-also ~ ul, #whats-next ~ ul"
   },
   "only_content_level": true,
   "conversation_id": [
--- a/configs/cockroachlabs.json
+++ b/configs/cockroachlabs.json
@@ -17,7 +17,7 @@
   ],
   "sitemap_urls": [
     "https://www.cockroachlabs.com/sitemap.xml"
-  ]
+  ],
   "selectors": {
     "lvl0": "#main-content .post-header h1",
     "lvl1": "#main-content .post-content h2",


s-pace:
There is one mistake, I will fix it

j-low:
Great, thanks, @s-pace.

s-pace:
Please update your [sitemap](https://www.cockroachlabs.com/sitemap.xml) with your documentation content, we can't use it.

s-pace:
I would also recommend you to add a specific class to the element looking like : 
image
You will be able to add the class as a selector exclude.

I would also recommend you to wrap your element like `#whats-next` in order to avoid possible side effect of the general sibling combinator `~ `

j-low:
Re: sitemap: did something go out-of-date?  Is this just a matter of regenerating it and pushing the new sitemap live?

Re: adding classes: I agree this is ideal.  Currently all markup is generated from markdown documents. I'll need to do a bit of digging to automate the wrapping of these sections.  For now I think the side effects are minimal (if any).  In the near future I'll submit a new PR for our config once I figure out what needs to be done to automate the adding of a class and wrapping elements.  Ideally I'd like to avoid using the sibling combinator altogether.

s-pace:
Re sitemap: yes might be that. Should be compliant to [the reference](https://www.sitemaps.org/)

👌 for the rest 👍

j-low:
@s-pace so it looks like we use a separate sitemap for our docs content: https://www.cockroachlabs.com/docs/sitemap.xml

Can this be used?

s-pace:
Yes, updated!

j-low:
@s-pace we're getting a strange pattern of results for some of our queries.  The results list is returning results for pages which contain a search term (e.g. "selection clause"), but:

- `_highlightedResult.content.value` only contains the exact search term (might still be getting pulled from the "See Also" section)
- the result URL points to a section of the page that doesn't contain the search term

results

For the issue of `content` pulling from the "See Also" section, I'm guessing I can solve this by targeting the content more accurately.  

On the issue of the results pointing to an irrelevant section of the page I don't see anything in our config that is jumping out to me as a problem.  Do you have any thoughts on this one?

j-low:
@s-pace sure thing.  Right now it's in dev but we're going to be pushing to prod early today.  I'll ping you the URL once that's live.  Thanks again 👍

s-pace:
@j-low 👌 awesome

j-low:
@s-pace DocSearch is live: https://www.cockroachlabs.com/docs/stable/

s-pace:
Awesome thanks @j-low 

I notice that you didn't leverage the real-time feature of algolia.

Cheers

j-low:
@s-pace no we haven't yet.  Given time constraints and other projects we've opted to first get a feature-for-feature replacement of GSS in place and then evaluate implementations of more the advanced Algolia features.  I'm currently prototyping an implementation of the autocomplete feature and will likely do the same with instantsearch.

s-pace:
ok @j-low 

Keep us posted in order to really leverage tour DocSearch index. Having real-time definitely changes the search experience.

Cheers

PR #350, opened 2018-03-21
Update armorlabs.json

prstn:
# Pull request motivation(s)

This should include the code examples other missing documentation in the crawler's path.

Icon classes and code examples are not being included in the search results.

Updated selectors should better cover the developer examples.


s-pace:
We don't recommend to scrap code tags since it introduces a lot of noise but let's have a look

prstn:
@s-pace Thanks, I will keep an eye out for noise.

PR #351, opened 2018-03-22
Update start_urls on zoomdata.json

jonavila:
# Pull request motivation(s)


s-pace:
👋 @jonavila ,

Removing https://mystifying-roentgen-e8c480.netlify.com/docs/ since it is not available:
image

Cheers

PR #352, opened 2018-03-22
Prevent duplicated indexation

oliviertassinari:
# Pull request motivation(s)

capture d ecran 2018-03-23 a 00 15 25

We have the same content indexed multiple times:
- https://material-ui-next.com/customization/default-theme/
- https://material-ui-next.com/customization/default-theme/?expend-path=%24.breakpoints
- https://material-ui-next.com/customization/default-theme/?expend-path=%24.palette.secondary 
- https://material-ui-next.com/customization/default-theme/?expend-path=%24.palette.error
- …

I believe it's the right way to fix the issue following the documentation.
https://github.com/algolia/docsearch-configs#duplicated-content

### DIFF
--- a/configs/material-ui.json
+++ b/configs/material-ui.json
@@ -11,7 +11,7 @@
     "https://material-ui-next.com/discover-more/vision/",
     "https://material-ui-next.com/"
   ],
-  "stop_urls": [],
+  "stop_urls": ["/$"],
   "selectors": {
     "lvl0": {
       "selector": ".algolia-lvl0",


s-pace:
Might be better to stop on `?expend`, let's have a look!

s-pace:
Didn't work since most of the pages are stopped then

s-pace:
@oliviertassinari 

I have made some changes in order to enhance your search experience:
- Remove the duplicates (first reason of this PR) with avoiding URL containing `?expend`
- Simplify the config
- Encompass the first and last columns of your tables since you have some relevant content in those parts.
- Avoid having records without content matching the `text` selector (`only_content_level`)

Let us know if these changes suits you better. It is live.

oliviertassinari:
@s-pace This looks great! Thank you for taking the time to look at our configuration <3.

PR #354, opened 2018-03-25
update valr config

jayhesselberth:
# Pull request motivation(s)

Fine-tune searching for `valr` by adding `title` and `usage` selectors

### DIFF
--- a/configs/valr.json
+++ b/configs/valr.json
@@ -9,8 +9,8 @@
     "LICENSE.html$"
   ],
   "selectors": {
-    "lvl0": ".page-header h1",
-    "lvl1": ".contents h2",
+    "lvl0": ".title, .page-header h1",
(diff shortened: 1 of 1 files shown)


s-pace:
LGTM

s-pace:
@jankoenig ⚠️ 

Using the `.usage` class as lvl1 might be a myst&ke since it is long and look like text. (e.g. https://rnabioco.github.io/valr/reference/db.html) This is why I have moved it as `text`. We don't recommend do crawl `<code/>` elemnts since it may introduce a lot of noise.

Last but not least, could you wrap your `see-also` elements in one element `id=see-also` ? I have removed them from the scrap given that it is "external" content. Since we use a **direct** sibling combinator `+`, we do not remove all of them.

You can try it on your own, it is live and deployed.

PR #357, opened 2018-03-27
Add v7

brendt:
Add v7


s-pace:
Last but not least, you can update your front end if you want. We have a v2.

PR #359, opened 2018-03-28
Tweak wording for duplicated content

hadley:


s-pace:
@hadley thank you for spotting it.

I would even enhanced it:

```
It could happen that the crawling populates duplicated data from your website. This is mostly because we have crawled the same page several times (e.g. from different urls). If we have URLs like `http://website.com/page` and `http://website.com/page/` (notice the trailing `/` from the second one), the scraper will consider them as different. This can be fixed by adding a regex to the `stop_urls` in your `config.json`:
```

WDYT? If you like it, please commit it

PR #361, opened 2018-03-29
Update vite_labs.json

soliury:
edit stop_urls for doc.vite.org

# Pull request motivation(s)


s-pace:
👍 Deployed

PR #362, opened 2018-03-29
Update pkgdown config for new page structure

hadley:
This contains an updated spec based on discussion in https://github.com/r-lib/pkgdown/issues/608


s-pace:
Changes live. Let me know if it suits you bets!

PR #368, opened 2018-04-06
fix config for antora

mojavelinux:
I am the owner of docs.antora.org. I would like to request the following changes to my configuration so that the search results are accurate.

The main change is to the xpath. It was grabbing the title from the previous sibling of the current page from the navigation instead of the title for the current page. This page corrects that problem. I also made the selectors for the other levels and the text more accurate based on the structure of the content Antora produces.

- update xpath to grab correct lvl0 title from navigation tree
- query for sections using .doc instead of article
- grab some additional text elements
- change default value for lvl0 to Home


mojavelinux:
Thanks for merging, @s-pace! Now that I better understand how the levels work, could you merge this one too? https://github.com/algolia/docsearch-configs/pull/370 Thanks!

PR #369, opened 2018-04-06
Update quanteda.json

koheiw:
# Pull request motivation(s)

Make search results more informative

There are too may 'Unsorted' in the result

Some thing similar to the pkgdown website: http://pkgdown.r-lib.org/index.html

We have functions named like `tokens_select()` and `dfm_group()`. DocSearch seems to separate these strings into 'tokens' + 'select' and 'dfm' + 'group' unless quanteda. I don't think it should segment query by underscore.

Otherwise, its is great responsive search tool!

### DIFF
--- a/configs/quanteda.json
+++ b/configs/quanteda.json
@@ -8,20 +8,13 @@
     "LICENSE-text.html"
   ],
   "selectors": {
-    "lvl0": {
-      "selector": ".page-header h1",
-      "default_value": "No Header"
-    },
-    "lvl1": {
-      "selector": ".contents h1",
-      "default_value": "Unsorted"
-    },
-    "lvl2": ".contents h2",
-    "lvl3": ".contents h3, .contents th",
-    "lvl4": ".contents h4",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 

Thanks for the PR.

With this PR you will have the same number of Unsorted element since we replace the missing ones with the `default_value`.  With this PR you will miss `.page-header h1` which contains some meaningful information. You have some inconsistency between your pages. Some of them Could you add a specific class instead? like `.docSearch-lvl0`

Regarding you *NB2*, you can use the [algolia API parameter`separatorsToIndex`](https://www.algolia.com/doc/api-reference/api-parameters/separatorsToIndex/)  exposed by `custom_settings` like for the [improbable configs](https://github.com/algolia/docsearch-configs/blob/master/configs/improbable.json)

s-pace:
👋 @koheiw,

Please update your pkgdown set up and everything should work fine.

Closing it in the meantime, feel free to re open.

PR #375, opened 2018-04-11
update config param for doc search following the publication of the v4.1 of Boosted

Lausselloic:
Hello,

Following the publication of the documentation of our v4.1 which will coexist with the old v4.0 we need to update the index.

As I understand it will indexed both urls as http://boosted.orange.com/docs/4.0 and http://boosted.orange.com/docs/4.1.

There's still one question for me, how could I filter the search result from my website to serve the right content on each versions?

Thanks to you for your great job

### DIFF
--- a/configs/boosted-orange.json
+++ b/configs/boosted-orange.json
@@ -1,7 +1,13 @@
 {
   "index_name": "boosted-orange",
   "start_urls": [
-    "http://boosted.orange.com/docs/4.0/"
+    "http://boosted.orange.com/docs/(?P<version>.*?)/",
+    "variables": {
+      "version": [
+        "4.0",
+        "4.1"
+      ]
+    }
   ],
   "stop_urls": [],
   "selectors": {


s-pace:
👋 @Lausselloic 

Thank you for your feedback.

You will be able to refine the version thanks to [`facetFilters`](https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/).

See the code below:

```html
  apiKey: 'a2fb9f18ccc85658e152aeb2dd350860',
  indexName: 'boosted-orange',
});
</script>
```

- Replace $VERSION with the version you want to search on.
   So as of today you have: 4.0 (since your 4.1 version isn't live yet)

I have also updated your configuration in order to remove noise. Let me know if these changes suits you well.

I have also grant you access to the analytics we provide for your DocSearch index. Check the algolia dashboard.

Cheers

Lausselloic:
Thanks for your quick reply, I will put that option.

And take a look on the dashboard

s-pace:
Hi @Lausselloic ,

We needed your URLs no be available in order to update your index.

Changes are live with the latest version facet.

Cheers

Lausselloic:
Hello,

I think it was my pull request with param version who does the job? The new url is : http://boosted.orange.com/docs/4.1/ and previous one is till available http://boosted.orange.com/docs/4.0/

Loïc

s-pace:
Hello @Lausselloic 

Your PR did the job but at the time I have merged your PR and trigger the scrapping, the latest version wasn't available. Thus your DocSearch index wasn't updated with the 4.1 content. Since our scrapper is launched every 24h, your index wasn't already updated yet.

I have [run it again[(https://github.com/algolia/docsearch-configs/commit/0891fe8eef0272ae165bf6849ae57225c67796c1) and your index is now encompassing both version (`nb_hits` almost doubled)

If you check now, it is working.

Let us know if you need any other details.

Cheers

Lausselloic:
Yes all is now indexed wel. Thanks a lot

PR #383, opened 2018-04-13
set min indexed level to avoid naming collisions

TheSharpieOne:
# Pull request motivation(s)
Trying to avoid this (these are all separate pages with similar section titles): 
image

image

Have the component title appear on the left side, not just the word component over and over again.

implementation configuration change.

Nope

### DIFF
--- a/configs/reactstrap.json
+++ b/configs/reactstrap.json
@@ -19,6 +19,7 @@
   "selectors_exclude": [
     ".docs-example"
   ],
+  "min_indexed_level": 1,
   "conversation_id": [
     "555128868"
   ],


s-pace:
👌

TheSharpieOne:
@s-pace I just made a change to the semantic markup of the headers to better align the h1, h2, and h3s for accessibility (a page should not skip levels as ours did)
I didn't see `only_content_level` in the docs, but looking around at the issues and PR which mention it I think it would ignore `lvl`s which are empty?
With the change I made to our docs, now it would be something like this:
h1: Components
h2: Collapse
h3: Properties
I am hoping that change will address the issue, but I just made those changes so I need to wait for it to crawl before verifying.
Even with that change, do you think I still need `only_content_level` or to increase the level more?

Also, it would be awesome to have a way to preview/test configurations. Like a site where you can provide a configuration and some HTML and it will give a preview of how a search dropdown would look at some metadata about what it found. I can open an issue/feature request if that seems like something that is wanted.

s-pace:
👋 

the 📖 is [here](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/#only_content_level-optional). Thank you for sharing, I have enhanced the docs.

>  I just made a change to the semantic markup of the headers to better align the h1, h2, and h3s for accessibility (a page should not skip levels as ours did)
👍 

> `only_content_level`

When `only_content_level` is set to `true`, we only index builded records which match the `text` selectors. Every other record will be skipped. This parameter is more flexible than `min_indexed_level`. Once `only_content_level` is used, `min_indexed_level` becomes pointless.

If you have a record like this:
```
h1: Components
h2: Collapse
h3: Properties
```

It will not be indexed since it miss `text` content. We will create it for every element matching `text` after **h3: Properties**


> Also, it would be awesome to have a way to preview/test configurations. Like a site where you can provide a configuration and some HTML and it will give a preview of how a search dropdown would look at some metadata about what it found.

We recommend [to run it on your own first](https://community.algolia.com/docsearch/documentation/docsearch-scraper/overview/) in order to better understand it since it need a lot of iteration. Feel free to open an issue if you think it is relevant.

I have 🚀 your index. Check it. It is live

TheSharpieOne:
Thanks a ton.

This is what it looks like now. The component name is on the left so even though "Properties" is duplicated, it's more clear as it which one it is. This is what I was hoping to achieve 👍 
image

s-pace:
💯 

If you can avoid that kind of duplicates, just use `only_content_level`

Cheers

PR #387, opened 2018-04-17
Add adsum client api documentation search

moroine:
Hi,

It's another site configuration similar to the one I already have (adsum). 

Can I use the same api key than the one provided for the other website ? 

Thanks

### DIFF
--- a/configs/adsumClientApi.json
+++ b/configs/adsumClientApi.json
@@ -1,0 +1,24 @@
+{
+    "index_name": "adsumClientApi",
+    "start_urls": [
+        "https://adactivesas.github.io/adsum-client-api",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 
Why would you want to keep the same api key? An key is linked to an `index_name`.

I can enhance the former config and add this new link as a `start_urls` with [a specific tag on which you could filter](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/#start_urls-mandatory).

Let me know which solution you prefer.

moroine:
Actually I don't really care it's more about how do you prefer, or want do you advice ? 

If I can put all website on one single configuration file it's file to me. I just want to have a separate search for each website (I am using [https://docusaurus.io/](https://docusaurus.io/))

PR #390, opened 2018-04-19
Update definition to index our blog posts as well

XVincentX:
Hey,

we write a lot of good content on our blog. Therefore we thought that it makes sense to index its content as well.

Therefore I have modified the index in order to crawl all the pages we write except https://www.express-gateway.io/blog/ — that page is just the index blog page we're not interested in.

Thanks a lot as always for the great service!

### DIFF
--- a/configs/express-gateway.json
+++ b/configs/express-gateway.json
@@ -6,15 +6,18 @@
   "sitemap_urls": [
     "http://www.express-gateway.io/sitemap.xml"
   ],
+  "sitemap_urls_regexs": [
+    "https:\/\/www\.express-gateway\.io\/((?!\/blog\/).)*$"
+  ],
   "stop_urls": [],
   "selectors": {
     "lvl0": ".algoliaLv0",
-    "lvl1": ".algoliaLv1",
+    "lvl1": ".algoliaLv1, article h1",
     "lvl2": ".algoliaLv2",
     "lvl3": "section h1",
     "lvl4": "section h2",
     "lvl5": "section h3",
-    "text": "section p, section li"
+    "text": "section p, section li, article p"
   },
   "conversation_id": [
     "444614439"
--- a/configs/express-gateway.json
+++ b/configs/express-gateway.json
@@ -12,7 +12,7 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": ".algoliaLv0",
-    "lvl1": ".algoliaLv1, article h1",
+    "lvl1": ".algoliaLv1",
     "lvl2": ".algoliaLv2",
     "lvl3": "section h1",
     "lvl4": "section h2",


s-pace:
Sounds great

s-pace:
👋 @XVincentX 

Could you put your blog pages under `/blog/` pls?
`https://www.express-gateway.io/orchestrating-kubernetes-and-express-gateway/`
➡️ 
`https://www.express-gateway.io/blog/orchestrating-kubernetes-and-express-gateway/`

Could you wrap the external content in an element having a class like "docSearch-dont-index" ? It is introducing noise unreachable content.

Example of external content:
image

Keep us in the loop

XVincentX:
@s-pace

Couple of questions on my side:

1. Is it mandatory to change the blog pages location? As we've been doing different SEO initiatives changing the current content would be kind of difficult.

2. Isn't your crawler only indexing the pages listed in our `sitemap.xml` file? These links are not listed there.

Thanks!

s-pace:
> Is it mandatory to change the blog pages location? As we've been doing different SEO initiatives changing the current content would be kind of difficult.

You can add a meta. We will then be able to facet it in order to refine the scope of the search. [like tags](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/#start_urlstags)

> Isn't your crawler only indexing the pages listed in our sitemap.xml file? These links are not listed there.

We do crawl too so if a page is referenced from another and that it matches the `start_urls`, it will be scrapped

XVincentX:
@s-pace 

> You can add a meta. We will then be able to facet it in order to refine the scope of the search. 

Would it be possible to defer this step? We'd like to gather results from people seeing blog posts as part of search result and see if it's worth such change.

> We do crawl too so if a page is referenced from another and that it matches the start_urls, it will be scrapped

I'm not completely sure I follow here. If I understand correctly what you wrote — the URL must point anything that matches the `start_urls` pattern, or they'll be discarded.

If that's the case, we should be safe, as all our links in the **More resources** part point to a different website: Github, MailChimp, Twitter and Gitter.

s-pace:
> Would it be possible to defer this step? We'd like to gather results from people seeing blog posts as part of search result and see if it's worth such change.

Sure

> I'm not completely sure I follow here. If I understand correctly what you wrote — the URL must point anything that matches the start_urls pattern, or they'll be discarded

The URL must match `start_urls` and be referenced from another matching page or from a sitemap.

XVincentX:
> The URL must match start_urls and be referenced from another matching page or from a sitemap.

Ok — so these links then would not be indexed because they do not match the start url, right?

In this case then, so why is that placeholder creating a problem?

s-pace:
Which links?

The placeholder is introducing some noise

XVincentX:
Hmm 🤔 

Is there any way I can followup in private/chat? I think it's be better in real time 😀

s-pace:
Use the email conversation you have used for applying

XVincentX:
Ok @s-pace 

I've extracted all the "More Resource" sections as a template and put the css class you mentioned there. You can see the result here: https://github.com/ExpressGateway/express-gateway.io/pull/231/files#diff-663f387b6a1a407ab38de055a12bc7c8R21

I understood only now you were referring to the text content creating noisy indexes; I thought you were referring to links, please apologize.

s-pace:
💯 
PR the new change and 🎉

PR #393, opened 2018-04-23
Added selectors to index code and table bodies

samueller:
I have terms I want to index within `<table>` bodies and within `<code>` elements.

### DIFF
--- a/configs/udocs.json
+++ b/configs/udocs.json
@@ -16,7 +16,7 @@
     "lvl2": ".post h2",
     "lvl3": ".post h3",
     "lvl4": ".post h4",
-    "text": ".post article p, .post article li"
+    "text": ".post article p, .post article li, .post article tbody td, .post article code"
   },
   "selectors_exclude": [
     ".hash-link"
--- a/configs/udocs.json
+++ b/configs/udocs.json
@@ -15,8 +15,8 @@
     "lvl1": ".post h1",
     "lvl2": ".post h2",
     "lvl3": ".post h3",
-    "lvl4": ".post h4",
-    "text": ".post article p, .post article li, .post article tbody td, .post article code"
+    "lvl4": ".post h4, .post article tbody td:first-of-type",
+    "text": ".post article p, .post article li, .post article tbody td:nth-child(2)"
   },
   "selectors_exclude": [
     ".hash-link"


s-pace:
👋 

Thanks for the PR.

We do not recommend to index `<code/>` elements since it will introduce a lot of noise.

As for table, I would recommend to follow this pattern:

``json
  "selectors": {
    "lvl0": "FIXME h1",
    "lvl1": "FIXME h2",
    "lvl2": "FIXME h3",
    "lvl3": "FIXME h4",
    "lvl4": "FIXME h5, FIXME table td:first-of-type",
    "text": "FIXME p, FIXME li, FIXME table td:last-child"
  }
```

samueller:
Thank you for the comments. Why the `:first-of-type` and `:last-child` pseudo-classes? As an example, [our page on the JavaScript Math object](https://udocs.com/docs/math_object.html) has a table of constants. I'd like people to be able to search for "Math.E", "Euler", or even "2.718281828459045" and bring up that page.

Should I include the table headings as a `lvl4` selector? I'm not sure if that makes sense in the context of Algolia's search mechanism.

samueller:
Makes sense about the table headers being something obvious, that's why I asked if it makes sense in this context.

Wouldn't `table td:nth-child(2)` only get the 2nd cell in a table row? Don't I want all cells in a table row?

s-pace:
yes. The first one will be a title

samueller:
In my tables, the first cell is not the title necessarily, all cells are indexable, like the example I gave above for JavaScript’s Math object

s-pace:
👋 @samueller 

The example suits perfectly what I try to say. The first column is related to the second one since the latter explain the former. Having these selectors make sense.

Otherwise feel free to add the a class like `.docSearch-index`

Cheers

samueller:
Oh! I get what you're saying. So `lvl4` would be `table td:first-of-type` while `text` would be `table td:nth-child(2)`. Yes, that makes sense!

samueller:
I made it `tbody` instead of `table` so that the table headers wouldn't be included

s-pace:
No need of `tbody` given that [`td` are data cell only](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td)

✅ I will merge it then

PR #394, opened 2018-04-23
Fixed selector for text (ezplatform doc)

damianz5:
After some tests we would like to add table content to text selector

### DIFF
--- a/configs/ezplatform.json
+++ b/configs/ezplatform.json
@@ -30,7 +30,7 @@
     "lvl4": ".md-content h4",
     "lvl5": ".md-content h5",
     "lvl6": ".md-content h6",
-    "text": ".md-content p, .md-content ul li, .md-content .code pre"
+    "text": ".md-content p, .md-content ul li, .md-content .code pre, .md-content table tr td"
   },
   "conversation_id": [
     "549525560"


s-pace:
👋 @damianz5 

⚠️ It will introduce a lot of noise:
image
Solution:
```"selectors_exclude":[".code"]```

I would recommend to use something like
```json
  "selectors": {
    "lvl0": "FIXME h1",
    "lvl1": "FIXME h2",
    "lvl2": "FIXME h3",
    "lvl3": "FIXME h4",
    "lvl4": "FIXME h5, FIXME table td:first-of-type",
    "text": "FIXME p, FIXME li, FIXME table td:last-child"
  }
```
or to add a `docSearch-index` class

SylvainGuittard:
Hi @s-pace !
Thanks for the review. 

IMHO, we should index the code blocks, because the documentation on https://doc.ezplatform.com is developer oriented. It would be nice to search for a namespace or a function.

s-pace:
I can merge it but big disclaimer here

damianz5:
I will create a new test instance with this change and we will compare the results

s-pace:
Good idea @damianz5

damianz5:
Results are better with those changes (content of the tables), please merge
ping @SylvainGuittard

s-pace:
👋 @damianz5 cc @SylvainGuittard 

It looks like your selector `lvl0` doesn't match anything. What would like to pick up?

(Deployed and live)

PR #396, opened 2018-04-23
Index tuturials

DirtyF:
# Pull request motivation(s)

Index https://jekyllrb.com/tutorials/ and related subsections

Users can't currently search content in the tuturials section.
For instance, when you type "bundler", the tutorial on using jekyll with bundler is not returned

Content from the tutorials is displayed in the search results.


s-pace:
👏 live 📺

DirtyF:
Thanks for the quick reindex @s-pace 🚀

PR #400, opened 2018-04-24
Fix Sourcegraph

attfarhan:
We recently changed the CSS selector for our docs pages, this PR addresses that change.

Also adds `td` elements as text, in order to index content in the table on this page: https://about.sourcegraph.com/docs/search/query-syntax/.  I think the change is correct, but let me know if otherwise! 

Thanks!


s-pace:
🙏 @attfarhan it was what we notice

PR #407, opened 2018-04-30
Update uniwebview.json

onevcat:
Hi,

We recreated our doc page for UniWebView: http://docs.uniwebview.com/

I am not quite sure about the config file since initially, some Algolia staff helped me to config the site (thank you a lot by the way!)

Previously it was not a static page, viewers have to render it in the browser (so the js render was used). Now it is a pure static website so I guess I need to change something. I read the recommendation and tried to update our config. It would be also great if you could check it since you are experts on this. :P

There is some tricky implementation on the API reference (like on this page http://docs.uniwebview.com/api/uniwebviewmessage.html)

snip20180430_1

The `.api-anchor` is used to determine the navigation position and the `.api-heading` is where the DOM of the documentation entry (item) actually is. Is it possible to write a config file for this case, to make the search result correct and the navigating also going to right place?

Thank you!

### DIFF
--- a/configs/uniwebview.json
+++ b/configs/uniwebview.json
@@ -1,27 +1,21 @@
 {
   "index_name": "uniwebview",
   "start_urls": [
-    "http://docs.uniwebview.com/#/"
-  ],
-  "stop_urls": [
-    "id=",
-    "/archived/"
+    "http://docs.uniwebview.com"
   ],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


maxiloc:
Thanks @onevcat. I merge and update if needed.
I'll deploy after that.

onevcat:
@maxiloc Thank you!

PR #409, opened 2018-05-02
update mdanalysis.org search config

orbeckst:
# Pull request motivation(s)

Improve the relevance of the index for the site https://www.mdanalysis.org

- do not index development docs (https://www.mdanalysis.org/mdanalysis) or paginated blog
- use sitemap (probably now needed to find all blog posts now that pagination has been removed); however, the sitemap currently only includes the content generated by jekyll (using the jekyll-sitemap plugin), not the sub-sites that are accessible by crawling. I am reading the ["How it works" docs](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/) as if Algolia will crawl starting from the pages in the sitemap so I hope this will still index most of our site.
- closes MDAnalysis/MDAnalysis.github.io#77

Content shows up duplicated or near duplicated (and the [Recommendations](https://community.algolia.com/docsearch/documentation/docsearch/recommendations/) rightly advise against such duplication).

Users only expect to see a single relevant topic of the same name.

Update to our indexer config.

We now provide a sitemap.xml and we hope that the indexer will simply start crawling from there. If not then we have to come up with additional ways to create more sitemaps.

### DIFF
--- a/configs/mdanalysis.json
+++ b/configs/mdanalysis.json
@@ -1,13 +1,17 @@
 {
   "index_name": "mdanalysis",
+  "sitemap_urls": [
+    "https://www.mdanalysis.org/sitemap.xml",
+  ]
   "start_urls": [
     "https://www.mdanalysis.org"
   ],
   "stop_urls": [
     "https://www.mdanalysis.org/.*?//.*?",
-    "https://www.mdanalysis.org/mdanalysis/_modules",
+    "https://www.mdanalysis.org/mdanalysis",
     "https://www.mdanalysis.org/docs/_modules/",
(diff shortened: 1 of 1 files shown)


maxiloc:

orbeckst:
Thanks @maxiloc – I assume I can expect to see changes after ~24h when your index has been rebuilt?

maxiloc:
I deployed it manually, so it launch a reindex at the same time.
Should be live now

orbeckst:
Yes, just checked again and it looks as if the changes took place. Thanks again!

PR #411, opened 2018-05-03
Update zoomdata.json with new URLs

jonavila:
# Pull request motivation(s)


maxiloc:

PR #412, opened 2018-05-03
Update crunch.json

nealrichardson:
This file is also available at https://crunch.io/r/crunch/docsearch.json


maxiloc:
Thanks I'll deploy right away

PR #413, opened 2018-05-03
update embark.json

jrainville:
Update to real site url plus add tutorials and plugins pages


maxiloc:
Thanks. I'll deploy that right away

PR #376, opened 2018-04-11
updated jss host name and text configs

sitecorian:
updated jss host name and text (testing with new version of the site)

PR #415, opened 2018-05-07
Updated JSS config

kamsar:
* We enabled SSR so no need for JS rendering (this should also resolve concerns from https://github.com/algolia/docsearch-configs/pull/376)
* No longer using hash routing
* Sending HTTP 404 so presumably no need for stop content


maxiloc:
Thanks I'll deploy that right away

kamsar:
Looks like it works great; thanks @maxiloc 👍

PR #419, opened 2018-05-14
Update the riot domain

GianlucaGuarini:
# Pull request motivation(s)
We have updated our documentation domain
see https://github.com/riot/riot.github.io/issues/207
and https://github.com/riot/riot.github.io/issues/212

The search links still to the old domain

The search should link to the newest domain


s-pace:
👋 updated and tailored @GianlucaGuarini 
Let us know if you need anything!

PR #420, opened 2018-05-15
Update hackerone search config due to URL change and add sitemap

reedloden:
We launched our new docs site, so need to update the URL to be https://docs.hackerone.com/.

Also, we have a sitemap now, so use it to make the scraper more efficient. :-)

### DIFF
--- a/configs/hackerone.json
+++ b/configs/hackerone.json
@@ -1,7 +1,10 @@
 {
   "index_name": "hackerone",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


s-pace:
👋@reedloden

Update and tailored. Let me know what do you think about that.

Btw, if you ad `id` or `name` attribute to your `lvlX` elements, the redirection will be enhanced (scroll down automatically thanks to the anchors)

Cheers

PR #423, opened 2018-05-22
Update pilosa

codysoyland:
# Pull request motivation(s)

Update pilosa config to include latest documentation version.


s-pace:
👋 @codysoyland 

Looks like your `v0.10` version pages aren't available yet. Is it expected?

codysoyland:
Hi @s-pace, thanks for responding so quickly. I did things a bit out of order, so the website was *just* updated.

Thanks!
Cody

s-pace:
You are welcome,
Makes sense, just ping here when it is ready, I will update it accordingly (no rush, we will be noticed of the change of `nb_hits` otherwise)

s-pace:
✅ 👍

PR #424, opened 2018-05-24
setup multiple version crawl for 'gauge'

sriv:
# Pull request motivation(s)
docs.gauge.org is set to crawl only one version (latest) at the moment. This PR enables crawling of `latest` and `master` version of the site.

The intention is to compartmentalize the search to the version context. 

The use case is that version searches are independant of each other.

Only crawls root, which is the latest version.

Should crawl both `latest` and `master` versions.

I assume that with this PR (pending review/corrections) I should be able to let my users search in one version (i.e. the version that they invoke the search from). Is this correct?

### DIFF
--- a/configs/gauge.json
+++ b/configs/gauge.json
@@ -1,15 +1,15 @@
 {
   "index_name": "gauge",
   "start_urls": [
-    "https://docs.gauge.org"
+    "https://docs.gauge.org/latest/",
+    "https://docs.gauge.org/master/"
   ],
   "stop_urls": [
-    "https://docs.gauge.org/0",
-    "https://docs.gauge.org/master/"
+    "https://docs.gauge.org/0"
   ],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @sriv 

I have tailored it a bit in order to make it as simple as possible. You will be able to use our facet feature:

You're now a few steps away from having it working on your website:

```html

  apiKey: '5008b0d9ea4d9e639a17a123bea077fe',
  indexName: 'gauge',
});
</script>
```

   So as of today you have: `latest`, `master`

  For example if you want to refine the search to the version `"latest"` just specify: ```js
'facetFilters': ["version:latest"]```

Let me know if you need anything else.

Cheers

sriv:
@s-pace cheers for the correction, and the detailed instructions. You folks are awesome!

I have deployed the changes and can see them work very well.

s-pace:
Thank you very much. Glad to help :)

PR #427, opened 2018-05-25
Update nexenta.json

antonfisher:
Hello @s-pace,

There was identical "start_urls" configuration for two different websites: https://github.com/algolia/docsearch-configs/blob/master/configs/nexenta.json and https://github.com/algolia/docsearch-configs/blob/master/configs/nexentaedge.json, that causes to the same search indexes for both websites.
Websites we have:
- https://nexentaedge.io/ - has right configuration in `nexentaedge.json` file
- https://nexenta.github.io - another community website, this PR fixes "start_urls" configuration for this site in `nexenta.json` file.

Thank you!

### DIFF
--- a/configs/nexenta.json
+++ b/configs/nexenta.json
@@ -1,8 +1,8 @@
 {
   "index_name": "nexenta",
   "start_urls": [
-    "https://nexentaedge.io/docs/",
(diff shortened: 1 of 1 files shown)


s-pace:
💯

PR #428, opened 2018-05-27
Update vite_labs.json

soliury:


soliury:
Can someone merge it into master?

configs/vite_labs.json
```diff
@@ -1,32 +1,34 @@
 {
   "index_name": "vite_labs",
   "start_urls": [
+    {
+      "url": "http://doc.vite.org/",
+      "variables": {
+        "lang": [
+          "en"
```
s-pace: These webpages will not have the right variables since it isn't injected somewhere

configs/vite_labs.json
```diff
@@ -1,32 +1,34 @@
...
       }
     }
   ],
-  "stop_urls": [
-    "/about/?$",
-    "/faq/?$",
-    "/tech/?$",
-    "/whitePaper/?$"
-  ],
+  "stop_urls": [],
```
s-pace: These values are set on purpose in order to restrain the crawl to your documentation content only. Please put it back

configs/vite_labs.json
s-pace: You can use tags instead.

configs/vite_labs.json
soliury: I know, I have reset my site, so the `stop_url` need to change.

configs/vite_labs.json
soliury: ok, I see.

s-pace: The need to add a new language to the site is something required.
Otherwise we can inject it from another webpage but I would rather have it written properly.

soliury: Ok, I see

s-pace:
@soliury ✅

PR #433, opened 2018-06-04
Update pm2docs.json

robin-monnier:
# Pull request motivation(s)

### DIFF
--- a/configs/pm2docs.json
+++ b/configs/pm2docs.json
@@ -33,10 +33,9 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
-      "selector": "//div[contains(@class, 'topBlock')]/p/span[1]",
-      "type": "xpath",
+      "selector": "nav.header-links a.active",
       "global": true,
-      "default_value": "Documentation"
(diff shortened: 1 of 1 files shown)


robin-monnier:
I don't really understand the purpose of "global" option, explanations [here](https://community.algolia.com/docsearch/documentation/docsearch-scraper/config-options/) are not sufficient.

s-pace:
👋 @robin-monnier 

`global` attribute should be seen as a way to extract the matching elements from the HTML flow since we rely on this flow in order to [build records from your pages](https://community.algolia.com/docsearch/documentation/docsearch/how-it-works/#building-your-index). These `global` elements will not be considered as breaking ones when we encounter different ones along the flow. If this token is enabled :
- we will not create a new record from the current building stack if we encounter a new `global` element.
- we will apply its contextual value ( or `default_value` if not matched) to every records

We mostly use this parameter for page that miss context. It enable us to pick up the context from another common part without having duplicates since this context might be shared from one page to another.

Hope it is clear enough, happy to ear your feedback and discuss about what is your use case.

PR #434, opened 2018-06-04
update endiliey config

endiliey:
# Pull request motivation(s)
Update my website config to allow blog to be crawled instead of just docs

### DIFF
--- a/configs/endiliey.json
+++ b/configs/endiliey.json
@@ -22,6 +22,12 @@
           "zh-CN"
         ]
       }
+    },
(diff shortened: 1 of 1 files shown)


endiliey:
@s-pace thank you for the clarification. I was thinking of settling the infra first so I can write my content later on without worrying about the configuration since it's re-crawled every 24 hours.
I guess that can wait later.

s-pace:
Nice @endiliey 

Happy yo review it asap :)

PR #435, opened 2018-06-06
Changed lvl 0 to be the same for all posts.

vojtechruz:
changed lvl 0 to be the same for all posts.

Currently, my results look like this, where each article has its own section:
clipboard02

And I want them all to be in the same category.

clipboard02d

Feel free to edit if I didn't implement correctly


s-pace:
👋 @vojtechruz 

Updated and deployed.

vojtechruz:
Thank you 👍

PR #437, opened 2018-06-06
Change url in the docusaurus demo config

JoelMarcey:
# Pull request motivation(s)

The URLs pointed to the live Docusaurus site for this config. Just changing it back to the demo site.

### DIFF
--- a/configs/docusaurus_demo.json
+++ b/configs/docusaurus_demo.json
@@ -2,7 +2,7 @@
   "index_name": "docusaurus_demo",
   "start_urls": [
     {
-      "url": "https://docusaurus.io/docs/(?P<lang>.*?)/(?P<version>.*?)/",
+      "url": "https://joelmarcey.github.io/docusaurus-demo/docs/(?P<lang>.*?)/(?P<version>.*?)/",
       "variables": {
         "lang": [
           "en"
@@ -13,7 +13,7 @@
       }
     },
     {
-      "url": "https://docusaurus.io/docs/(?P<lang>.*?)/",
+      "url": "https://joelmarcey.github.io/docusaurus-demo/docs/(?P<lang>.*?)/",
       "variables": {
         "lang": [
           "en"
@@ -26,7 +26,7 @@
       }
     },
     {
-      "url": "https://docusaurus.io/blog/",
+      "url": "https://joelmarcey.github.io/docusaurus-demo/blog/",
       "tags": [
         "blog"
       ]
@@ -37,7 +37,7 @@
     "users"
   ],
   "sitemap_urls": [
-    "https://docusaurus.io/sitemap.xml"
+    "https://joelmarcey.github.io/docusaurus-demo/sitemap.xml"
   ],
   "selectors": {
     "lvl0": {
--- a/configs/docusaurus_demo.json
+++ b/configs/docusaurus_demo.json
@@ -56,7 +56,7 @@
     ".hash-link"
   ],
   "conversation_id": [
-    "415526327"
+    "516502774"
   ],
   "nb_hits": 1014
 }


configs/docusaurus_demo.json
```diff
@@ -59,4 +59,4 @@
     "415526327"
```
JoelMarcey: I wasn't sure if this needed to be changed -- is is docusuarus specific, opposed to docusaurus-demo?

configs/docusaurus_demo.json
```diff
@@ -59,4 +59,4 @@
...
   ],
   "nb_hits": 1014
```
JoelMarcey: This too? Does this need to be changed to for docusaurus_demo?

configs/docusaurus_demo.json
s-pace: It is docusaurus specific. You do not need to change it.

configs/docusaurus_demo.json
s-pace: It will be changed automatically based on how many hits this PR will introduce

s-pace:
👋 @JoelMarcey 

Thank you for the update.

I have some feedback:

- It looks like [your sitemap isn't so far updated](https://joelmarcey.github.io/docusaurus-demo/sitemap.xml)

- There is no languages for this website so I have removed it.

- Could you do a redirection from the general doc root to one of the documentation pages pls ?:
```https://joelmarcey.github.io/docusaurus-demo/docs/ -> https://joelmarcey.github.io/docusaurus-demo/docs/doc1.html``` (this stands for Docusaurus too)


I have updated it and I would be happy to receive any of your feedback regarding DocSearch.

**Comment by JoelMarcey** [comment]
Hi @s-pace -- since this is a demo, I am keeping the site pretty simple with just sample docs. I just updated the site to Docusaurus 1.2.1, and the sitemap should have been updated with it.

Do things look good from your end now?

**Comment by s-pace** [comment]
We need some small changes. I will do it today.

**Comment by s-pace** [comment]
👋 @JoelMarcey @endiliey 

In order to prevent stoping the Docusaurus search I have made a temproray index: [`docusarus_meta`](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus_meta.json). It looks fantastic. I will let you try it out:

```html

  apiKey: 'c53337307f1787a9199348cd8263f666',
  indexName: 'docusaurus_meta',
});
</script>
```

   So as of today you have: `1.0.11`, `1.0.12`, `1.0.13`, `1.0.14`, `1.0.15`, `1.1.0`, `1.1.1`, `1.1.2`, `1.1.3`, `1.1.4`, `1.1.5`, `1.2.0`, `1.2.1`, `next`

- Replace `$LANGUAGE` with the language you want to search on.
   The list of possible language is hardcoded in the config.
   So as of today you have: `en`, `es-ES`, `ro`, `tr`, `zh-CN`

  For example if you want to refine the search to the version "1.0.11" and the language "en" just specify:
```js
 'facetFilters': ["version:1.0.11", "language:en"]
```

Please note that you can define several values for every facets. It can be an array (Might be useful for `latest` for example)

Happy to give further details.

Have a nice day :)

endiliey:
@s-pace
Is this related to the docsearch metadata we added recently ? Seems to me that now searching docs versions & language is very refined now.

Should we try this in Docusaurus @joelmarcey ?
```
'facetFilters': ["version:VERSION", "language:LANGUAGE"]
```

s-pace:
Yes it is directly because of the meta. We use them in order to find the right context of the page. It is now a [facetFilters](https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/)

PR #438, opened 2018-06-06
Update Docusaurus search configs for sitemap and start urls

JoelMarcey:
Use https://github.com/algolia/docsearch-configs/commit/db60b72c31f58a08b6ad07ce8277d7d363baf161 for the actual Docusaurus site

### DIFF
--- a/configs/docusaurus.json
+++ b/configs/docusaurus.json
@@ -1,13 +1,43 @@
 {
   "index_name": "docusaurus",
   "start_urls": [
-    "https://docusaurus.io/docs/en/",
-    "https://docusaurus.io/en/",
-    "https://docusaurus.io/docs/en/installation.html"
+    {
+      "url": "https://docusaurus.io/docs/(?P<lang>.*?)/(?P<version>.*?)/",
+      "variables": {
+        "lang": [
+          "en"
+        ],
+        "version": [
+          "next"
+        ]
+      }
+    },
+    {
+      "url": "https://docusaurus.io/docs/(?P<lang>.*?)/",
+      "variables": {
+        "lang": [
+          "en"
+        ]
+      },
+      "extra_attributes": {
+        "version": [
+          "latest"
+        ]
+      }
+    },
+    {
+      "url": "https://docusaurus.io/blog/",
+      "tags": [
+        "blog"
+      ]
+    }
   ],
   "stop_urls": [
-    "https://docusaurus.io/docs/en/[0-9].*",
-    "https://docusaurus.io/docs/en/next.*"
+    "help",
+    "users"
+  ],
+  "sitemap_urls": [
+    "https://docusaurus.io/sitemap.xml"
   ],
   "selectors": {
     "lvl0": {
@@ -28,5 +58,5 @@
   "conversation_id": [
     "415526327"
   ],
-  "nb_hits": 563
+  "nb_hits": 1014
 }


s-pace:
👋 @JoelMarcey 

I had to put the `stop_urls` on the numeral version pages since they introduce a lot of duplicates. These pages are really similar
```json
"stop_urls": [
    "help",
    "users",
    "https://docusaurus.io/docs/en/[0-9].*"
  ],
```

Could it be possible to only index major versions?

___

The integration will change:
```html

  apiKey: '3eb9507824b8be89e7a199ecaa1a9d2c',
  indexName: 'docusaurus',
  algoliaOptions: { 'facetFilters': ["lang:$LANG", "version:$VERSION", "tags:$TAGS"] },
});
</script>
```

- Replace `$LANG` with the lang you want to search on.
   The list of possible lang is hardcoded in the config.
   So as of today you have: en

   So as of today you have: `latest`, `next`

- Replace `$TAGS` with the tags you want to search on.
   The list of possible tags is hardcoded in the config.
   So as of today you have: `blog`

s-pace:
(Changes are live, you can check it now)

JoelMarcey:
Hi @s-pace 👋 

Thanks for your help here.

Currently for Docusaurus, the default documentation is corresponds to the latest version of Docusaurus. So in our current case, it is `1.1.5`. Updates to the docs before we publish a new version are shown in `next`. As it stands, with the current Algolia search config, `next` is what is searched upon. `latest` (which would correspond to `1.1.5` now), can be searched upon. 

In https://docusaurus.io/en/versions.html, that corresponds to:

```
<h3 id="latest">Current version (Stable)</h3>
```

> Could it be possible to only index major versions?

So to answer your question, maybe, we only search `latest` since that is the default? And we don't even search `next`?

> Replace $TAGS with the tags you want to search on. The list of possible tags is hardcoded in the config. So as of today you have: `blog`

Should we also have `docs` as an option here too?

s-pace:
👋 @JoelMarcey 

Thank you for your feedback.

I have checked with the team and we can index every versions of the website https://docusaurus.io if you need it. Otherwise we can only do the latest one. Up to you and let me know. 

As for the other forked projects, we will firstly only crawl the latest version.

> Should we also have docs as an option here too?

Done. It is live.

In order to avoid duplicates, please leverage:
```js
'facetFilters': ["lang:$LANG", "version:$VERSION", "tags:$TAGS"]
``` 

We  have also noticed that pages from the latest version are available under URLs with or without trailing `.html`. Could it be possible to add redirection on these pages? I have prevent our tool to crawl pages without the trailing pages `.html` thanks to the regex `^((?!\\.html).)*$`

JoelMarcey:
Hi @s-pace 

Thanks for updating the config...

A couple of questions:

1.

Right now, on https://docusaurus.io, we are seeing duplicates (from the current version and `next`). Are you saying, if I add `'facetFilters': ["lang:$LANG", "version:$VERSION", "tags:$TAGS"]` exactly as is to our Algolia config, that would stop? I don't need to replace `$LANG` with anything?

2.

> Could it be possible to add redirection on these pages?

Redirection within the Algolia config? 

Or redirection in the core Docusaurus code/site? 

By default, all Docusaurus sites have `.html`. However, we provide an option to where you can have your URLs be clean without the `.html`. If that option is set, both URLs work. Are you saying we should update the Docusaurus core code is set that if the clean URL option is set, we should redirect to the clean url instead of allowing both?
 
Thanks!

s-pace:
👋 @JoelMarcey 

> 1

In order to leverage the version you can only use the right version face:
```js
'facetFilters': ["version:latest"]
```
This is the classic usage of [algolia parameter `facetFilters`](https://www.algolia.com/doc/api-reference/api-parameters/facetFilters/).

> 2

The redirection will be within your core website. Since two different URLs are available under HTTP 200,  we do scrap both of them (<url> and <url>.html).

If you are fine with the DocSearch Search UI redirecting to .html pages, you don't have to bother.

> By default, all Docusaurus sites have .html. However, we provide an option to where you can have your URLs be clean without the .html. If that option is set, both URLs work. Are you saying we should update the Docusaurus core code is set that if the clean URL option is set, we should redirect to the clean url instead of allowing both?

👍 You got it. Since we pick up the whole URL, we are redirecting to this one. Thank you for the insight.

Feel free to ask us anything

JoelMarcey:
👋 @s-pace 

Does https://github.com/facebook/Docusaurus/pull/744 seem reasonable and like the right way to go?

s-pace:
@JoelMarcey Yep you will only need to replace VERSION and LANGUAGE with the right one.

Merging it and I will comment facebook/Docusaurus#744

PR #441, opened 2018-06-07
enable https for xmake.io

waruqi:
# Pull request motivation(s)


waruqi:
Thanks! 😄

PR #447, opened 2018-06-09
enable more languages for react-reason & BuckleScript

endiliey:
## Motivation

Based on https://github.com/reasonml/reasonml.github.io/pull/392#issuecomment-395927699
We want ReasonReact & BuckleScript to search more than just one language.

cc @chenglou

### DIFF
--- a/configs/bucklescript.json
+++ b/configs/bucklescript.json
@@ -1,8 +1,38 @@
 {
   "index_name": "bucklescript",
   "start_urls": [
-    "https://bucklescript.github.io/docs/en/installation.html",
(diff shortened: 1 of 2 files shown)


s-pace:
👋 @endiliey 

💯 

Live and deployed.

Could it be possible to avoid having 404 on root pages like https://bucklescript.github.io/docs/en/?

endiliey:
@s-pace

It's possible, but that is best fixed in Docusaurus which powers bucklescript & reasonreact.

s-pace:
Makes sense @endiliey  indeed. Let's use the other thread in order to make it work
thanks

PR #448, opened 2018-06-11
Update pm2docs.json

robin-monnier:
# Pull request motivation(s)

### DIFF
--- a/configs/pm2docs.json
+++ b/configs/pm2docs.json
@@ -47,8 +47,8 @@
     "text": "#main p, #main li"
   },
   "selectors_exclude": [
-    "#next-step",
-    "#next-step + p",
+    "#next-steps",
+    "#next-steps + p",
     "#questions",
     "#questions + p",
     "#gitalk"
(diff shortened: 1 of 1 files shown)


configs/pm2docs.json
```diff
@@ -47,14 +47,14 @@
     "text": "#main p, #main li"
   },
   "selectors_exclude": [
-    "#next-step",
-    "#next-step + p",
+    "#next-steps",
+    "#next-steps  p",
```
s-pace: Don't you miss the ~ here?

configs/pm2docs.json
```diff
@@ -47,14 +47,14 @@
...
     "#questions",
-    "#questions + p",
+    "#questions ~ p",
```
s-pace: Warning, every p after #questions will be skipped. Not only the last one

robin-monnier:
Thanks I had indeed forgotten the '~' !

configs/pm2docs.json
robin-monnier: Yes I had

PR #453, opened 2018-06-12
Add specific start URL

JoelMarcey:
# Pull request motivation(s)

> There is no redirection from `https://docusaurus.io/docs/<lang>/` to a document page like `https://docusaurus.io/docs/<lang>/installation`. Could it be possible to have such redirection straight away?

Temporarily stop the redirection problem.

ref https://github.com/facebook/Docusaurus/pull/744

No redirection

Some redirection. This is a temp fix until we can come up with something better

No

No

### DIFF
--- a/configs/docusaurus.json
+++ b/configs/docusaurus.json
@@ -2,7 +2,7 @@
   "index_name": "docusaurus",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


JoelMarcey:
cc @endiliey

s-pace:
I need to edit it a bit but got what you want to do

PR #460, opened 2018-06-19
Revert "Update Nebular docs config as per new release"

s-pace:
Reverts algolia/docsearch-configs#459 

@nnixaa There is some issue with your URLs. For example, when you browse it, you are redirecting to https://akveo.github.io/nebular//(docs/guides/install-based-on-starter-kit)

Could you please fix it? 

Cheers

### DIFF
--- a/configs/nebular.json
+++ b/configs/nebular.json
@@ -1,11 +1,11 @@
 {
   "index_name": "nebular",
   "start_urls": [
-    "https://akveo.github.io/nebular/docs/"
+    "https://akveo.github.io/nebular/#/docs/"
   ],
   "stop_urls": [
     "/themes/",
-    "https://akveo.github.io/nebular/docs/.*#.*"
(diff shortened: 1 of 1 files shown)


nnixaa:
Hey @s-pace, could you let me know a browser where you are able to reproduce such behavior? And your steps would be a good help too, as I can't see it just simply browsing in chrome.

s-pace:
Actually it is not. It has appeared when I start from the landing page.

In order to render the client side javascript, we use a selenium emulator with firefox v45.0.2 embed.

We have the following error:

image
[from this page](https://akveo.github.io/nebular/docs/guides/backend-integration)


Could you try to fix it or to avoid client side rendering?

Cheers @nnixaa

nnixaa:
Hey @s-pace, thanks for the details, could you try now? Should be fixed.

s-pace:
Done, live @nnixaa

PR #461, opened 2018-06-19
changed start_urls

leifbro:
https://mlflow.org/docs/latest/ isn't reachable the way the site is currently built. Hoping to start search at https://mlflow.org/docs/latest/index.html

# Pull request motivation(s)


s-pace:
We need to have both in order to enlarge the scope of the search

PR #467, opened 2018-06-23
update jest config

endiliey:
# Pull request motivation(s)

https://github.com/facebook/jest/issues/6531

Trying to use language & version facetfilter for https://facebook.github.io/jest/

https://facebook.github.io/jest/ have a docsearch metadata from Docusaurus.

Example:
```
<meta name="docsearch:version" content="22.3">
```

### DIFF
--- a/configs/jest.json
+++ b/configs/jest.json
@@ -2,21 +2,58 @@
   "index_name": "jest",
   "start_urls": [
     {
-      "url": "https://facebook.github.io/jest/docs/(?P<lang>.*?)/getting-started",
+      "url": "https://facebook.github.io/jest/docs/(?P<lang>.*?)/",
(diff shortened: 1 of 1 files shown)


endiliey:
Hi @s-pace,

Jest used Docusaurus as a framework, but there is a bug on how we translate the facetfilters on the currently used Docusaurus version. (https://github.com/facebook/Docusaurus/pull/797)

When jest is currently on 'en' and version '23.0'. It will send
```
facetFilters: [ "language:en", "version:current" ]
```
instead of

```
facetFilters: [ "language:en", "version:23.0" ]
```

Is it possible to add alias on `jest.json` such that if it use facetfilter `version:current` it will be translated to `version:23.0` ?

PR #472, opened 2018-06-27
Update selectors for StackOverflow.design after site update

hellohynes:
We recently updated the class names on StackOverflow.design and the Docsearch functionality isn't displaying the correct results anymore. Here's an example of what we're seeing now:

image

This PR attempts to update `stackoverflow_design.json` so it's grabbing the correct selectors.

I have not tested this, so I'm not 100% sure how to test this locally.

### DIFF
--- a/configs/stackoverflow_design.json
+++ b/configs/stackoverflow_design.json
@@ -6,20 +6,16 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
-      "selector": ".-item a.is-active",
-      "global": true,
+      "selector": ".stacks-h1",
       "default_value": "Documentation"
     },
     "lvl1": {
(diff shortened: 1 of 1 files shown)


configs/stackoverflow_design.json
```diff
@@ -6,24 +6,20 @@
...
+      "selector": ".stacks-h2",
+      "default_value": "Section"
     },
-    "lvl2": "main h2",
-    "lvl3": "main h3",
-    "lvl4": "main h4",
-    "text": "main p, main li"
+    "lvl2": ".stacks-h3",
+    "lvl3": ".stacks-h4",
+    "text": ".stacks-copy, .stacks--list li"
```
aaronshekey: Should we include our tables as well?

hellohynes: Oh good point

hellohynes:
Thanks for merging this in @s-pace!

PR #477, opened 2018-06-28
Update stackpath.json

masojus:
A few changes to better test Algolia search on our dummy site:

- Add /blog/ which for now is actually sort of a changelog. If later we don't actually use this feature of Docusaurus, or it ends up not being docs-related, we can rethink this.
- Add /api/ which is a ReDoc page.
- `js_render` and `js_wait` since the /api/ page is currently rendered client-side by ReDoc.
- `js_wait` is set to 5 seconds just to see how well that page gets indexed, but this can be lowered for release.
- Replace the dummy domain name with an IP address for now. When the site goes public, we'll switch to the real domain, of course.
- I'm not sure why ".../docs/en/doc1" and ".../docs/" were in the original list...if it has something to do with the 301 redirect and the port? Or if it's because our sitemap was using the wrong domain (patching that for the demo site).

# Pull request motivation(s)

### DIFF
--- a/configs/stackpath.json
+++ b/configs/stackpath.json
@@ -1,13 +1,33 @@
 {
   "index_name": "stackpath",
   "start_urls": [
-    "http://www.taniayjustin.com:3000/docs/",
-    "http://www.taniayjustin.com:3000/docs/en/doc1",
-    "http://www.taniayjustin.com/docs/"
+    {
+      "url": "http://144.217.92.155:3000/docs/(?P<lang>.*?)/",
+      "variables": {
+        "lang": [
+          "en"
+        ]
+      },
+      "tags": [
+        "docs"
+      ]
+    },
+    {
+      "url": "http://144.217.92.155:3000/api/",
+      "tags": [
+        "api"
+      ]
+    },
+    {
+      "url": "http://144.217.92.155:3000/blog/",
+      "tags": [
+        "blog"
+      ]
+    }
   ],
   "stop_urls": [],
   "sitemap_urls": [
-    "http://www.taniayjustin.com:3000/sitemap.xml"
+    "http://144.217.92.155:3000/sitemap.xml"
   ],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


s-pace:
Hi @masojus ,

Thank you for the details. I will 

> Add /blog/ which for now is actually sort of a changelog. If later we don't actually use this feature of Docusaurus, or it ends up not being docs-related, we can rethink this.

👍 we would contact you if the content is not technical documentation/

> Add /api/ which is a ReDoc page.
> js_render and js_wait since the /api/ page is currently rendered client-side by ReDoc.
> js_wait is set to 5 seconds just to see how well that page gets indexed, but this can be lowered for release.
> Replace the dummy domain name with an IP address for now. When the site goes public, we'll switch to the real domain, of course.

Let's try it to see how it goes.

> I'm not sure why ".../docs/en/doc1" and ".../docs/" were in the original list...if it has something to do with the 301 redirect and the port? Or if it's because our sitemap was using the wrong domain (patching that for the demo site).

Might be because of pages that were not avaialble indeed.

s-pace:
> http://144.217.92.155:3000/docs/en/
Is not reachable (404)

It looks like the sitemap is not properly crawled this apply to your sitemap too. Are you up to date with the latest version of Docusaurus? 

We will need to dig into it instead.

masojus:
Hi, @s-pace  I copied what the Docusaurus docs themselves do at https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus.json in their first URL, which I also thought was odd since it's not available, (https://docusaurus.io/docs/en/). 

Your own documentation on `start_urls` seems a little confusing since it says "If you don’t use a sitemap, you must define at least one reachable URL (HTTP 20x)." but doesn't say anything about what it means to have or not have unreachable URLs. Why does Docusaurus have an unreachable URL regular expression in their config, then? Are they just wrong too and entirely relying on their sitemap, maybe?

> It looks like the sitemap is not properly crawled this apply to your sitemap too. Are you up to date with the latest version of Docusaurus?

I'm not sure what you mean, but in case there's a bug with Docusaurus and how they generate the sitemap, I did update that server to Docusaurus 1.3.0 (we were already using that version elsewhere). I do notice a difference in the sitemap, so maybe that makes a difference?

PR #480, opened 2018-06-29
Update fusionjs.json config

nadiia:
# Pull request motivation(s)
Currently only "Getting Started" section of the doc website is indexed and shows up in the search results and it leaves out the API documentation completely. This fix is to make sure the whole content of the fusionjs.com is indexed and searchable.

Indexes only "Getting Started" section.

Index all content in fusionjs.com

### DIFF
--- a/configs/fusionjs.json
+++ b/configs/fusionjs.json
@@ -1,8 +1,9 @@
 {
   "index_name": "fusionjs",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


s-pace:
Thank you for sharing what you expect @nadiia. I have update your configuration. Let us know if these changes suit you well. It is already live

PR #489, opened 2018-07-09
Fix Moleculer URL

icebob:
# Pull request motivation(s)
The search is not working on my [Moleculer site](https://moleculer.services/docs/0.13/). 
I tried to update the URL and hope it will fix it.

Not working

Working

### DIFF
--- a/configs/moleculer.json
+++ b/configs/moleculer.json
@@ -2,17 +2,17 @@
   "index_name": "moleculer",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @icebob 

Thanks for the update. It should be fixe.

Please note that the loading of button.js from buttons.github.com is really long and might be an issue at some point:

image

icebob:
Thanks, now it's working again. 
I will fix the button.js

PR #491, opened 2018-07-10
[fusionjs.com] Add code blocks to text content

KevinGrandon:
As our content is technical in nature, we would like to add code blocks to indexed content so folks can surface code examples in search.

# Pull request motivation(s)
Lots of missing content from our current search implementation due to ignoring code blocks.

Code blocks are not indexed.

Code blocks are indexed and code is displayed in search.

### Open questions
I'm not sure how code will be formatted in the search results, but ideally linebreaks would be preserved.

### DIFF
--- a/configs/fusionjs.json
+++ b/configs/fusionjs.json
@@ -13,7 +13,7 @@
     "lvl2": ".docSearch-content h3",
     "lvl3": ".docSearch-content h4",
     "lvl4": ".docSearch-content h5",
-    "text": ".docSearch-content p, .docSearch-content li"
+    "text": ".docSearch-content p, .docSearch-content li, .docSearch-content code"
   },
   "conversation_id": [
     "585325680"


s-pace:
👋 @KevinGrandon 

Thanks for the PR.

We do not recommend to index code since it will introduce a lot of noise.

Could you add a specific class to the inner payload content you want us to extract instead?

KevinGrandon:
Thanks for the feedback. Basically since our docs are quite technical, and there's a lot of cross-cutting plugins, we want people to see references to places where plugins are used.

Also seemed to be a fairly common use-case after looking at search results: https://github.com/algolia/docsearch-configs/search?q=code&unscoped_q=code

> Could you add a specific class to the inner payload content you want us to extract instead?

I think we would just end up adding the class to every code block around the site.

Right now it feels like results and information is missing, so I would prefer to have a bunch of noise if it means showing all of the results.

s-pace:
> Also seemed to be a fairly common use-case after looking at search results: https://github.com/algolia/docsearch-configs/search?q=code&unscoped_q=code

It is a common use-case and extracting the right payload of your block is really important.

>  I think we would just end up adding the class to every code block around the site.
> Right now it feels like results and information is missing, so I would prefer to have a bunch of noise if it means showing all of the results.

Let's give it a try this way. You now that if you want to enhance the relevancy, you will need to spot the most meaningful parts from your block.

PR #494, opened 2018-07-13
replace url

zzzzzb:
Replace invalid url with new url.

### DIFF
--- a/configs/san.json
+++ b/configs/san.json
@@ -2,7 +2,7 @@
   "index_name": "san",
   "start_urls": [
     "https://baidu.github.io/san/tutorial/start/",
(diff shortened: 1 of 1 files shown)


pixelastic:
Thanks for the PR! We just deployed the new config, you'll see the new results on the next crawl (in max 24h).

zzzzzb:
hi，We only want to crawl the next layer of this url:https://baidu.github.io/san/practice/

For example, I just want to crawl this url:https://baidu.github.io/san/practice/how-to-show-or-hide-an-element

But don't want to crawl this url: https://baidu.github.io/san/practice/how-to-show-or-hide-an-element/#1-用-class-控制元素的显示与隐藏

thanks！

PR #495, opened 2018-07-16
Update starnow.json

wub:
We're changing the structure of our docs. We'll re-add the searchbox once the new settings have been deployed and scraped.

Note - the site does not currently use these selectors, so our index will be empty until our next deployment.

Cheers!

### DIFF
--- a/configs/starnow.json
+++ b/configs/starnow.json
@@ -7,16 +7,11 @@
     ".*/$"
   ],
   "selectors": {
-    "lvl0": ".mt3 h1",
-    "lvl1": ".mt3 h2",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @wub 

Please ping us via this PR or the email thread as soon as your new layout is deployed. We will update your configuration accordingly (`nb_hits`).

Your DocSearch index will be automatically update within 24h though.

Cheers

wub:
@s-pace The change has just been deployed, thanks!

s-pace:
@wub We required to update selectors since it needed to target textual content.
Changes are live.

PR #498, opened 2018-07-18
Update README for sitemap_alternate_links

endiliey:
## Motivation

1. Fix typo on jest ? I think the start_url should be `https` because `http://jestjs.io` won't match `https://jestjs.io` 
1. Update docsearch config README on sitemap_alternate_links


s-pace:
👋 
Thanks for this PR. We are scheme agnostic (http and https work both) but definitely better to update https://github.com/algolia/docsearch-configs/commit/b693f084bc7ad76b08aed998465763ceec044f43

We want people to use the dedicated documentation since it is more detailed. https://github.com/algolia/docsearch-configs/commit/b6f1c7e93d0b166e384f42d7e1d50760b81af925

PR #500, opened 2018-07-19
Changing of the url

Vladislav-Lyuminarskiy:
The site moved to another url.

I also want to use Algolia search in another site on the same domain:
https://vladislav-lyuminarskiy-websites.github.io/Java-course-website/

Is it possible to create two independent indexes in this situation?

### DIFF
--- a/configs/c_vladislav.json
+++ b/configs/c_vladislav.json
@@ -1,7 +1,7 @@
 {
   "index_name": "c_vladislav",
   "start_urls": [
-    "https://vladislav-lyuminarskiy.github.io/C-course-website"
+    "https://vladislav-lyuminarskiy-websites.github.io/C-course-website/"
   ],
   "stop_urls": [],
   "selectors": {


s-pace:
hi @Vladislav-Lyuminarskiy 

Thank you for the PR. I would recommend you to use `tags` instead since pages are from the same domain and have the same layout:

Done thanks to this https://github.com/algolia/docsearch-configs/commit/cb1c0e28a55ae049ee538f52bb0c720813d07dbd

The update needed from your snippet will be done thanks to `algoliaOptions`:
```html
  apiKey: '892ad28dc056e1eb225c126678ef1c09',
  indexName: 'c_vladislav',
  algoliaOptions: { 'facetFilters': ["tags:$TAGS"] },
});
</script>
```


- Replace $TAGS with the tags you want to search on.
   The list of possible tags is hardcoded in the config.
   So as of today you have: c, java

  For example if you want to refine the search to the tags "c" just specify: 
```js
'facetFilters': ["tags:c"]
```

Vladislav-Lyuminarskiy:
Thanks a lot!

PR #502, opened 2018-07-24
Updated Sitecore JSS domain name to final one

kamsar:
# Pull request motivation(s)

Went live with our new docs, changed domain name to the final one

### DIFF
--- a/configs/jss.json
+++ b/configs/jss.json
@@ -1,7 +1,7 @@
 {
   "index_name": "jss",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


kamsar:
Thanks @s-pace! As far as I can tell we are using the latest version of DocSearch (at least the JS I'm referencing matches the URLs above). We're using it with React so the `docsearch()` call is buried in a React component.

s-pace:
You are warmly welcome @kamsar 

I meant the CSS library :)

PR #519, opened 2018-08-03
Force sitemap url crawling

solugebefola:
# Pull request motivation(s)

### DIFF
--- a/configs/timescale.json
+++ b/configs/timescale.json
@@ -3,6 +3,7 @@
   "sitemap_urls": [
     "http://docs-exp.timescale.com/docs-exp-sitemap.xml" 
   ],
+  "force_sitemap_urls_crawling": true,
   "start_urls": [
     {
       "url": "http://docs-exp.timescale.com/(?P<version>.*?)",
@@ -30,6 +31,7 @@
     "text": ".main-content__text p, .main-content__text li, .main-content__text pre code"
   },
   "js_render": true,
+  "js_wait": 2,
   "conversation_id": [
     "628616287"
   ],


s-pace:
👋 @solugebefola 

Please use the issue's template correctly, we would not be able to clearly help you otherwise.

Cheers

solugebefola:
### Pull request motivation(s)
#### What is the current behaviour?
The search functionality does not catch many of the pages on the site.  Example: searching for "indexing" does not return the page http://docs-exp.timescale.com/v0.10/using-timescaledb/schema-management#indexing.  This seems to be a problem with many of the pages on the site. 

I suspect that this is due to the crawler not rendering pages properly (They are built with react, and the way they are built may be incompatible with selenium?), but honestly, I don't have any evidence that suggests a cause.  I thought that forcing a crawl using the sitemap might work.

NB: Do you want to request a feature or report a bug?
NB2: Any other feedback / questions ?

solugebefola:
Here is a screenshot example. 
screen shot 2018-08-06 at 2 05 31 pm
You can see that I am on a page with "Indexing" in the title, yet a search for "indexing" does not show that page in the results.

s-pace:
Thank your for the details

`force_sitemap_urls_crawling` is here to change the way we handle parsed links from the sitemap. Not helpful here.

I have noticed and send you an email regarding an issue with this page http://docs-exp.timescale.com/v0.5/main that freeze the crawl. Since this webpage is not useful for us (no documentation in it), could you provide us a sitemap without this page? We will then avoid to crawl this exacte page.

Cheers

PR #526, opened 2018-08-08
Changed Thumbprint URL from `ui` to `components`

danoc:
# Pull request motivation(s)
We recently changed our URL structure so that `/ui` is now `/components`.

Results within `/components` are not working.

Results should appear.


danoc:
Closing in favor of a different PR that changes a few things.

PR #527, opened 2018-08-09
Update thumbprint.json

tomgenoni:
# Pull request motivation(s)

We're changing domain names and adding additional doc sets.

The previous config does not contain these `start_urls`

### DIFF
--- a/configs/thumbprint.json
+++ b/configs/thumbprint.json
@@ -1,10 +1,12 @@
 {
   "index_name": "thumbprint",
   "start_urls": [
-    "https://thumbprint.thumbtack.com/guide/",
-    "https://thumbprint.thumbtack.com/tokens/",
-    "https://thumbprint.thumbtack.com/ui/",
-    "https://thumbprint.thumbtack.com/icons/"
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @tomgenoni?

I have update your PR in order to extract the right content from your page. Please note that:

- `https://api.thumbprint.design/atomic/` and `https://thumbprint.design/notes/` URLs are not available so far
- We [de not recommend to extract big code block](https://community.algolia.com/docsearch/documentation/docsearch-FAQ/customize-configuration-file/). The ones you have set were matching some tables several times since matching elements were nested into other matching ones (e.g. for this [page](https://api.thumbprint.design/tokens/#avatar))

I have set some new selectors that sound more relevant but please try it on your own first to check that you are extracting atomic element. You may need to define some specific classes like `.docSearch-content` in order to target the right elements.

tomgenoni:
Thank you! We will have the missing URLs up shortly but wanted to the index to be ready. In the future I'll mention if any URLs aren't yet active.

PR #528, opened 2018-08-09
Update collab.json for TeamForge 18.2

mahendranjayavel:
Please review the update to collab.json for TeamForge 18.2.

### DIFF
--- a/configs/collab.json
+++ b/configs/collab.json
@@ -20,6 +20,12 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @mahendranjayavel 

Changes are live and the new `facetFilter` teamforge182 is available.

Please note that we do not crawl and scrap the full java documentation since it is not matching the set of selectors you have defined. We do not recommend yo use DocSearch on it so far anyway.

Cheers

PR #536, opened 2018-08-16
Update thumbprint.json

tomgenoni:
# Pull request motivation(s)

Point search to new URL.

Pointing to old URL.

Index all pages except for those in `stop_urls`

### DIFF
--- a/configs/thumbprint.json
+++ b/configs/thumbprint.json
@@ -1,15 +1,12 @@
 {
   "index_name": "thumbprint",
   "start_urls": [
-    "https://guide.thumbprint.design/",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @tomgenoni 

I have fixed this PR:
- removing/adapting the `stop_urls` and `scrap_start_urls` in order to only scrap of the main pages. The crawl was aborted otherwise since one `stop_urls` was equal to the `start_url`
- removing superfluous `default` selectors_key
- Adding a new level based on the sidebar
- Deploying this according to what you have described from the motivation

Let us know id this suits you well.

PR #549, opened 2018-09-02
Update liveperson.json

EdenKupe:
# Pull request motivation(s)

New CSS structure in place, updating to fit.

SidebarTitle was used, is now gone so moved xpath to select breadcrumbs like lvl0 and lvl1. Did the same for lvl3 to make sure it is consistent, also goes to breadcrumbs now.

CSS class for content also changed so I updated to reflect.

Crawler should use the new selectors to populate the levels.

### DIFF
--- a/configs/liveperson.json
+++ b/configs/liveperson.json
@@ -18,11 +18,17 @@
       "type": "xpath",
       "selector": "//nav/a[@class='breadcrumb-item'][2]"
     },
-    "lvl2": ".sidebarTitle",
-    "lvl3": ".inside-page-content h2",
-    "lvl4": ".inside-page-content h3",
-    "lvl5": ".inside-page-content h4",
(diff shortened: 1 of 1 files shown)


EdenKupe:
Hey @s-pace ! Thanks for merging :) I'm still not seeing the changes; has the crawler run yet with the new specifications? Anything further I need to do to make sure the UI reflects the new entries?

Thanks!

EdenKupe:
Got it! Will ping you once the new site goes live! Appreciate it :)

EdenKupe:
Hey @s-pace ! I'm unable to tag the team (because it's private, I think) but in any case, the site is live! If you could trigger the crawler that would be great :)

s-pace:
Right, you were our beta tester for this handle. I was not totally such either it was only internal within the team.
Crawling is currently on going.

PR #550, opened 2018-09-03
katex: Add td to text selectors

ylemkimon:
# Pull request motivation(s)

KaTeX has lists of supported functions in https://katex.org/docs/supported.html and https://katex.org/docs/support_table.html using `table`s. It'd be nice to have them searchable.

@s-pace What do you think is the best course of action here? Thank you!

Supported functions in `<td>` are not searchable.

Supported functions in `<td>` are searchable.

### DIFF
--- a/configs/katex.json
+++ b/configs/katex.json
@@ -20,7 +20,7 @@
     "lvl2": ".post h2",
     "lvl3": ".post h3",
     "lvl4": ".post h4",
-    "text": ".post article p, .post article li"
+    "text": ".post article p, .post article li, .post article td"
   },
   "selectors_exclude": [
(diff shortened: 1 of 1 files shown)


s-pace:
Hi @ylemkimon 

Thank you for the details.

Adding this new selector might introduces a lot of noise since it is made with a lot of specific character. These characters are pre processed by our engining in order to split string in word. You may need to add the most important character from [separatorsToIndex](https://www.algolia.com/doc/api-reference/api-parameters/separatorsToIndex/) exposed under `custom_settings` in your DocSearch configuration.

Let's give it a try and you will update it based on how the relevancy suits you

PR #551, opened 2018-09-03
Update parseplatform.json

flovilmart:
Adds tags to start URL's (Taken from the legacy parse.json index)

# Pull request motivation(s)

We have a page for each section, as such we need to filter the search results.


flovilmart:
thx!

PR #552, opened 2018-09-04
katex.json: exclude KaTeX rendering

ylemkimon:
# Pull request motivation(s)

It seems KaTeX rendering, MathML, and annotations are scraped and included in the search result. This excludes KaTeX rendered nodes.

KaTeX KaTeX rendering, MathML, and annotations are included in the search result.

KaTeX KaTeX rendering, MathML, and annotations should not be included in the search result.


ylemkimon:
Thank you!

PR #559, opened 2018-09-07
Update Loadster paths and selectors.

azhawkes:
We've added some new documentation sections (/faqs and /articles) to the Loadster site that we'd like to make searchable. Also, we'll be removing the subheaders from the side menu so I'm changing lvl0 to be the top-level tab text ("Manual", "FAQs", "Articles") instead.

### DIFF
--- a/configs/loadsterperformance.json
+++ b/configs/loadsterperformance.json
@@ -1,19 +1,16 @@
 {
   "index_name": "loadsterperformance",
   "start_urls": [
-    "https://loadster.app/documentation/"
+    "https://loadster.app/documentation/",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @azhawkes 

Thanks for the PR

Regarding the updates of `start_urls, please note that we do host DocSearch for documentation content. You will need to run it on your own with you own algolia account otherwise. Since the pages are still educational one we will broaden the scope of the scrap.

s-pace:
I have made some changes in order to leverage the different `selectors_key`

azhawkes:
Thanks @s-pace really appreciate it!

PR #572, opened 2018-09-23
add new start_urls entry

kenfai:
Adding new start_urls entry.
Thanks!


s-pace:
👋 @kenfai 

In order to enable the needed javascript rendering, we are currently using a selenium emulator with a firefox v45.0.2 driver (we are working on improving it)

It looks like your website is not rendering properly on this browser. Could you have a look please?

image

The current crawl is failing. Your index will not be updated.

PR #574, opened 2018-09-24
Update jovo.json

jankoenig:
We moved the technical tutorials to a different subfolder (`/tutorials`) and would love to have them indexed as well. Is that possible?

Thanks!
Jan

### DIFF
--- a/configs/jovo.json
+++ b/configs/jovo.json
@@ -9,6 +9,13 @@
       ]
     },
     {
+      "url": "https://www.jovo.tech/tutorials/",
+      "page_rank": 1,
+      "tags": [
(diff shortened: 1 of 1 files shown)


s-pace:
Since the content is still helpful documentation, we will merge it

s-pace:
👋 @jankoenig 

The crawler is not behaving as you might be expected . You needed to use the [`selectors_key`](https://community.algolia.com/docsearch/config-file.html#selectors_key%2C-tailor-your-selectors) attribute.

Should be fixed thanks to the commit above

jankoenig:
Ah, got it. Thank you!

PR #575, opened 2018-09-25
Update Scalafix configuration for new website.

olafurpg:
Updates scalafix.json to match [scalameta.json](https://github.com/algolia/docsearch-configs/blob/master/configs/scalameta.json). The two
websites use the same documentation tool https://docusaurus.io/

- This one works great: https://scalameta.org/
- This one has outdated config: https://scalacenter.github.io/scalafix/

# Pull request motivation(s)

I updated the Scalafix site recently to use Docusaurus (https://docusaurus.io/) instead of sbt-microsite (https://47deg.github.io/sbt-microsites/)

Searching "Cookbook" yields no results.

I expected "cookbook" to yield a result linking for example here https://scalacenter.github.io/scalafix/docs/developers/semantic-tree.html#cookbook 

Algolia DocSearch is great, we are using it in a few Scala libraries and our users love it. Thank you so much!

### DIFF
--- a/configs/scalafix.json
+++ b/configs/scalafix.json
@@ -1,29 +1,35 @@
 {
   "index_name": "scalafix",
   "start_urls": [
-    "https://scalacenter.github.io/scalafix/docs/",
-    "https://scalacenter.github.io/scalafix/docs/users/installation"
+    "https://scalacenter.github.io/scalafix/",
   ],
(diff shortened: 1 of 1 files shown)


s-pace:
Thank you for the detailed info, 💯

PR #578, opened 2018-09-26
Added "only_content_level": true for instantsearch

bildungsroman:
In an attempt to solve my issue #457


s-pace:
👋 @bildungsroman  I think you meant algolia/docsearch#457

I will merge and deploy it in order to avoid having record without `content` attribute level set.

bildungsroman:
Thank you!

PR #584, opened 2018-10-02
Updated moleskinestudio.com for new site

alexlouden:
Hi, we've rolled our support site into our homepage, so I've updated the scraper config to suit [moleskinestudio.com](https://moleskinestudio.com) instead. I've run it locally against a trial Agolia instance to test with - seems to work well! 😃 

Thanks so much for offering this!

### DIFF
--- a/configs/moleskinestudio.json
+++ b/configs/moleskinestudio.json
@@ -15,8 +15,14 @@
       "global": true,
       "default_value": "Documentation"
     },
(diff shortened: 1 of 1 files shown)


s-pace:
🙏 if you run it loacally 💯 
Deployed & live ✅

alexlouden:
Thank you! 🙏

PR #586, opened 2018-10-05
blocking a specific URL from showing up in SendGrid docssearch

ksigler7:
# Pull request motivation(s)
We are releasing functionality in a beta and right now, we don't want this page to show up in any search results.

Currently the page doesn't exist. We are deploying it today.

We're expecting this page not to show up in search results when we deploy it today

### DIFF
--- a/configs/sendgrid_hc.json
+++ b/configs/sendgrid_hc.json
@@ -9,7 +9,7 @@
     },
     "https://sendgrid.com/docs/"
   ],
-  "stop_urls": [],
+  "stop_urls": ["https://sendgrid.com/docs/ui/sending-email/getting-started-with-automation/"],
   "selectors": {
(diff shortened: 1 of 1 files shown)


ksigler7:
thanks @s-pace!!

s-pace:
👋 @ksigler7 

It looks like that some custom code are currently failing. It is [broking the search](https://sendgrid.com/docs/ui/sending-email/getting-started-with-automation/). I can confirm you that your DocSearch index is working fine with the built-in UI

image

ksigler7:
@s-pace - we are trying to block this page from showing up in the search results. is there another way we can do this with docsearch?

ksigler7:
@s-pace - this is a great catch! we are working on a patch now from our side

PR #587, opened 2018-10-05
Add laravel-blade-x

freekmurze:
Hi, we added a new documentation site that is ready to be indexed.

https://docs.spatie.be/laravel-blade-x/v2/introduction


freekmurze:
Thank you!

PR #590, opened 2018-10-09
Added 2.3 version for ezplatform

damianz5:
Added new 2.3 version for doc.ezplatform.com

ping @s-pace


s-pace:
👋 @damianz5,

Changes are live. I have updated the lvl0 selector since it was not matching anything.

PR #592, opened 2018-10-11
Adding test Redis config

pixelastic:
This is the config I used on my fork of the Redis website. I'm submitting it as a PR to get @s-pace expert eyes on it, and if it looks ok we'll merge itt (but first we'll need to put the real url)

### DIFF
--- a/configs/redis.json
+++ b/configs/redis.json
@@ -11,7 +11,8 @@
     },
     {
       "url": "https://docsearch-redis.netlify.com/topics",
-      "selectors_key": "topics"
+      "selectors_key": "topics",
+      "page_rank": 1
     }
   ],
   "stop_urls": [],
@@ -31,12 +32,15 @@
         "global": true,
         "default_value": "Commands"
       },
-      "lvl1": ".text h1 .name",
-      "lvl2": ".article-main h2",
-      "lvl3": ".article-main h3",
-      "lvl4": ".article-main h4",
-      "lvl5": ".article-main h5",
-      "text": ".article-main p, .article-main li, .arg"
+      "lvl1": {
+        "selector": ".article-aside li strong",
+        "global": true
+      },
+      "lvl2": ".text h1.command",
+      "lvl3": ".article-main h2",
+      "lvl4": ".article-main h3",
+      "lvl5": ".article-main h4",
+      "text": ".article-main p, .article-main li"
     },
     "topics": {
       "lvl0": {
@@ -53,7 +57,8 @@
     }
   },
   "selectors_exclude": [
-    "#topic > p:first-child"
+    "#topic > p:first-child",
+    ".site-content .metadata"
   ],
   "nb_hits": 7883
 }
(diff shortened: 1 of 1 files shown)


pixelastic:
image

TODO:
- [ ] Have commands displayed before articles
- [ ] Have command overview displayed (first §) instead of 3 x the command name

pixelastic:
I updated the config, that's better, I now have commands and topics grouped together. But there's still something I don't manage to do:
image

Instead of:
image

How can I have?
![image](https://user-images.githubusercontent.com/283419/46818310-35aff780-cd81-11e8-923b-1db4450c2945.pn

Considering that my selectors are matching this:

```json
{
  "lvl0": "Commands",
  "lvl1": "XTRIM",
  "lvl2": "XTRIM key MAXLEN [~] count",
  "lvl3": "Return value"
}```

**Comment by s-pace** [comment]
You would need to set `min_indexed_level=2` in order to get what you expect

**Comment by s-pace** [comment]
To promote a set of URLs, you can use [the DocSearch parameter `page_rank`](https://community.algolia.com/docsearch/config-file.html#using-page-rank).

**Comment by s-pace** [comment]
I would avoid having empty selectors, it can always match something in the future.

File: configs/redis.json [hunk 1]
```diff
@@ -0,0 +1,64 @@
...
+    {
+      "url": "https://docsearch-redis.netlify.com/topics",
+      "selectors_key": "topics",
+      "page_rank": 1
+    }
+  ],
+  "stop_urls": [],
+  "strip_chars": "*",
+  "selectors": {
+    "default": {
```
pixelastic: There is no real `default` page, everything is either a command or a topic, so I added the two custom selector entries and set the default to a set of empty selectors.

I tried removing the `default` key but the crawling failed.

What do you suggest putting instead of empty selectors? something like `#lkdjqkksqdkjhqsjkdkjsq`?

s-pace: I would recommend to set selectors in order to have something that will evolve properly.

pixelastic: I don't see how I could do that. There is no "default" page, they are either topics or commands, and each require their own specific selectors.

What selector would you suggest I add?

s-pace: Not yet but it can happrn later

s-pace: Using the same selector

pixelastic: > Using the same selector

I don't get that. The same selector as what? Do you mean copy-pasting the same selectors as the `topics` or `commands` into the `default`?

s-pace: Yes

pixelastic: But which one then? `topics` or `commands`? And I don't see why it's better than empty selectors. We have no idea what the new pages (that don't exist yet) will look like, so no idea if those selectors will match anything

s-pace: The more generic one

pixelastic: That's the thing :) They're both specific, none of them is generic.

And I tried to remove the `default` key, but they DocSearch was giving me errors, so the key needs to be present. It's just that I don't know what to put in the selectors. I put `#will-never-match` (or something similar) to document that those selectors are garbage and not meant to be used. See it as a way to document the hack

s-pace: Fine with it. You can merge and deploy it when you need

PR #596, opened 2018-10-16
Phalcon config

pixelastic:
I'm having trouble selecting the correct lvl1. The element in the
sidebar is only bolded through `style` and there is no class attached
to it. Maybe we should ask the owner to add one?

Also I hard-coded the versions, maybe I should have used the JS
snippet to get the versions, but that seems brittle.

### DIFF
--- a/configs/phalcon.json
+++ b/configs/phalcon.json
@@ -16,7 +16,7 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
-      "selector": ".Sidebar li.__isActive span",
+      "selector": ".Sidebar li.__isActive > span",
       "global": true,
       "default_value": "Documentation"
     },
--- a/configs/phalcon.json
+++ b/configs/phalcon.json
@@ -1,17 +1,8 @@
 {
   "index_name": "phalcon",
   "start_urls": [
-    {
-      "url": "https://phalc.one/docs/(?P<version>.*?)",
-      "variables": {
-        "static_variable": [
-          "3.1",
-          "3.2",
-          "3.3",
-          "3.4"
-        ]
-      }
-    }
+    "https://phalc.one/docs/",
+    "https://phalc.one/docs/3.4/introduction"
   ],
   "stop_urls": [],
   "selectors": {
@@ -20,12 +11,21 @@
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": {
-      "selector": "???"
-    },
-    "lvl2": ".DocsContainer article h1",
-    "lvl3": ".DocsContainer article h2",
-    "lvl4": ".DocsContainer article h3",
+    "lvl1": ".DocsContainer article h1",
+    "lvl2": ".DocsContainer article h2",
+    "lvl3": ".DocsContainer article h3",
+    "lvl4": ".DocsContainer article h4",
+    "lvl5": ".DocsContainer article h5",
     "text": ".DocsContainer article p, .DocsContainer article li"
-  }
+  },
+  "selectors_exclude": [
+    ".article-menu"
+  ],
+  "js_render": true,
+  "custom_settings": {
+    "attributesForFaceting": [
+      "version"
+    ]
+  },
+  "nb_hits": 3198
 }
--- a/configs/phalcon.json
+++ b/configs/phalcon.json
@@ -1,0 +1,31 @@
+{
+  "index_name": "phalcon",
+  "start_urls": [
+    {
+      "url": "https://phalc.one/docs/(?P<version>.*?)",
+      "variables": {
+        "static_variable": [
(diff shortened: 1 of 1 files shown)


s-pace:
I would use `.Sidebar li.__isActive > span` instead in order to avoid future confusion.

Regarding the version, I would ask them to add a meta

pixelastic:
> I would use `.Sidebar li.__isActive > span` instead in order to avoid future confusion.

Good call. Done.

> Regarding the version, I would ask them to add a meta

Ok, ping me when added, I'll update the config.

configs/phalcon.json
```diff
@@ -1,17 +1,8 @@
...
-        "static_variable": [
-          "3.1",
-          "3.2",
-          "3.3",
-          "3.4"
-        ]
-      }
-    }
+    "https://phalc.one/docs/",
+    "https://phalc.one/docs/3.4/introduction"
```
pixelastic: So we don't index other versions? And as `/docs/` redirects to the most recent version, why don't we simply index this one?

configs/phalcon.json
```diff
@@ -20,12 +11,21 @@
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": {
-      "selector": "???"
-    },
-    "lvl2": ".DocsContainer article h1",
-    "lvl3": ".DocsContainer article h2",
-    "lvl4": ".DocsContainer article h3",
+    "lvl1": ".DocsContainer article h1",
```
pixelastic: This is not correctly grabbing the title. For example on https://phalc.one/docs/3.4/config, `lvl1` should be "Config" (as selected in the sidebar), not `Reading Configurations`.

configs/phalcon.json
```diff
@@ -20,12 +11,21 @@
...
+    "lvl5": ".DocsContainer article h5",
     "text": ".DocsContainer article p, .DocsContainer article li"
-  }
-}
+  },
+  "selectors_exclude": [
+    ".article-menu"
+  ],
+  "js_render": true,
+  "custom_settings": {
```
pixelastic: Where is the version defined?

configs/phalcon.json
s-pace: We do but we extract the version from the meta

configs/phalcon.json
s-pace: Yep I thought about that. We will have duplicate if we do that. Let's do that and we will iterate if needed.

configs/phalcon.json
s-pace: meta

s-pace:
@pixelastic Please do not forget to bootstrap from thread in order to have `conversation_id`

configs/phalcon.json
pixelastic: Oh, so they updated their website. Got it.

configs/phalcon.json
pixelastic: What kind of duplicate?

configs/phalcon.json
s-pace: In their title, they repeat the lvl 0

configs/phalcon.json
pixelastic: I understand we extract the version from the `docsearch:version` meta, but why do we need to set `"https://phalc.one/docs/3.4/introduction"` as part of the start urls?

`"https://phalc.one/docs/"` redirects to the latest version (today it's 3.4, but tomorrow it could be 3.5). I would think it's enough to put only `"https://phalc.one/docs/"` in the start urls. Why do we need to put the other one?

configs/phalcon.json
pixelastic: > In their title, they repeat the lvl 0

I don't understand what you mean by that? You mean the `<title>` tag?

configs/phalcon.json
s-pace: Duplicate from the website itself. Title remind the lvl0 in their text

configs/phalcon.json
s-pace: Thanks to their sitemap we do not need it anymore.
It was required because https://phalc.one/docs/ was not available

configs/phalcon.json
s-pace: The time I have cheked it was builded this way: lvl1= Getting started: Introduction (lvl0=Getting started)

configs/phalcon.json
pixelastic: Ah yes, I understand now. `./docs` is a 404, so we cannot crawl anything from there. But we need it as part of the `start_urls` because it's used as a reference to see if we should index a page or not. And we need to pass the `./3.4/introduction` to have an actual entry point to crawl.

Thanks!

PR #597, opened 2018-10-16
Adding Yliyun config

pixelastic:
This seems ok, the markup is pretty straigtforward but I still wanted
you review @s-pace :)

### DIFF
--- a/configs/yliyun.json
+++ b/configs/yliyun.json
@@ -1,0 +1,16 @@
+{
+  "index_name": "yliyun",
+  "start_urls": [
+    "http://167.ylicloud.com/docs/index.html",
+    "http://167.ylicloud.com/docs"
+  ],
+  "stop_urls": [],
+  "selectors": {
+    "lvl0": "#content h1",
+    "lvl1": "#content h2",
+    "lvl2": "#content h3",
+    "lvl3": "#content h4",
+    "lvl4": "#content h5",
+    "text": "#content p, #content li"
+  }
+}


s-pace:
On [index.html](http://167.ylicloud.com/docs/index.html): 
image

Let's wait for a production ready version

pixelastic:
Ah weird, it was working for me when I did the config.

configs/yliyun.json
```diff
@@ -0,0 +1,18 @@
...
+  "stop_urls": [],
+  "selectors": {
+    "lvl0": "#content h1",
+    "lvl1": "#content h2",
+    "lvl2": "#content h3",
+    "lvl3": "#content h4",
+    "lvl4": "#content h5",
+    "text": "#content p, #content li"
+  },
+  "user_agent": "Googlebot",
```
pixelastic: What is this for?

s-pace: Sometime body is empty with our regular user_agent

pixelastic: Is that happening on all configs or only this one? If it's happening on several configs it might be something we need to fix on a higher level in DocSearch and not in the config.

If the default Scrappy user-agent is rejected I would suggest we either pass "DocSearch" as a UA and ask them to authorize it (this will be more fair than masquerading as Googlebot).

s-pace: Just few ones so far. Good idea

pixelastic: I added an issue in https://github.com/algolia/docsearch-scraper/issues/404 to track this

PR #606, opened 2018-10-25
Update rafaelmaia.net/pavo config with pkgdown auto-generated config

Bisaloo:
# Pull request motivation(s)

This PR fine-tunes the configuration of docsearch for http://rafaelmaia.net/pavo by using the recommended config for `pkgdown` websites.

I also expect this change to fix the `HTTP 404` errors we get by following the steps below.

Some links in the search suggestion cause `HTTP 404` errors.

Steps to reproduce:

1. Visit http://rafaelmaia.net/pavo/index.html
1. Search `dichromat`
1. Click on `Arguments` in one of the suggested items
1. It leads to http://rafaelmaia.net/pavo/undefined?q=dichromat#arguments which does not exist

No `HTTP 404` errors.

I am not 100% sure the 404 errors come from this but this PR is still useful anyways because it makes sense for us to stick to the recommended `pkgdown` documentation.

This file was auto-generated using the latest `pkgdown` version (`1.1.0.9000`, still in development).

Please let me know if I need to change anything on this PR.

cc @rmaia

### DIFF
--- a/configs/rafaelmaia_pavo.json
+++ b/configs/rafaelmaia_pavo.json
@@ -1,35 +1,94 @@
 {
   "index_name": "rafaelmaia_pavo",
   "start_urls": [
-    "http://rafaelmaia.net/pavo/"
+    {
+      "url": "http://rafaelmaia.net/pavo/index.html",
+      "selectors_key": "homepage",
+      "tags": [
+        "homepage"
+      ]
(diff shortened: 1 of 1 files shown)


Bisaloo:
Yes, I can confirm everything works great now. Thank you!

PR #609, opened 2018-10-30
Add newest version of docs

solugebefola:
# Pull request motivation(s)
Update config with docs for the latest version of the database.


s-pace:
👋 @solugebefola 
It looks like the sitemap is not available:
image

PR #610, opened 2018-10-30
Zowe documentation has added versioning support

jackjia-ibm:
# Pull request motivation(s)

Our documentation has version information in the URL, we wish the search only shows result of current version. We currently only have one version "latest", but expect within this year, we may have a new legacy version of documentations along with "latest".

Currently will show results of all versions.

Show only current version of documentation.

I will update our search client to add algoliaOptions.facetFilters with version filter.

Thank you very much!

### DIFF
--- a/configs/zowe.json
+++ b/configs/zowe.json
@@ -1,7 +1,12 @@
 {
   "index_name": "zowe",
   "start_urls": [
-    "https://zowe.github.io/docs-site/"
+    {
+      "url": "https://zowe.github.io/docs-site/(?P<version>.*?)/",
+      "variables": {
+        "version": ["latest"]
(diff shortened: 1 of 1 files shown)


s-pace:
👋  @jackjia-ibm,

You can update the script as follow:

```html
  apiKey: '59ff39ed48d0820010c7e09fc4b677bf',
  indexName: 'zowe',
  algoliaOptions: { 'facetFilters': ["lang:$LANG", "version:$VERSION"] },
});
</script>
```
  For example if you want to refine the search to the lang "en-US" and the version "latest" just specify: 

```js
'facetFilters': ["lang:en-US", "version:latest"]
```

PR #614, opened 2018-11-01
Update djrad.json

davecoates:
For djrad config:
- Add new selectors_exlude
- Change selector for lvl1
- Change default_value for lvl0

# Pull request motivation(s)

Resolve issues crawling djrad documentation

Currently output from code examples are crawled resulting in following links that go to the same page.

Ignore the code examples


s-pace:
Let's give it a try. Usually, the pages is parsed before removing excluded elements to check for links to follow.

PR #616, opened 2018-11-02
URL updates

mbahar:
I have updated URLs for start_url and sitemap.

We have multiple document locations, will this work?
```
  "start_urls": [
    "https://aws-amplify.github.io/docs",
    "https://aws-amplify.github.io/amplify-js/"
```

# Pull request motivation(s)


s-pace:
Sure!

PR #618, opened 2018-11-05
Update activepivot.json

pbares:
# Pull request motivation(s)
The url of my website has changed so this file needed to be updated to reflect these changes. 

##### NB: Other question
May I know what the dynamic_variable is? Because the url refers to the '5.7-next' version (which is a minor version) but minor versions are released once in a while. Do I have to change the dynamic_variable url to "https://activeviam.com/activepivot/5.8-next/versions.html", "https://activeviam.com/activepivot/5.9-next/versions.html" each time a new minor version is shipped? That is to say doing a new PR.

### DIFF
--- a/configs/activepivot.json
+++ b/configs/activepivot.json
@@ -2,17 +2,17 @@
   "index_name": "activepivot",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


s-pace:
Everything is live and deployed

PR #630, opened 2018-11-13
Update axibase.json. New selector.

unrealwork:
* Added additional selector for settings search in `charts` doc


unrealwork:
@s-pace 
Hi, Sylvain! Thanks for the warning! Is the `table` selector` appropriate for indexing?

s-pace:
Hi @unrealwork,

`table` element are too big. Selectors must focus on textual element. 

Could you provide us an example of the element you would like to include in the search?

unrealwork:
@s-pace 
We have some settings and their descriptions stored in the table.
**Example**: [Time Chart](https://axibase.com/docs/charts/widgets/time-chart/#widget-settings)

configs/axibase.json
```diff
@@ -36,9 +36,9 @@
     "lvl1": ".content h1",
     "lvl2": ".content h2",
     "lvl3": ".content h3",
-    "lvl4": ".content h4",
+    "lvl4": ".content h4, .content table tr td:first-child",
     "lvl5": ".content h5",
-    "text": ".content p, .content li, .content pre",
+    "text": ".content p, .content li, .content pre, .content table, .content table td:nth-child(2)",
```
s-pace: Please add the `<p/>` since this selector will also index the noise introduces by **Example:....**

unrealwork: @s-pace 
It's hard to implement, our docs are generated from markdown source by `vuepress`. We understand and accept the noise. Probably, in the future, we will try to restructure our docs, but it requires too much time. But know it is ok, that our examples are indexed too. I hope that it doesn't increase index complexity too much.

s-pace: Makes sense. If you are aware and fine with it let's go :)

s-pace: Please note that pages with regular table such as https://axibase.com/docs/atsd/changelogs/2017_feb/ these pages introduce a lot of noise too

PR #635, opened 2018-11-13
Update bazel.json

laurentlb:
Index pre.rule-signature

The signature includes the fully qualified names (e.g. "ctx.actions" in
https://docs.bazel.build/versions/master/skylark/lib/ctx.html#actions),
which is what the users are going to search for.


s-pace:
👋 @laurentlb 

Thank you for the update. This PR has been deployed.

PR #637, opened 2018-11-14
Update supportpal.json

jshah4517:
# Pull request motivation(s)

Adding new versions of the SupportPal documentation.

The `nb_hits` value also probably needs to be updated, but I'm not sure what the new number is.

### DIFF
--- a/configs/supportpal.json
+++ b/configs/supportpal.json
@@ -6,6 +6,8 @@
       "variables": {
         "version": [
           "current",
+          "2.4.0",
+          "2.3.1",
           "2.3.0"
         ]
       }


s-pace:
👋 @jshah4517 

No problem, this value show the nb of record

>nb_hits automatically updated by DocSearch every time it runs your config. It is set to the number of records that were extracted and indexed. We check this key internally to keep track of any unintended spike or drop that could reveal a misconfiguration.

> You don't have to touch it, we're documenting it here in case you were wondering what this was about.

[REF](https://community.algolia.com/docsearch/config-file.html#nb_hits-special)

PR #639, opened 2018-11-15
Update warp10.json

Giwi:
# Pull request motivation(s)

### DIFF
--- a/configs/warp10.json
+++ b/configs/warp10.json
@@ -1,10 +1,10 @@
 {
   "index_name": "warp10",
   "start_urls": [
-    "http://www.warp10.io"
+    "https://www.warp10.io"
   ],
   "stop_urls": [
-    "/tags/"
+     "/$"
   ],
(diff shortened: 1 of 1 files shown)


configs/warp10.json
```diff
@@ -1,10 +1,10 @@
 {
   "index_name": "warp10",
   "start_urls": [
-    "http://www.warp10.io"
+    "https://www.warp10.io"
   ],
   "stop_urls": [
-    "/tags/"
+     "/$"
```
s-pace: Using this regex will avoid crawling `/content/` pages.

We will only crawl `/doc/`page. Is it fine for you?

s-pace: You can try it out straight away (Changes are live)

Giwi: Hi, thanks, I will remove any stop url.
My initial idea was to not crawl urls with a trailing /

PR #640, opened 2018-11-15
Update warp10.json

Giwi:
# Pull request motivation(s)

### DIFF
--- a/configs/warp10.json
+++ b/configs/warp10.json
@@ -3,10 +3,7 @@
   "start_urls": [
     "https://www.warp10.io"
   ],
-  "stop_urls": [
-    "/tags/",
-    "/$"
-  ],
+  "stop_urls": [],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


s-pace:
Changes are live from 84cde80. We still do not crawl `/tags/` pages as these pages redirect to `/doc/` pages

Giwi:
Hi, I do not want any stop url, why did you add /tags/ ?

Giwi:
Hi, we need /tags/* pages to be crawled to index all of our functions.
For example : https://www.warp10.io/doc/RTFM is not indexed without navigating through https://www.warp10.io/tags/misc

s-pace:
@Giwi Alright. So please transform the `h3` tags from your /tags/ pages to a `h1` since it introduces noise within your DocSearch record.

Example for https://www.warp10.io/tags/logic

Giwi:
Thanks, I will change that.

PR #641, opened 2018-11-15
Update warp10.json

Giwi:
# Pull request motivation(s)

We have a javascritpt generated menu

this menu is not crawled

the menu must be crawled :)

### DIFF
--- a/configs/warp10.json
+++ b/configs/warp10.json
@@ -20,6 +20,7 @@
   "user_agent": "AlgoliaBot",
   "js_render": true,
   "js_wait": 2,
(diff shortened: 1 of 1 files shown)


s-pace:
> otherwise you can update the config by removing this lvl0 and replace every lvl selectors by its deeper one from the current config. It will be a quick win and will not require any change for both sides.

Solution offered actually implement for you 🔝 https://github.com/algolia/docsearch-configs/commit/61cd84a8bac39c59f4742aa8d32af47ec45dffc8

s-pace:
Live and deployed

PR #642, opened 2018-11-15
[Update axibase.json]scan last column in tables as well

unrealwork:
### Summary
* Added indexing of the last column in `tr`

### DIFF
--- a/configs/axibase.json
+++ b/configs/axibase.json
@@ -38,7 +38,7 @@
     "lvl3": ".content h3",
     "lvl4": ".content h4, .content table tr td:first-child",
     "lvl5": ".content h5",
-    "text": ".content p, .content li, .content pre, .content table td:nth-child(2)",
+    "text": ".content p, .content li, .content pre, .content table td:nth-child(2), .content table td:last-child",
     "lang": {
       "selector": "/html/@lang",
       "type": "xpath",
--- a/configs/axibase.json
+++ b/configs/axibase.json
@@ -21,6 +21,7 @@
     },
     {
       "url": "https://axibase.com/docs/charts/",
+      "selectors_key": "charts",
       "tags": [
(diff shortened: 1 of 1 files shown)


configs/axibase.json
```diff
@@ -38,7 +38,7 @@
     "lvl3": ".content h3",
     "lvl4": ".content h4, .content table tr td:first-child",
     "lvl5": ".content h5",
-    "text": ".content p, .content li, .content pre, .content table td:nth-child(2)",
+    "text": ".content p, .content li, .content pre, .content table td:nth-child(2), .content table td:last-child",
```
s-pace: Please replace `.content table td:nth-child(2), .content table td:last-child` with only `.content table td:last-child` It will properly handle the two columns array while avoiding too much irrelevant records.

unrealwork: @s-pace The `td:last-child` doesn't catch the description of setting in [`charts docs`](https://axibase.com/docs/charts/widgets/time-chart/#widget-settings). Last cells contain a link to a live example. Is there a way to specify different text selectors for different start_urls?

s-pace: You can [use `selectors_key`](https://community.algolia.com/docsearch/config-file.html#selectors_key%2C-tailor-your-selectors) to do so

unrealwork: @s-pace It's a nice feature. But I'm not sure about the `lang` object. On which level it should be defined? Should I redefine it for each selectors' set?

s-pace: @unrealwork you can add it to every keys.

unrealwork: @s-pace Can I redefine only selectors that differ from `default` or I should do it for each selector?

s-pace: You need to define it for each of them

PR #645, opened 2018-11-16
removing suppressed page

ksigler7:
# Pull request motivation(s)
We suppressed this page because it was in closed beta, now that we are in open beta, we want it to show up in our search results.

Currently, this does not show up in our search results.

Now we want it to!


s-pace:
Deployed and live

PR #651, opened 2018-11-20
removed /reference/ from stop_urls

cglosier:
# Pull request motivation(s)

We migrated a large amount of content to `/reference` and made that part of the site much more prominent. We want it indexed for search.

/reference is excluded from search.

/reference appears in search.


s-pace:
👋 @cglosier 

Since it seems that these pages are documentation, we will allow these pages to be crawled and scraped

PR #654, opened 2018-11-23
add more selectors to text selector

marcgreenstock:
# Pull request motivation(s)

I would like the content in `td` and `code` tags to be indexed.

Content in `td` and `code` tags are not indexed.

Content in `td` and `code` tags should be indexed.

**feature**

Love it!

### DIFF
--- a/configs/mbaasy.json
+++ b/configs/mbaasy.json
@@ -11,7 +11,7 @@
     "lvl3": ".main .article h4",
     "lvl4": ".main .article h5",
     "lvl5": ".main .article h6",
-    "text": ".main .article p, .main .article li, .main .article td, .main .article code"
+    "text": ".main .article p, .main .article li, .main .article td"
   },
   "min_indexed_level": 2,
   "conversation_id": [
--- a/configs/mbaasy.json
+++ b/configs/mbaasy.json
@@ -11,7 +11,7 @@
     "lvl3": ".main .article h4",
     "lvl4": ".main .article h5",
     "lvl5": ".main .article h6",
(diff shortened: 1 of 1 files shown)


configs/mbaasy.json
```diff
@@ -11,11 +11,11 @@
     "lvl3": ".main .article h4",
     "lvl4": ".main .article h5",
     "lvl5": ".main .article h6",
-    "text": ".main .article p, .main .article li"
+    "text": ".main .article p, .main .article li, .main .article td, .main .article code"
```
s-pace: We do not recommend to index code as it introduced a lot of noise and algolia [does remove some specific character](https://www.algolia.com/doc/api-reference/api-parameters/separatorsToIndex/).

Please add a specific class like `.title` `.value` to the `td` element you want to index. There is a relation between them and it can be used as an upper `lvl`. Not every `td` are relevant and can be seen as noise

marcgreenstock: hi @s-pace i've removed the `code` selector, however the `td` elements contain a lot of relevant content, see https://docs.mbaasy.com/client_api/apple_app_store/ as an example. The table headers are defined by a `th` so they won't get indexed. The only irrelevant column I see on this page is the "Required" column and this I can sort out by placing an icon instead.

s-pace: Ok, I will add a `first-child` selector too

s-pace: b34ccaf

marcgreenstock:
nice, thanks

PR #658, opened 2018-11-28
Update liveperson.json

EdenKupe:
Adding a stop_url which we need to not be indexed right now.

# Pull request motivation(s)


configs/liveperson.json
```diff
@@ -5,6 +5,7 @@
   ],
   "stop_urls": [
     "assets"
```
s-pace: miss , to be a correct JSON

PR #661, opened 2018-11-30
Remove dead start url for Metals page

olafurpg:
The current search on https://scalameta.org/metals/ returns 404s
results. It might need to be re-indexed.

### DIFF
--- a/configs/metals.json
+++ b/configs/metals.json
@@ -1,8 +1,7 @@
 {
   "index_name": "metals",
   "start_urls": [
-    "http://scalameta.org/metals/docs/",
-    "http://scalameta.org/metals/docs/installation-contributors.html"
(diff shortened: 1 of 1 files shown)


configs/metals.json
```diff
@@ -1,8 +1,7 @@
 {
   "index_name": "metals",
   "start_urls": [
-    "http://scalameta.org/metals/docs/",
-    "http://scalameta.org/metals/docs/installation-contributors.html"
+    "http://scalameta.org/metals/"
   ],
   "stop_urls": [],
   "selectors": {
```
olafurpg: The HTML structure of the page may also have changed due to upgrading https://docusaurus.io/, the site generator we use. It might be good to re-generate these selectors.

PR #664, opened 2018-12-04
Updating Packet KB

mustela:
Hi there, we have move our KBs to another service, so we are trying to update the links :) 

Thanks!


s-pace:
👋 

@mustela thank you for the PR. It seems that the selector you have set were not CSS one.

I have updated your configuration and the changes are live.

mustela:
@s-pace you rock! Thanks!

PR #667, opened 2018-12-11
TestCafe - Add tables to the index

VasilyStrelyaev:
# Pull request motivation(s)

Currently, DocSearch does not index table contents in TestCafe docs and thus they are not searchable.

It sometimes becomes painful as we occasionally use tables to describe some minor API members with simple meaning and usage.

Table content is not searchable.

Users can search for API members described within tables.

It's more of a bug I guess.


s-pace:
`table` element are too big to be indexed. Selectors should target textual elements. I will update this PR to better handle this case.

PR #671, opened 2018-12-13
Update ppfish.json

and80506:

add config  "js_wait": 5


s-pace:
👋 @and80506 

Please provide explanation regarding this PR. Changing this parameter `js_wait` will massively extend the crawl, we would rather use another way to reach what you are looking for.

and80506:
I think that because the network speed is slow, some keywords such as tabs are not in the search results.

PR #672, opened 2018-12-13
Update cosmos_network.json

zramsay:
same as #668 

# Pull request motivation(s)


zramsay:
thanks for the fast merge!

PR #675, opened 2018-12-15
index the 3.4 directory instead of 3.3

acmetech:
Updating the config to reflect the doc versions that I would like to be searchable.

### DIFF
--- a/configs/todc-bootstrap.json
+++ b/configs/todc-bootstrap.json
@@ -5,7 +5,7 @@
       "url": "https://todc.github.io/todc-bootstrap/docs/(?P<version>.*?)/",
       "variables": {
         "version": [
-          "3.3",
+          "3.4",
           "4.1"
         ]
       }


s-pace:
Updated and deployed https://github.com/algolia/docsearch-configs/commit/4bea08d8bd4c2ae967cd891b7c39e1fd239f2b69

Please do not forget to check that the selectors do match something from your HTML. Keeping the same layout is a good way to ensure the consistency between every versions.

Changes are live :)

acmetech:
Thank you.

Is the underlying crawler configured so that I can do version specific queries using algoliaOptions: { 'facetFilters': ["version:$VERSION"]?

s-pace:
Yes @acmetech

s-pace:
Yes @acmetech

acmetech:
I receive results when I query with 'facetFilters': ["version:3.4"], but I get not results when using 'facetFilters': ["version:4.1"]

acmetech:
It's working. Thank you

PR #677, opened 2018-12-17
disable antora config

mojavelinux:
The Antora project has decided to graduate to managing it's own docsearch config and scraper.

Information about how we run it can be found here https://gitlab.com/antora/docs.antora.org/tree/master/docsearch and the CI job that runs it here https://gitlab.com/antora/docs.antora.org/blob/master/.gitlab-ci.yml.

Thanks for all the support. We couldn't have graduated without the excellent docs that Algolia provides. :beers:


s-pace:
Thank you for the great details!

PR #680, opened 2018-12-19
updates docksal selectors

shelane:
# Pull request motivation(s)
We have updated the top level selector and adjusted the rest of the levels up as well. We are hoping that this gets us the right level.

We are seeing a strange behavior with results displayed as you type.

algolia

The expected behavior is for the search results to display correctly.

We are taking a guess as to what to fix. We are starting with the selectors because we don't have the style class that was marked for lvl0. You can see our attempt at adding the Algolia code to our site: https://github.com/docksal/docksal/pull/893

### DIFF
--- a/configs/docksal.json
+++ b/configs/docksal.json
@@ -9,15 +9,15 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
-      "selector": ".links > a:nth-child(2)",
+      "selector": "#body-inner h1",
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": "#body-inner h1",
-    "lvl2": "#body-inner h2",
-    "lvl3": "#body-inner h3",
-    "lvl4": "#body-inner h4",
-    "lvl5": "#body-inner h5",
+    "lvl1": "#body-inner h2",
+    "lvl2": "#body-inner h3",
+    "lvl3": "#body-inner h4",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @shelane 
We do not face any issue with the dropdown. Must be a CSS issue.
image

s-pace:
I do not think this PR will fix this issue. Please review your CSS.

In the meantime, I have updated and deploy your config to better handled `_` as you can see here:
image

shelane:
Thanks for the response. We are using Hugo Learn theme and I have only a few custom css beyond that. There are 3 specifically to prevent the Algolia css from resizing our search box and the surrounding positioning and to prevent the page body text from bleeding through onto the search box. I’ve tried changing the order of the css, but that did nothing.

s-pace:
Please note that we crawl https://docs.docksal.io and not your staging website.

Using the CSS property `all: initial;`allow you to reset every CSS properties. Really helpful when you want to integrate one element without inherit from the global rules? Did you try it (even if browser support is not the best)

shelane:
Resetting css properities with `all: initial` did not work. However, I did discover that the Hugo theme creates and defines the `searchbox` class and the Algolia stylesheet had one as well. I found that the cleanest approach in this case was to rename our class to `searchboxed` and include all the original theme css to our custom theme css and all is well. Thanks for your help!

PR #684, opened 2018-12-22
Update bootstrap with meta

s-pace:
Waits for twbs/bootstrap#27902


configs/webdriver.json
```diff
@@ -2,24 +2,20 @@
   "index_name": "webdriver",
   "start_urls": [
     {
-      "url": "http://webdriver.io/guide.html",
+      "url": "http://beta.webdriver.io/docs/gettingstarted.html",
```
XhmikosR: It seems you modified webdriver.json here too @s-pace

s-pace: Yep under control, I will discard these changes

XhmikosR:
hey @s-pace. We'll release v4.3 later or today, 🤞 which will include the 2 meta tags.

s-pace:
Great @XhmikosR 🎉 

Do the pages from version 4.1 and 4.0 will also includes these meta tags?

Ping me when it is ready, we will deploy it asap

XhmikosR:
No, the old versions, none of them, will change, because this requires rebuilding the old docs and it's quite troublesome.

s-pace:
@XhmikosR Thanks for the details, I will do a clearer PR that will only affect the latest version.

XhmikosR:
Thanks! Please CC me there so that i can ping you.

PR #689, opened 2018-12-31
Switch level 0 selector

solugebefola:
# Objective of the pull request
Switch the level 0 selector back to using hidden index.


s-pace:
Deployed and already live

PR #691, opened 2019-01-03
Updating the URL for zephir-lang

niden:
Configuration update

PR #694, opened 2019-01-04
Revert "Revert "Updating the URL for zephir-lang""

s-pace:
Reverts algolia/docsearch-configs#693
@niden 
After the update of redirection from #691

### DIFF
--- a/configs/zephir-lang.json
+++ b/configs/zephir-lang.json
@@ -2,7 +2,7 @@
   "index_name": "zephir-lang",
   "start_urls": [
     {
-      "url": "https://docs.zephir-lang.com/(?P<lang>.*?)/(?P<version>.*?)/",
+      "url": "https://docs.zephir-lang.com/(?P<version>.*?)/(?P<lang>.*?)/",
       "variables": {
         "lang": [
           "en",


niden:
Thank you!

s-pace:
@niden there is still an issue. This new crawl is only working with `en` website.

It sounds like the URL are following this pattern, (mind the case of the two last letter):

`<root>/<version>`/ru-RU/ `301` to `<root>/<version>`/ru-RU/welcome `301` to `<root>/<version>`/ru-ru/welcome `200`

while:
`<root>/<version>`/ru-ru/ is not available (`404`).

Could you please decide to keep consistency in your languages and only lowercase or uppercase letter? This is a common issue happening on every languages ("el-GR", "es-ES", "ru-RU", "uk-UA", "zh-CN").

I would recommend you to shortcut the redirection by only using low case language.
 <root>/<version>/ru-RU/ `301` to <root>/0.11/ru-ru/ `301` to /<version>/ru-ru/welcome `200`

Once this is done, please submit a PR of your configuration with only lowercase language.

niden:
@s-pace Thank you for this. I was looking at it earlier and need to figure out a good way to do this since this happens automatically from the host. 

I will ping you when I get this resolved. Thanks again!

niden:
@s-pace I think I fixed everything. Forced all URLs to be lowercase (even in the sitemap) so that should not appear again. Curious though, on my local testing they did remain capitalized as before. Oh well.

Kindly let me know what you need me to do.

Thanks again

s-pace:
👋 @niden 

Solved by this update and crawl ^

You do not have anything to do :)

PR #695, opened 2019-01-04
add a new version of boosted documentation

Lausselloic:
Hello,

Just need to add a new version on the boosted search configuration.

Thanks in advance


s-pace:
👋 @Lausselloic 

We needed to update the `<main/>` element change from your selectors. Every is live and deployed.

PR #700, opened 2019-01-11
Update ckeditor_docs_site.json and remove ckeditor.json

msamsel:
# Pull request motivation(s)

Few weeks ago we merged two our websites into one.
> `sdk.ckeditor.com` ->`https://ckeditor.com/docs/ckeditor4/latest/examples/`

Repository which was responsible for generating `sdk.ckeditor.com` is not maintained any longer (https://github.com/ckeditor/ckeditor-sdk/). Pages are moved to different place, more details about entire process might be found here: https://github.com/ckeditor/ckeditor-docs/pull/184. Because of that change we can now remove completely one config and update current one.

Theres is obsolete config which should be updated to actual state.


s-pace:
👋 @msamsel 

Changes are live. Old index has been removed.

PR #715, opened 2019-01-29
Update sendbird.json

astyfx:
Add facet filter for versioning (to fix old data still remaining)

# Pull request motivation(s)

Old document data still remains

Clear index & Re-index data
(to fix above, we want to add version facet)

### DIFF
--- a/configs/sendbird.json
+++ b/configs/sendbird.json
@@ -36,7 +36,7 @@
   ],
   "custom_settings": {
     "separatorsToIndex": "_",
-    "attributesForFaceting": ["version(meta)"]
+    "attributesForFaceting": ["version"]
   },
   "only_content_level": true,
   "scrap_start_urls": false,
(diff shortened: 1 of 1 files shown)


configs/sendbird.json
```diff
@@ -35,10 +35,11 @@
     "#change_log"
   ],
   "custom_settings": {
-    "separatorsToIndex": "_"
+    "separatorsToIndex": "_",
+    "attributesForFaceting": ["version"]
```
s-pace: Records extracted from your website do not have any `version` attribute. Please [add a meta tag within your webpage](https://community.algolia.com/docsearch/recommended-configuration.html#introduces-global-information-as-meta-tags) in order to let our crawler extract this information.

s-pace: Perfect, Merging and deploying

PR #723, opened 2019-02-02
Update expressionengine.json

Jordan-Ellis:
# Pull request motivation(s)

We re-designed our docs, and the HTML structure has changed.

## Questions

The headings in our docs all have an `<a>` tag in them, and some of the headings can also contain inline code elements. Do the selectors need to account for them?

### DIFF
--- a/configs/expressionengine.json
+++ b/configs/expressionengine.json
@@ -5,9 +5,7 @@
       "url": "https://docs.expressionengine.com/(?P<version>.*?)/",
       "variables": {
         "version": [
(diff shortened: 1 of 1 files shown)


s-pace:
> The headings in our docs all have an <a> tag in them, and some of the headings can also contain inline code elements. Do the selectors need to account for them?

This is not required as we transform the extracted element to its text value

s-pace:
PR has been merged, update to encompass element from your table and it is live @Jordan-Ellis

PR #726, opened 2019-02-05
Update francoischalifour.com

francoischalifour:
Check the comments for more context.

### DIFF
--- a/configs/francoischalifour.json
+++ b/configs/francoischalifour.json
@@ -3,9 +3,6 @@
   "start_urls": [
     "https://francoischalifour.com/"
   ],
-  "sitemap_urls": [
-    "https://francoischalifour.com/rss.xml"
-  ],
   "stop_urls": [],
   "selectors": {
     "lvl0": "article h1",
@@ -14,7 +11,7 @@
     "lvl3": ".post h4",
     "lvl4": ".post h5",
     "lvl5": ".post h6",
-    "text": ".post p, .post li, header p"
+    "text": ".post p, .post li"
   },
   "only_content_level": true,
   "scrap_start_urls": false,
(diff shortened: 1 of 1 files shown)


configs/francoischalifour.json
```diff
@@ -3,9 +3,6 @@
   "start_urls": [
     "https://francoischalifour.com/"
   ],
-  "sitemap_urls": [
```
francoischalifour: This one using a wrong RSS file — I removed it.

configs/francoischalifour.json
```diff
@@ -14,16 +11,15 @@
     "lvl3": ".post h4",
     "lvl4": ".post h5",
     "lvl5": ".post h6",
-    "text": ".post p, .post li, header p"
+    "text": ".post p, .post li"
```
francoischalifour: I removed the headline.

configs/francoischalifour.json
```diff
@@ -14,16 +11,15 @@
...
   },
   "only_content_level": true,
   "scrap_start_urls": false,
   "conversation_id": [
     "547690260"
   ],
   "selectors_exclude": [
-    "article aside",
```
francoischalifour: This shouldn't be crawled because we now specify `.post` (not `article`).

configs/francoischalifour.json
```diff
@@ -14,16 +11,15 @@
...
-    "article p span"
+    ".post p span"
```
francoischalifour: I added more specificity with the `.post` class.

configs/francoischalifour.json
s-pace: Would be nice to implement one at some point. (Especially if you want to have a good SEO)
Fine for us

configs/francoischalifour.json
s-pace: Great

PR #728, opened 2019-02-07
Added event ref selectors

ArshRai17:
@s-pace I have added selectors for the Event reference guide. Would be great if you could take quick look. Many thanks!

# Pull request motivation(s)
Event reference guide is not currently indexed, resulting in no search results.

There are no results for event reference searches

Event reference guide should be indexed and searchable.


s-pace:
Hi @ArshRai17 ,

I would add `.event_recommended_actions` as `text` selector since these text element are not indexed yet.

s-pace:
I would also add `[class='event_label']` as `lvl3`

PR #737, opened 2019-02-15
fix baseui

gergelyke:


gergelyke:
Am I correct to assume that the `selectors_exclude` just affects what's added to the index, but those elements are still used for traversing?

gergelyke:
this is our current way of thinking about this: https://github.com/uber-web/baseui/pull/900, so we are excluding the navigation items

s-pace:
Hi @gergelyke,
Every elements matching `selectors_exclude` are completely removed from the DOM of the page. This step occurs before building the Algolia records from the page. However, these elements are still used while crawling your website. Hyperlink contained in these elements are still followed.

The `selectors` you have define are too broad. They should only target atomic and textual element. Please add a static class to define the main container of your textual payload. This needed class is `docSearch-content` as [presented here](https://community.algolia.com/docsearch/recommended-configuration.html#overview-of-a-clear-layout).

gergelyke:
@s-pace thanks for the feedback! we've just merged a pr fixing the doc site, and updated this PR too - please let me know what you think! thanks!

gergelyke:
@s-pace is there anything else missing here?  cc @maxiloc @ElPicador

s-pace:
@gergelyke The selectors you have defined are still too broad. It should only match atomic and textual DOM element. For example, `ul` should be replaced by `li`

gergelyke:
thanks @s-pace - updated

gergelyke:
hey folks! I am sorry for keep bugging you, but our search is broken for a week now :( can you please let me know if anything else is missing? @s-pace @maxiloc @ElPicador

s-pace:
@gergelyke configuration merged and deployed.

In order to add more context to your results, we would like to enhance your `lvl0`. Could you add a static `active` class to the anchor object related to the page from the left sidebar? 
For example [for this page](https://baseui.design/getting-started/comparison/), the `active` class should be set to the following highlighted element:
image
Thanks to a XPATH selector, we will be able to extract the main chapter of this page, i.e. "Getting started"

Could you also [provide a sitemap](https://community.algolia.com/docsearch/tips.html#use-a-sitemap.xml)? It will help to not miss any page from your website.

The previous changes have been deployed and the search is working now. Sorry for the delay.

PR #750, opened 2019-03-01
Updated nemtech configuration

dgarcia360:
**Pull request motivation(s)**

I'd like to exclude some pages from the search, search on tables and enable different languages.

### DIFF
--- a/configs/nemtech.json
+++ b/configs/nemtech.json
@@ -1,22 +1,10 @@
 {
   "index_name": "nemtech",
   "start_urls": [
-    {
-      "url": "https://nemtech.github.io/(?P<lang>.*?)/",
-      "variables": {
-        "lang": [
-          "ja"
-        ]
-      }
-    },
     "https://nemtech.github.io/"
   ],
   "stop_urls": [
-    "/nem2-workshop-document-notarization/",
-    "/nem2-workshop-nem-applied-to-supply-chain/",
-    "/nem2-curricular-framework/",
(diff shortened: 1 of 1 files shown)


configs/nemtech.json
```diff
@@ -2,15 +2,20 @@
...
+        ]
+      }
     },
     "https://nemtech.github.io/"
   ],
   "stop_urls": [
+    "/nem2-workshop-document-notarization/",
+    "/nem2-workshop-nem-applied-to-supply-chain/",
+    "/nem2-curricular-framework/",
+    "/nem2-sdk-typescript-javascript/",
```
s-pace: You can merge these URL by using "/nem2-"

configs/nemtech.json
```diff
@@ -2,15 +2,20 @@
   "index_name": "nemtech",
   "start_urls": [
     {
-      "url": "https://nemtech.github.io/nem2",
-      "tags": [
-        "nem2"
-      ],
-      "selectors_key": "nem2"
+      "url": "https://nemtech.github.io/(?P<lang>.*?)/",
```
s-pace: Could you set the `lang` attribute from the main `<htm/>` element within your DOM instead ?
You can find some [explanation here](https://www.w3.org/International/questions/qa-html-language-declarations).

dgarcia360:
Thanks for the review @s-pace! Just edited the JSON file.

PR #754, opened 2019-03-04
Update Front-Commerce’s config to match v2 docs markup

real34:
# Pull request motivation(s)

We have released a new version of our documentation with different markup and many more pages.
This PR contains the updated config.

The search redirects to old urls

Search results must go to the new url

I ran it locally using the docker image and it seemed to work fine (I got relevant results from https://community.algolia.com/docsearch-preview/).

That being said, the output froze at some point in my terminal so I am not sure it works perfectly. Could it be related to the [`nb_hits`](https://community.algolia.com/docsearch/config-file.html#nb_hits-special) current value detecting a spike because of new content on this new version?
I would really appreciate if you could do a test run before merging (you can override the current index)… to have a more experienced eye looking into this :smile: 

Feel free to tweak the configuration if needed (it should be close to the `apollodata` one).

Thank you very much.

### DIFF
--- a/configs/front-commerce.json
+++ b/configs/front-commerce.json
@@ -1,28 +1,23 @@
 {
   "index_name": "front-commerce",
-  "start_urls": [
-    "https://developers.front-commerce.com/docs/welcome.html",
-    "https://developers.front-commerce.com/docs/"
+  "start_urls": ["https://developers.front-commerce.com/docs/"],
+  "stop_urls": [
+    "https://developers.front-commerce.com/$",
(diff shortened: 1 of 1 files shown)


s-pace:
@real34 ?

Thank you for the PR. I have changed the `lvl0` since it was not matching anything.

Could you please update the `search by Algolia` logo please? You can find [the related style here](https://github.com/algolia/docsearch/blob/master/src/styles/_dropdown.scss#L496). Please note that our search-UI has been updated. You might want to update it altogether.

Changes are live. Let us know.

PR #755, opened 2019-03-05
[material-ui] Support v3 and v4

oliviertassinari:
I would like the search to work on v3 and v4. I'm also preparing the support for multiple search languages. We still need to work on the SEO part on material-ui.com side.

### DIFF
--- a/configs/material-ui.json
+++ b/configs/material-ui.json
@@ -1,7 +1,14 @@
 {
   "index_name": "material-ui",
   "start_urls": [
-    "https://material-ui.com/"
+    {
+      "url": "https://material-ui.com/",
+      "version": "v3"
+    },
+    {
+      "url": "https://next.material-ui.com/",
+      "version": "v4"
+    }
   ],
   "stop_urls": [
     "\\?expend-path"
@@ -15,14 +22,17 @@
     "lvl2": ".markdown-body h2",
     "lvl3": ".markdown-body h3, .markdown-body table td:first-of-type",
     "lvl4": ".markdown-body h4",
-    "text": ".markdown-body p, .markdown-body li,.markdown-body .home-text-wrap, .markdown-body table td:last-child"
+    "text": ".markdown-body p, .markdown-body li, .markdown-body table td:last-child"
   },
-  "selectors_exclude": [
-    ".anchor-link-style"
-  ],
   "only_content_level": true,
   "conversation_id": [
     "357227758"
   ],
+  "custom_settings": {
+    "attributesForFaceting": [
+      "version",
+      "language"
+    ]
+  },
   "nb_hits": 4234
 }


configs/material-ui.json
```diff
@@ -15,14 +22,17 @@
...
   },
-  "selectors_exclude": [
-    ".anchor-link-style"
-  ],
   "only_content_level": true,
   "conversation_id": [
     "357227758"
   ],
+  "custom_settings": {
+    "attributesForFaceting": [
```
s-pace: If you want to leverage these two attributes, you need to implement these meta tags within your pages:

```HTML
<meta name="docsearch:version" content="1.0.0">
```
See [more context here](https://community.algolia.com/docsearch/recommended-configuration.html#introduces-global-information-as-meta-tags)

configs/material-ui.json
```diff
@@ -1,7 +1,14 @@
 {
   "index_name": "material-ui",
   "start_urls": [
-    "https://material-ui.com/"
+    {
+      "url": "https://material-ui.com/",
+      "version": "v3"
```
s-pace: This is not the correct way to add version within your pages. You can use `tags`. 
Please implement the meta tags as highlighted in the next review instead.

configs/material-ui.json
oliviertassinari: I was confused by https://community.algolia.com/docsearch/config-file.html#using-regular-expressions:

> The beneficial side effect of using this syntax is that every records extracted from pages matching http://www.example.com/docs/en/latest will have attributes lang: en and version: latest.

configs/material-ui.json
oliviertassinari: Should I use variables instead?
https://community.algolia.com/docsearch/config-file.html#using-regular-expressions

oliviertassinari:
I should have done the necessary work on next.material-ui.com to get some new meta attributes. I will release the changes this weekend: https://github.com/mui-org/material-ui/pull/14806.

oliviertassinari:
I have implemented the i18n subfolder pattern too. The crawler should be able to index the languages (en, zh, fr, es, ru, pt) and the versions (v3, v4).

configs/material-ui.json
s-pace: Please use the meta tags instead.

oliviertassinari:
The i18n subfolder strategy is live: https://next.material-ui.com/fr/, Algolia bot should be able to crawl the data with no problem.

s-pace:
Great, will update it then.

oliviertassinari:
@s-pace 🙏 Thank you so much!

s-pace:
You are welcome,

Update from the snippet should follow this:
```html

  apiKey: '1d8534f83b9b0cfea8f16498d19fbcab',
  indexName: 'material-ui',
  inputSelector: '#docsearch-input',
});
</script>
```
- Replace $VERSION with the version you want to search on. So as of today you have: `next`. Please make sure that the pages https://material-ui.com/ includes these meta:

```html
<meta name="docsearch:version" content="master">
```

- Replace $LANGUAGE with the language you want to search on.  So as of today you have: `en`

  For example if you want to refine the search to the version "next" and the language "en" just specify: 'facetFilters': ["version:next", "language:en"]

oliviertassinari:
> Please make sure that the pages https://material-ui.com/ includes these meta:

Deploy in progress :hourglass:.

s-pace:
💯 Ping once it is achieved, I will run the crawl again :) 🚀

oliviertassinari:
Oops, I have messed up. It's live now.

s-pace:
Right, Update and you can now use the token `master` as version 🥇

oliviertassinari:
It's working great so far, I'm working on the integration. Regarding the other languages, how should we get the crawling to discover the content?



We can create a translation page that lists the content if that helps.

s-pace:
You should make sure that the right meta is available within the DOM. For example:
```html
<meta name="docsearch:language" content="zh">
```

It will be available at the next crawl. Please make sure that pages from different languages are available under a different URL.

oliviertassinari:
@s-pace I believe the metadata are correct:
- https://next.material-ui.com/zh/
- https://next.material-ui.com/ru/
- https://next.material-ui.com/es/
- https://next.material-ui.com/fr/
- https://next.material-ui.com/pt/

What do you think of us adding a /translations page that links all the versions? This way, we wouldn't have to change the configuration here when adding or removing a language.

s-pace:
@oliviertassinari thank you for the details. The issue is that there is no english pages containing a hyperlink`<a/>`that references one of these pages. You need to interact with your languages button to generate these tags. Our crawler is not able to follow these missing elements. There is 3 solutions then:
- Your [provide us a sitemap](https://www.sitemaps.org/). It will help your SEO
- You make sure that these tags are created without the need of interacting with your page.
- We add all of these URLs as `start_urls`. You will need to PR your configuration every times you want to add a new language.

Let us know which solution fits you best.

oliviertassinari:
> Your provide us a sitemap. It will help your SEO

We are already providing the `hreflang` meta attributes.

> You make sure that these tags are created without the need of interacting with your page.

Ok, let's go for a `/translations` page.

oliviertassinari:
Thanks for the assistance! We will sort the issue on our side, I don't think that we need anymore change here 🚀. I will share you the links once it's working :).

oliviertassinari:
This should do it https://next.material-ui.com/discover-more/languages/.

oliviertassinari:
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

Capture d’écran 2019-03-11 à 14 28 49

s-pace:
Yep, works great :)

oliviertassinari:
@s-pace Thanks man, it wouldn't have been possible without you!

PR #761, opened 2019-03-07
accengage-config-update

SofianeAbdelfettah:
# Pull request motivation(s)

Old Accengage Configuration

We created a new documentation website and we would like to update the configuration, so it could crawl the new website.

### DIFF
--- a/configs/accengage.json
+++ b/configs/accengage.json
@@ -1,7 +1,6 @@
 {
   "index_name": "accengage",
   "start_urls": [
-    "https://documentation.accengage.com",
     "https://documentation.accengage.com/sdk",
     "https://documentation.accengage.com/api",
     "https://documentation.accengage.com/dataflow/introduction",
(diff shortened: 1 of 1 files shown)


configs/accengage.json
```diff
@@ -1,30 +1,41 @@
 {
   "index_name": "accengage",
   "start_urls": [
-    "https://docs-alpha.accengage.com"
+    "https://documentation.accengage.com",
```
s-pace: The order of stat_urls matters. The first matching one is applied to the page, thus this object should be set as the last element as it matches every crawled page. I will do it and merge it

configs/accengage.json
```diff
@@ -1,30 +1,41 @@
...
-  "stop_urls": [],
+  "stop_urls": ["https://documentation.accengage.com/admin"],
   "selectors": {
-    "lvl0": {
-      "selector": "(//nav//a[contains(@class, 'active')])[1]/../../preceding::a[1]",
-      "type": "xpath",
-      "global": true,
-      "default_value": "Documentation"
+    "default": {
+      "lvl0": "h1, #title-text",
```
s-pace: These selectors are too broad since they can match any element within the page. They should be scope to the to smallest element gathering your whole documentation. This element should be the main container of your textual content. `.main-content` seems to be a good fit.

configs/accengage.json
```diff
@@ -1,30 +1,41 @@
...
+      "lvl1": "h2",
+      "lvl2": "h3",
+      "lvl3": "h4",
+      "lvl4": "h5",
+      "text": "p, li, ul, ol"
```
s-pace: Please add only atomic tags like `p` and `li` tags (scoped into the main element as reviewed before). `ul` and `ol` are too broad and not oriented for textual content

configs/accengage.json
```diff
@@ -1,30 +1,41 @@
...
-      "default_value": "Chapter"
-    },
-    "lvl2": "main h2",
-    "lvl3": "main h3",
-    "lvl4": "main h4",
-    "text": "main p, main li"
+    "faq": {
+      "lvl0": "p, li, h1, #title-text",
+      "lvl1": "h2",
+      "lvl2": "h3"
```
s-pace: You still need to define the common selectors from the `default` object that you want to be applied.

s-pace:
@SofianeAbdelfettah any more details required from your end? 

Let us know

configs/accengage.json
SofianeAbdelfettah: To define a common selector i must apply the global key to true, or there's another way?

s-pace: No, defining a global selector will impact the way we process the elements matching global selectors. It is not related to such use case. [See here](https://community.algolia.com/docsearch/recommended-configuration.html#recommended-selectors) for more details.

You will need to duplicate the common selectors from one object to the other.

s-pace:
@SofianeAbdelfettah please make sure you removed pages having wrong // in their URL in your [sitemap.xml](https://documentation.accengage.com/sitemap.xml). For example https://documentation.accengage.com//sdk/ios/

I have fixed this issue from our side thanks to https://github.com/algolia/docsearch-configs/commit/1f26b564eff2959fec7823e95a6088307abf1a87. Changes are live.

PR #762, opened 2019-03-07
Update sass-lang.json

nex3:
Fix the function documentation URL

### DIFF
--- a/configs/sass-lang.json
+++ b/configs/sass-lang.json
@@ -2,11 +2,11 @@
   "index_name": "sass-lang",
   "start_urls": [
     {
-      "url": "http://staging.sass-lang.com/functions/css",
+      "url": "http://staging.sass-lang.com/documentation/functions/css",
       "selectors_key": "default"
     },
     {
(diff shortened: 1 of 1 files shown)


nex3:
@s-pace It looks like this hasn't deployed yet... is that right? How long should I expect a deploy to take?

s-pace:
👋 @nex3. This PR has been deployed (fcccc58 & ddb1c69). What brings you this is mind?

s-pace:
@nex3 I think the issue you had was that pages available from` http://staging.sass-lang.com/documentation/functions/*` were not retrievable.

It is solved thanks to 8d25677. The `text` selector wasn't matching anything. I have updated it and these page are now correctly indexed.

Please note that the `lvl1` was not matching anything either because the class is `.docSearch-function` instead of `docSearch-function`.

image
8d25677 solved this too. You might want to resolves this. Please PR your configuration it once it is live. 

Please note that pages available from `http://staging.sass-lang.com/styleguide/*` are not indexed since neither of these pages implement the class `docSearch-function`

PR #767, opened 2019-03-11
Update facetwp.json

mgibbs189:
Apologies, this was an oversight in my part. Our tutorials were no longer getting indexed because their URLs aren't prefixed with `/documentation/`. This change re-adds `/` as a start URL.


s-pace:
👋 @mgibbs189 ,

Configuration enhanced and index updated 🚀

PR #771, opened 2019-03-13
WIP: update spacy.json

ines:
# Pull request motivation(s)

⚠️ This shouldn't be merged just yet – the site described by this update isn't officially live yet. But I already wanted to run this by you @s-pace to make sure the config is correct (and give me time to fix things if needed).

The current config describes the structure of the live site at [spacy.io](https://spacy.io). 

We're about to launch a new version of the site (probably this week). The staging site that also features the search can be viewed here: https://spacy.netlify.com

It looks very similar, but it's actually been re-written from scratch and I've also added various markup improvements to make the crawling easier. Once the new site is live, it should be crawled and the search results should reflect the new content.

### DIFF
--- a/configs/spacy.json
+++ b/configs/spacy.json
@@ -11,12 +11,7 @@
     "https://spacy.io/sitemap.xml"
   ],
   "selectors": {
-    "lvl0": {
-      "selector": "//*[contains(@class,'is-active')]/preceding::*[contains(@class,'h0')][1]",
-      "type": "xpath",
-      "global": true,
-      "default_value": "Documentation"
-    },
+    "lvl0": ".h0",
     "lvl1": "article header h1 .heading-text",
     "lvl2": "article h2 .heading-text",
     "lvl3": "article h3 .heading-text",
--- a/configs/spacy.json
+++ b/configs/spacy.json
@@ -1,7 +1,7 @@
 {
   "index_name": "spacy",
   "start_urls": [
-    "https://spacy.io"
+    "https://spacy.io/"
   ],
   "stop_urls": [
     "\\?",
@@ -12,7 +12,7 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "//*[contains(@class,'sidebar')]//*[contains(@class,'is-active')][1]/preceding::*[contains(@class,'h0')][1]",
+      "selector": "//*[contains(@class,'is-active')]/preceding::*[contains(@class,'h0')][1]",
       "type": "xpath",
       "global": true,
       "default_value": "Documentation"
@@ -37,5 +37,5 @@
     "753259253"
   ],
   "js_render": true,
-  "nb_hits": 7203
+  "nb_hits": 7117
 }
--- a/configs/spacy.json
+++ b/configs/spacy.json
@@ -7,9 +7,12 @@
     "\\?",
     "index$"
   ],
+  "sitemap_urls": [
+    "https://spacy.io/sitemap.xml"
+  ],
   "selectors": {
     "lvl0": {
-      "selector": "//*[contains(@class,'sidebar')]//*[contains(@class,'is-active')][1]/preceding::*[contains(@class,'.h0')][1]",
+      "selector": "//*[contains(@class,'sidebar')]//*[contains(@class,'is-active')][1]/preceding::*[contains(@class,'h0')][1]",
       "type": "xpath",
       "global": true,
       "default_value": "Documentation"
@@ -34,5 +37,5 @@
     "753259253"
   ],
   "js_render": true,
-  "nb_hits": 6198
+  "nb_hits": 7203
 }
--- a/configs/spacy.json
+++ b/configs/spacy.json
@@ -11,7 +11,10 @@
     "https://spacy.io/sitemap.xml"
   ],
   "selectors": {
-    "lvl0": ".h0",
+    "lvl0": {
+        "selector": ".h0",
+        "defaultValue": "Documentation"
+    },
     "lvl1": "article header h1 .heading-text",
     "lvl2": "article h2 .heading-text",
     "lvl3": "article h3 .heading-text",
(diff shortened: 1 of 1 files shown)


configs/spacy.json
```diff
@@ -9,28 +9,26 @@
...
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": {
-      "selector": "//article//h1/text()",
-      "type": "xpath"
-    },
-    "lvl2": "article h2",
-    "lvl3": "article h3",
-    "lvl4": "article h4, article table td:first-child, .c-table__row--foot td:nth-child(2)",
```
ines: `.c-table__row--foot td:nth-child(2)` is usually the return type of the documented function/method, which isn't very relevant in our case. So I've removed that.

s-pace:
Could you please grant me access [to your repo](https://github.com/ines/docsearch-configs)?  I have done some changes and I would like to push them.

You can try it live [on this page](https://community.algolia.com/docsearch-preview/).
appID: Y7BGGRAPHC
apiKey: f7dbcd148fae73db20b6ad33d03cc9e8
indexName: dev_spacy_netlify

The selector should be: `//*[contains(@class,'sidebar')]//*[contains(@class,'is-active')][1]/preceding::*[contains(@class,'h0')][1]`

But even with the corrected selector, t doesn't match anything. It seems that the `is-active` required some javascript in order to be correctly generated. Could you set it at the server-side?

ines:
@s-pace Sure, done! And thanks 👍

> It seems that the `is-active` required some javascript in order to be correctly generated. Could you set it at the server-side?

Ah, damn, I think you're right. I'll fix this later – in the sidebar, we definitely know the current page during rendering, so this shouldn't be a problem. ([Source is here](https://github.com/explosion/spaCy/tree/develop/website) btw.)

s-pace:
@ines thank you for the details,

Updated, please ping me here to let me know once this `is-active` class is set from the server side. I will run the crawl again.

ines:
@s-pace Ok, should be fixed now!

s-pace:
@ines it is not working yet. Could you make sure to deploy it please?

We need "is-active" & "h0" class set without any javascript rendering. We couldn't extract it from https://spacy.netlify.com/usage/vectors-similarity for example

ines:
@s-pace Hmm, that's strange 🤔 I made sure to check the compiled static HTML and it was in there. If I look at the source of the page, it's also there (had to do a hard refresh, though): 

`view-source:https://spacy.netlify.com/usage/vectors-similarity/`

s-pace:
Just try to match the CSS selector `.is-active` from the developer tool at `view-source:https://spacy.netlify.com/usage/vectors-similarity/` and you will see that this is not matching anything 🤷‍♂️

s-pace:
Might be that we crawl 'https://spacy.netlify.com/usage/vectors-similarity'. Please mind that there is not the trailing slash from the URL.

s-pace:
It seems that the class attribute is not correctly closed:
image

ines:
Hmm, could it be some interaction of the service worker / caching? This is all very confusing. It's a pretty regular Gatsby build, so I don't see what could be going so wrong here.

I tried matching the selector on that page with JavaScript disabled and no service worker and it worked for me:

Screenshot 2019-03-14 at 14 33 43

If there's anything else I can try or do, let me know.

s-pace:
Alright, I have found a ways to make it work given your source.

You can try it live on this page.
appID: Y7BGGRAPHC
apiKey: f7dbcd148fae73db20b6ad33d03cc9e8
indexName: dev_spacy_netlify

Please add a unique `name` or `id` attributes to your heading so that the redirection will be more accurate. You can find [more guidance here](https://community.algolia.com/docsearch/tips.html#add-anchors-to-headings).

ines:
@s-pace Thanks so much for your help on this! I'll double-check the headings and make sure they all have ids. I've also found some sections that should be excluded, so I'll add a custom class to them.

Btw, one question: Is there a way to read the text from a data attribute instead of the element's text content? For most pages, the preceeding `.h0` in the sidebar is fine. But for [project subpages in `/universe`](https://spacy.netlify.com/universe/project/lemmy), that should be "Universe" (not "Projects" or "Education").

I was thinking about solving this by adding something like `data-algolia-h0="Universe"` etc. across all sidebars and then read from that attribute instead of the text. Does this make sense? Or is there a better way?

s-pace:
@ines you are welcome

This is doable, the XPATH selector will be something like: `//*[contains(@class,'is-active')]/@data-algolia-h0`. However, using HTML element is style better since you can use broader set of character (like emoji). You can create a unique hidden element for every pages that has the CSS proper `display:none;` to define this lvl0. It should be define as follow:
```html
<h1 class=''@data-algolia-h0">VALUE OF LEVEL 0</h1>
```

ines:
@s-pace Thanks! I updated a few more things, including the IDs, and changed the sidebar to generate the active heading at compile time. So we actually have a clean `.h0` selector now 🎉

configs/spacy.json
```diff
@@ -11,12 +11,7 @@
     "https://spacy.io/sitemap.xml"
   ],
   "selectors": {
-    "lvl0": {
-      "selector": "//*[contains(@class,'is-active')]/preceding::*[contains(@class,'h0')][1]",
-      "type": "xpath",
-      "global": true,
-      "default_value": "Documentation"
-    },
+    "lvl0": ".h0",
```
s-pace: @ines Why did you remove this? Your current `.h0` element is not only the active element. I will reset the XPATH I have defined

s-pace: My bad, I needed to hard refresh. Make sense

s-pace:
👍 @ines , You can try it out on `dev_spacy_netlify`

ines:
@s-pace Just switched the site over on spacy.io, so this can be merged now!

s-pace:
Changes are live 🎉

PR #775, opened 2019-03-14
te3201_2019.json: tweak level 0 headings

damithc:
I'm trying to tweak the level 0 headings in search results. The top one should be `Programming Textbook` and the second one should be `SE Textbook`.

image

On a related note, I'm wondering if you can help to tweak the config file so that the lower level headings of [the admin info page](https://nus-te3201.github.io/2019/admin/te3201-admin.html) can be captured in results too.
image

### DIFF
--- a/configs/te3201_2019.json
+++ b/configs/te3201_2019.json
@@ -51,11 +51,16 @@
       "text": ".website-content p, .website-content li, .website-content .table td"
     },
     "programming": {
-      "lvl0": ".website-content h1",
-      "lvl1": ".website-content h3",
-      "lvl2": ".website-content h4",
-      "lvl3": ".website-content h5",
-      "lvl4": ".website-content h6",
+       "lvl0": {
+        "selector": "",
+        "global": true,
+        "default_value": "Programming Textbook"
+      },
+      "lvl1": ".website-content h1",
+      "lvl2": ".website-content h3",
+      "lvl3": ".website-content h4",
+      "lvl4": ".website-content h5",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @damithc,

The issue raised in the note is solved.

Please keep consistency in the header you use. It seems more relevant to use `<h1/>`, `<h2/>`, `<h3/>`, `<h4/>`, `<h5/>`, `<h6/>` in succession instead of not using some of them from one page to another. It will ease the selector to define and will not be intricate.

Changes are live.

damithc:
> The issue raised in the note is solved.

Thanks for the help @s-pace 

> Please keep consistency in the header you use. It seems more relevant to use `<h1/>`, `<h2/>`, `<h3/>`, `<h4/>`, `<h5/>`, `<h6/>` in succession instead of not using some of them from one page to another. It will ease the selector to define and will not be intricate.

Will do in the future.

What needs to be done to get the search result in the box to appear under an `Admin` heading, similar to results appearing under `Programming Textbook` and `SE Textbook` headings? i.e., divide the search results into three categories, `Admin`, `Programming Textbook` and `SE Textbook` 
image

s-pace:
Please add a unique element that can be hidden (CSS rule `display:none;`) as follow:
```html
<div class='DocSearch-lvl0'>VALUE OF YOUR TITLE</div>
```
Then we will define this lvl0 selector instead:

```json
    "lvl0": {
      "selector": ".DocSearch-lvl0",
      "global": true,
    },
```

PR #778, opened 2019-03-15
Update quasar-framework.json

rstoenescu:
# Pull request motivation(s)
Need to index table contents too on Quasar Framework documentation.

Not indexing tables.

Indexing table content.


rstoenescu:
Thank you!

PR #779, opened 2019-03-15
te3201_2019.json: add a level0 selector for 'Admin Info'

damithc:
Following from [this comment](https://github.com/algolia/docsearch-configs/pull/775#issuecomment-473192414), I'm trying to divide the search results into three sections. @s-pace would this work too? If not, I'll try adding a hidden element to relevant pages as suggested.

### DIFF
--- a/configs/te3201_2019.json
+++ b/configs/te3201_2019.json
@@ -10,6 +10,7 @@
     },
     {
       "url": "https://nus-te3201.github.io/2019/admin/",
+      "selectors_key": "admin",
       "tags": [
         "admin"
       ]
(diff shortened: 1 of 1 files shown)


s-pace:
@damithc This will work but it will not scale if you want to add further sections. It might end with a long configuration then.

damithc:
> @damithc This will work but it will not scale if you want to add further sections. It might end with a long configuration then.

Understood. Thanks @s-pace 
Shall we go with this approach for the time being? The reason: at the moment our website building tool is not optimized for adding hidden elements to a group of files.

s-pace:
👋 @damithc makes sense. Deployed.

damithc:
Works! Thanks @s-pace

PR #780, opened 2019-03-17
te3201_2019.json: add js_render option

damithc:
I'm adding `js_render` option because some of our contents are rendered on the client side. Our objective is to prevent indexing contents that are not visible on the page e.g., content in modal windows, collapsed panels, etc.

### DIFF
--- a/configs/te3201_2019.json
+++ b/configs/te3201_2019.json
@@ -1,5 +1,7 @@
 {
   "index_name": "te3201_2019",
+  "js_render": true,
+  "js_wait": 10,
   "start_urls": [
     {
       "url": "https://nus-te3201.github.io/2019/schedule/",


s-pace:
👋 @damithc,
These changes are too resources demanding. Please avoid to rely on client-side rendering in order to correctly build you page.

damithc:
@s-pace is it the 10 second delay that is the problem, or `"js_render": true` is not acceptable at all?

s-pace:
The delay `10 second` is way too big.

damithc:
@s-pace That number was somewhat arbitrary. We can try a lower number. What's the highest number acceptable? Or you can recommend a number to start with and see if that works.
Again, indexing based on client-side rendering is a temporary measure until we find a better way to shield hidden content from indexing. We are hoping to implement a way to auto-tag all hidden content so that we can go back to index based on server-side rendering.

s-pace:
@damithc 
Using `js_render:true` should be avoided as much as possible.

Please implement this auto-tag for hidden client on the server-side as quick as possible. You should also avoid having duplicated content available from one page to another. For instance, please avoid to duplicate information from [this page](https://nus-te3201.github.io/2019/schedule/week8/) to [this one](https://nus-te3201.github.io/2019/programming/toc/thirdparty.html#appendix-a-installing-3rd-party-modules).

I have updated your configuration in order to reach the goals of this PR.

damithc:
Noted @s-pace 
Thanks for the help.

PR #784, opened 2019-03-19
Remove ghost.json

ErisDS:
Submitting this to cleanup and make sure unnecessary indexing isn't going on.

- We (Ghost) used docsearch as a demo during development, but newdocs.ghost.org no longer exists.
- The now live https://docs.ghost.org uses Algolia with a custom implementation of indexing and react instantsearch https://github.com/TryGhost/docs


s-pace:
Awesome, thank you very much @ErisDS for this care. 🙏 

Let us know if you need anything.

PR #795, opened 2019-03-29
[patternfly] update site url and selectors

jschuler:
Updated some selectors and added synonyms

### DIFF
--- a/configs/patternfly.json
+++ b/configs/patternfly.json
@@ -1,10 +1,16 @@
 {
   "index_name": "patternfly",
   "start_urls": [
-    "https://v2.patternfly.org/"
+    {
+      "url": "https://www.patternfly.org/v4/get-started/",
+      "selectors_key": "getStarted"
+    },
+    {
+      "url": "https://www.patternfly.org/v4/"
+    }
   ],
   "sitemap_urls": [
-    "https://v2.patternfly.org/sitemap.xml"
+    "https://www.patternfly.org/v4/sitemap.xml"
   ],
   "stop_urls": [
     "/blog/",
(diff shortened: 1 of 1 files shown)


configs/patternfly.json
```diff
@@ -13,30 +13,317 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "header nav [aria-current]",
+      "selector": "header nav li a",
```
s-pace: Why did you remove the `[aria-current]`selector? Globals selectors should only match one element from the page, the active one.

configs/patternfly.json
```diff
@@ -13,30 +13,317 @@
...
       "global": true,
       "default_value": "Documentation"
     },
     "lvl1": {
-      "selector": "#page-sidebar [aria-expanded='true']",
+      "selector": "#page-sidebar li a",
```
s-pace: Same review as lvl0

configs/patternfly.json
```diff
@@ -13,30 +13,317 @@
...
       "global": true,
       "default_value": "Chapter"
     },
-    "lvl2": "main h1",
-    "lvl3": "main h2",
-    "lvl4": "main h3",
-    "lvl5": "main h4",
-    "lvl6": "main h5",
+    "lvl2": {
+      "selector": "#page-sidebar section li a",
```
s-pace: Same review as lvl0

configs/patternfly.json
```diff
@@ -13,30 +13,317 @@
...
   "conversation_id": [
     "772189694"
   ],
-  "nb_hits": 15754
+  "nb_hits": 15754,
+  "custom_settings": {
+    "separatorsToIndex": "_",
+    "synonyms": [
+      [
+        "about modal",
```
s-pace: The scraper supports only supports regular *one* word synonyms. Please remove the 2 words one.

configs/patternfly.json
```diff
@@ -13,30 +13,317 @@
...
-    "separatorsToIndex": "_"
-  },
   "conversation_id": [
     "772189694"
   ],
-  "nb_hits": 15754
+  "nb_hits": 15754,
+  "custom_settings": {
+    "separatorsToIndex": "_",
+    "synonyms": [
```
s-pace: You can only have up to 20 synonyms if you need more details, you can [have a look here](https://community.algolia.com/docsearch/config-file.html#custom_settings-optional)

configs/patternfly.json
jschuler: Thank you, I will revert this or at least ensure it targets just the active one

configs/patternfly.json
jschuler: Okay, in the examples I saw something like `ECMAScript 6`, so that is also not valid, right?

configs/patternfly.json
jschuler: 20 synonym arrays or each of the arrays is limited to 20 synonyms? Alternatively I can set algoliaOptions.synonyms with docsearch from the client?

configs/patternfly.json
s-pace: Good point, will fix it. Thanks for pointing out.

configs/patternfly.json
s-pace: The array of synonyms must be inferior or equals to 20. You can not edit synonym at the search time. Only at the indexing one.

jschuler:
@s-pace Thank you, I have removed the custom synonyms for now, instead modifying the site url and the selectors

s-pace:
@jschuler I have fine tuned the configuration to avoid extracting too many element from your page. Changes are deployed and live. Let us know if you want to change anything.

PR #797, opened 2019-03-29
Update apereo.json for CAS search configuration

mmoayyed:
Adds two additional versions into the search config.


s-pace:
👋 @mmoayyed seems that these version are not release yet. Is it expected?

mmoayyed:
Yes they are not, though will be in the very near future (few months). Just planning ahead. I hope that is not an issue?

s-pace:
No it is fine but the best way would be to do it only once the version is available.

PR #802, opened 2019-04-02
fix start_urls for graphql-compose

nodkz:
# Pull request motivation(s)

Make crawling alive.

Search does not work properly. When click by search results it redirects to the wrong location (to 404 page).

### DIFF
--- a/configs/graphql-compose.json
+++ b/configs/graphql-compose.json
@@ -1,8 +1,8 @@
 {
   "index_name": "graphql-compose",
   "start_urls": [
-    "https://graphql-compose.github.io/docs/en/intro-quick-start.html",
-    "https://graphql-compose.github.io/docs"
+    "https://graphql-compose.github.io/docs/intro/quick-start.html",
+    "https://graphql-compose.github.io"
   ],
   "stop_urls": [],
   "selectors": {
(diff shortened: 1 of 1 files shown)


nodkz:
On graphql-compose site was removed languages

Old url:
https://graphql-compose.github.io/docs/en/intro/quick-start.html

New url:
https://graphql-compose.github.io/docs/intro/quick-start.html

configs/graphql-compose.json
```diff
@@ -1,8 +1,8 @@
 {
   "index_name": "graphql-compose",
   "start_urls": [
-    "https://graphql-compose.github.io/docs/en/intro-quick-start.html",
-    "https://graphql-compose.github.io/docs"
+    "https://graphql-compose.github.io/docs/intro/quick-start.html",
+    "https://graphql-compose.github.io"
```
s-pace: Can you keep the "/docs" so that we only crawl documentation pages?

nodkz: Page `https://graphql-compose.github.io/docs` does not exists (404 code).
Is this ok that it in `start_urls`? It should be removed or it will be used as prefix check for urls?

I may revert last commit if need to remove `https://graphql-compose.github.io/docs`

PR #843, opened 2019-05-14
Update rsshub.json

DIYgod:
# Pull request motivation(s)

Fix `stop_urls` value for `rsshub.json`

`*.html` is my correct urls, but it's added in `stop_urls`

The steps to reproduce:

1. Search `twitter` in <https://docs.rsshub.app>

1. Click the first item

1. Page will jump to `https://docs.rsshub.app/social-media#twitter`, but the correct url is `https://docs.rsshub.app/social-media.html#twitter`

Page jumps to correct url

### DIFF
--- a/configs/rsshub.json
+++ b/configs/rsshub.json
@@ -3,9 +3,7 @@
   "start_urls": [
     "https://docs.rsshub.app/"
   ],
-  "stop_urls": [
-    "\\.html$"
-  ],
+  "stop_urls": [],
   "selectors": {
     "lvl0": {
       "selector": "p.sidebar-heading.open",


s-pace:
👋 @DIYgod 

The issue here is that you will introduce a lot of duplicates. Could you make sure pages without html redirect thank to a `30X` code to the html version?

DIYgod:
Sorry, it's controlled by netlify, I don't find any configuration for it.

But there are no such links in the pages, I think the crawler won't fetch the no `.html` page.

There is a correct expamle: <https://fakerjsdocs.netlify.com>, it uses the same docs system and the same backend(netlify) as <https://rsshub.app>, it's [configured with empty stop_urls](https://github.com/algolia/docsearch-configs/blob/master/configs/fakerjsdocs.json#L6) and works well.

s-pace:
It is fetching them this is why I have added these `stop_urls`

DIYgod:
😅 But it works for `fakerjsdocs`, do you have any other way to fix it? Maybe listing all no `.html` urls to `stop_urls`?

s-pace:
Yes, add a redirect :)

DIYgod:
It works well now, thanks!

PR #844, opened 2019-05-14
Update flowable.json

dbmalkovsky:
# Pull request motivation(s)

We are still not seeing some of the content so the changes made are to be more like the documentation site for the tools (antora.org) that we are using.  The antora file is: https://gitlab.com/antora/docs.antora.org/blob/master/docsearch/config.json

Missing content.

Not sure if the "attributesToRetrieve" and "customRanking" element in the antora config are needed.

### DIFF
--- a/configs/flowable.json
+++ b/configs/flowable.json
@@ -177,20 +177,18 @@
     "https://documentation.flowable.com/sitemap.xml"
   ],
   "selectors": {
-    "default": {
-      "lvl0": {
-        "global": true,
-        "selector": ".navigation-explore .context"
-      },
-      "lvl1": ".doc > h1:first-of-type",
-      "lvl2": ".doc .sect1 > h2:first-child",
-      "lvl3": ".doc .sect2 > h3:first-child",
-      "lvl4": ".doc .sect3 > h4:first-child",
-      "text": ".doc p, .doc dt, .doc td.context, .doc th.tableblock",
-      "version": {
-         "selector": "span.version",
-         "global": true
-      }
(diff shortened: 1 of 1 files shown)


configs/flowable.json
```diff
@@ -177,25 +177,20 @@
...
+    "default": {
+      "lvl0": {
+        "global": true,
+        "selector": ".navigation-explore .context"
+      },
+      "lvl1": ".doc > h1:first-of-type",
+      "lvl2": ".doc .sect1 > h2:first-child",
+      "lvl3": ".doc .sect2 > h3:first-child",
+      "lvl4": ".doc .sect3 > h4:first-child",
+      "text": ".doc p, .doc dt, .doc td.context, .doc th.tableblock"
```
s-pace:
```suggestion
      "text": ".doc p, .doc dt, .doc td.context, .doc th.tableblock",
```

configs/flowable.json
```diff
@@ -177,25 +177,20 @@
...
-    "lvl2": ".doc h1",
-    "lvl3": ".doc h2",
-    "lvl4": ".doc h3",
-    "lvl5": ".doc h4",
-    "lvl6": ".doc h5",
-    "text": ".doc p, .doc li",
-    "version": {
-      "selector": "span.version",
-      "global": true
+    "default": {
```
s-pace: Do not use a default selector_key if none of them are defined

dbmalkovsky: Thanks for your help.

PR #855, opened 2019-05-22
Update morningstar_design.json

norfali:
Fix `lvl0` selector to ensure it picks up active item from top navigation


s-pace:
Updated

PR #860, opened 2019-05-24
Update wepay_developer.json

Sarkar:
# Pull request motivation(s)

Adding in a new path to index. Currently, the implementation doesn't search our API reference content, and I think this fixes it.

Current behavior only shows search results for what is under /docs on developer.wepay.com. 

We should be showing search results for API calls on developer.wepay.com/api


Sarkar:
Typo 🤦‍♂. Let me fix.

PR #866, opened 2019-05-28
Update vant.json

chenjiahan:
1. add `js_wait` to improve the accuracy of search results
2. add `changelog` to `stop_urls `

Thanks!

### DIFF
--- a/configs/vant.json
+++ b/configs/vant.json
@@ -11,7 +11,9 @@
       }
     }
   ],
-  "stop_urls": [],
+  "stop_urls": [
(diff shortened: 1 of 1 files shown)


s-pace:
Thanks for the PR @chenjiahan 

Please try to avoid to rely on javascript to render your page. We highly recommend to use server side rendering since it make the crawl longer. We might stop it if the crawl is longer than 45 minutes.

chenjiahan:
Got it, we will consider to use SSR to render our site ^_^

PR #879, opened 2019-06-08
updated: riot indexing

GianlucaGuarini:
# Pull request motivation(s)

The Riot.js website was redesigned

The old site content are still indexed in the search point to 404 sites

The new site content should be indexed

### DIFF
--- a/configs/riotjs.json
+++ b/configs/riotjs.json
@@ -1,30 +1,26 @@
 {
   "index_name": "riotjs",
   "start_urls": [
-    "https://riot.js.org/guide/",
-    "https://riot.js.org/api/",
-    "https://riot.js.org/faq/",
-    "https://riot.js.org/play/"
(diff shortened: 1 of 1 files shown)


configs/riotjs.json
```diff
@@ -1,31 +1,27 @@
...
+    "lvl3": "main h3",
+    "lvl4": "main h4",
+    "lvl5": "main h5",
+    "text": "main p, main li"
   },
   "conversation_id": [
     "580098926"
   ],
   "selectors_exclude": [
-    "#toc"
```
s-pace: Why did you remove that?

GianlucaGuarini: because that id doesn't exist anymore

s-pace:
Deployed ✅

GianlucaGuarini:
@s-pace thank you

PR #880, opened 2019-06-10
Update environmental-kurashev.json

antohabio:
# Pull request motivation(s)


s-pace:
Deployed #880

PR #888, opened 2019-06-11
Trying to get code blocks indexed

dbmalkovsky:
# Pull request motivation(s)
I am attempting to get code blocks indexed so we can go live with this search tool with our latest release.

Thanks for your help and understanding.

Searches for text of values within JSON or Java will be found.

If it is easy to launch another crawl after this push it will be much appreciated.  Thanks.


s-pace:
It is not recommended to do so . :)

PR #889, opened 2019-06-11
Update branchmetrics.json

aopreaa:
# Pull request motivation(s)

### DIFF
--- a/configs/branchmetrics.json
+++ b/configs/branchmetrics.json
@@ -1,13 +1,11 @@
 {
-  "index_name": "branchmetrics",
-  "start_urls": [
-    {
-      "url": "https://docs.branch.io/(?P<lang>.*?)/",
-      "variables": {
-        "lang": ["", "ko", "ja", "zh"]
-      }
-    }
+ "index_name": "branchmetrics",
+ "start_urls": [
+   "https://docs.branch.io/"
   ],
+  algoliaOptions: {
+    'facetFilters': ["lang:en", "version:latest"]
+  }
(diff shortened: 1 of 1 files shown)


configs/branchmetrics.json
```diff
@@ -1,13 +1,11 @@
...
-      "url": "https://docs.branch.io/(?P<lang>.*?)/",
-      "variables": {
-        "lang": ["", "ko", "ja", "zh"]
-      }
-    }
+ "index_name": "branchmetrics",
+ "start_urls": [
+   "https://docs.branch.io/"
   ],
+  algoliaOptions: {
```
s-pace: 👋 @aopreaa ,

This parameter is here to filter [from your JS snippet](https://community.algolia.com/docsearch/config-file.html#using-custom-tags), it should not be included it here.

s-pace:
```html

  apiKey: '887017d17a2c1e9490ad2c44caa7e817',
  indexName: 'branchmetrics',
  algoliaOptions: { 'facetFilters': ["lang:$LANG", "version:$VERSION"] },
});
</script>
```

- Replace $LANG with the lang you want to search on. The list of possible lang is hardcoded in the config. So as of today you have: en, ko, zh

- Replace $VERSION with the version you want to search on. The list of possible version is hardcoded in the config. So as of today you have: latest

  For example if you want to refine the search to the lang "en" and the version "latest" just specify: 
```js
'facetFilters': ["lang:en", "version:latest"]
```

PR #894, opened 2019-06-16
Start crawling at root

JoshOrndorff:
Search should also index pages like these:
* https://substrate.dev/tutorials/
* https://substrate.dev/recipes/

# Pull request motivation(s)
We would like search to index our entire domain. While much of the content is hosted in docusaurus, we would also like results for non docusaurus content like the links above.

Currently only the documentation that lives directly in docusaurus is being indexed by search

The entire domain should be indexed.

No

Is is possible to crawl subdomains? For example we will eventually serve some content at crates.substrate.dev and would like that indexed as well.

### DIFF
--- a/configs/substrate.json
+++ b/configs/substrate.json
@@ -1,6 +1,7 @@
 {
   "index_name": "substrate",
   "start_urls": [
+    "https://substrate.dev",
     "https://substrate.dev/docs/",
     "https://substrate.dev/docs/en/quickstart/installing-substrate"
   ],


JoshOrndorff:
cc @shawntabrizi @kaichaosun

JoshOrndorff:
@s-pace Thanks for merging this PR for us. It seems our index still does not contain results for non-docusaurus portions of our documentation such as https://substrate.dev/rustdocs/v1.0/

Is this likely to be a problem with [our selectors](https://github.com/algolia/docsearch-configs/blob/master/configs/substrate.json#L13)?

Also, when I start to diagnose this, is there a faster way to get feedback than submitting a PR here and then waiting 24 hours after it is merged for another crawl?

s-pace:
Yes it is

s-pace:
I update your index with the PR asap. Basically 10 sec after

PR #896, opened 2019-06-17
Update unraveldata.json

ada-unravel:
# Pull request motivation(s)
Try to improve DocSearch efficiency. Right now it's returning undesirable results.

For example, if we search for "spark-s" it returns "Spark Sensor". If we search for <property prefix> it returns <properties that don't include this prefix>. Both of these seem to be a problem with the typo tolerance settings.

We need to tighten up the typo tolerance so that we aren't getting matches that make no sense.

### DIFF
--- a/configs/unraveldata.json
+++ b/configs/unraveldata.json
@@ -19,9 +19,27 @@
     }
   },
   "custom_settings": {
+    "allowTyposOnNumericTokens": "false",
     "attributesForFaceting": [
       "lang"
-    ]
+    ],
+	"allowTyposOnNumericTokens" : "false",
+	"hitsPerPage": "25",
+	"minWordSizefor1Typo": 20,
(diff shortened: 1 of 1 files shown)


s-pace:
Did you try this configuration first?

ada-unravel:
No, but I'm willing to try it--it would be great if we could test our config changes--please tell me how.

s-pace:
https://community.algolia.com/docsearch/run-your-own.html

ada-unravel:
Running DocSearch myself is going to take a long time for me to set up.

s-pace:
Use the docker image

ada-unravel:
ok. I'll reply again asap

s-pace:
Closing it in the meantime. Feel free to reopen

PR #901, opened 2019-06-23
Updated configuration for ComponentKit

fabiomassimo:
# Pull request motivation(s)

The ComponentKit website also provides an API reference source that it's valuable to index as well.

Content from `https://componentkit.org/appledoc/*/*` will show up in search

### DIFF
--- a/configs/componentkit.json
+++ b/configs/componentkit.json
@@ -1,19 +1,9 @@
 {
   "index_name": "componentkit",
   "start_urls": [
-    {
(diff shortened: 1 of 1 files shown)


fabiomassimo:
Hi @s-pace,
Thank you for taking the time for reviewing this.

Are the selectors you are mentioning the `"selectors_key": "doc" / "api"` ? I think I have properly specified them, or am I missing something else?

PR #906, opened 2019-06-26
Fix missing comma

SeBassTian23:
# Pull request motivation(s)

Forgot to add in the first pull request, so this is the fix to have a valid JSON


s-pace:
Already solved

PR #915, opened 2019-07-07
Update Substrate Configuration

shawntabrizi:
The Substrate Developer Hub is a GitHub Pages based website for documenting Substrate, an open source blockchain development framework.

We use a number of different tools for documentation including:
- [Docusaurus](https://docusaurus.io/) | [[Example]](https://github.com/substrate-developer-hub/substrate-developer-hub.github.io)
- [Docsify](https://docsify.js.org/#/) | [[Example]](https://github.com/substrate-developer-hub/substrate-collectables-workshop)
- [RustDoc](https://doc.rust-lang.org/rustdoc/what-is-rustdoc.html) | [[Example]](https://github.com/substrate-developer-hub/rustdocs)
- [MDBook](https://github.com/rust-lang-nursery/mdBook) | [[Example]](https://github.com/substrate-developer-hub/recipes)

These various projects are all "combined together" using the veil of GitHub pages which hosts all of these documentation pages under the shared parent URL `substrate.dev`.

We would like to use Algolia as a cross project search engine for all of our Substrate docs.

This PR removes the `sitemap_urls` and `sitemap_alternate_links` properties since these are not correctly generated for our multiple different projects. Instead, we would like to rely on the Algolia crawler to follow our links to find the various docs under the `substrate.dev` domain.

Further, this PR adds `js_render`, `js_wait`, and `use_anchors` specifically to address the crawling and documentation of [Docsify](https://docsify.js.org/#/), which is a client rendered website which does use `#` urls and may take a second or two to load. I have looked into statically generating these pages, but Docsify does not currently support this. It possible I may move to an alternate solution as a result, but for now, this is what I would like to try.

@JoshOrndorff @s-pace 

https://github.com/substrate-developer-hub/substrate-developer-hub.github.io/issues/145

### DIFF
--- a/configs/substrate.json
+++ b/configs/substrate.json
@@ -3,10 +3,17 @@
   "start_urls": [
     {
       "url": "https://substrate.dev/rustdocs/",
-      "selectors_key": "rustdocs"
+      "selectors_key": "rustdocs",
+      "page_rank": 50
     },
-    "https://substrate.dev/docs/",
-    "https://substrate.dev"
+    {
+      "url": "https://substrate.dev/docs/",
+      "page_rank": 100
+    },
+    {
+      "url": "https://substrate.dev",
+      "page_rank": 10
+    }
   ],
   "stop_urls": [],
   "selectors": {
--- a/configs/substrate.json
+++ b/configs/substrate.json
@@ -1,8 +1,5 @@
 {
   "index_name": "substrate",
-  "js_render": true,
-  "js_wait": 2,
-  "use_anchors": true,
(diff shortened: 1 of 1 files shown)


s-pace:
This is really heavy from a crawling point of view. Could you only rely on Server Side Rendering instead please?

shawntabrizi:
@s-pace I have removed the client side rendering parts of the config. I will look into replacing those aspects of our docs with SSR. That being said, I think removing the sitemap will still help us get the right docs found across our various platforms.

PR #923, opened 2019-07-10
Fix Wooclap's docsearch config

Mickael-van-der-Beek:
The software generating our FAQ (https://docs.wooclap.com/) recently changed the way they exposed the language code information. This means we now have to extract it from the DOM in a non-trivial way instead of the URL. 

I wrote an XPATH expression that should extract the language code successfully.

- Are language selector rules applies sequentially?
e.g: If a selector has results for rule n°1 and rule n°2, which result will be taken into account n°1 or n°2?

- Does the language code casing matter?
The expression returns an uppercased version of the string and the `lower-case()` function is only available in XPath v2 which is not widely available for browsers (not sure about DocSearch).

### DIFF
--- a/configs/wooclap.json
+++ b/configs/wooclap.json
@@ -33,6 +33,7 @@
     },
     {
       "url": "https://docs.wooclap.com/",
+      "selectors_key": "faq",
       "tags": [
         "faq"
       ]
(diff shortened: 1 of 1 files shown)


Mickael-van-der-Beek:
@s-pace Thank you for your quick reply. Sadly, we don't have control over the content of the DOM.

Which part of the solution I posted would not work / break in Docsearch?

s-pace:
Ah ok, let's try it then

PR #926, opened 2019-07-13
Update getbench.json

h-enk:
`start_urls` changed

# Pull request motivation(s)


h-enk:
Reverted back to old `start_urls`

PR #931, opened 2019-07-18
Add start urls and remove sitemaps from config

nagpalkaran95:
# Pull request motivation(s)

### What are the changes?

*Need to change the start urls for crawler and disable use of sitemaps*


s-pace:
Should be already fixed

PR #942, opened 2019-07-27
pytorch: ready to filter over categories.

s-pace:
The website should be built as follow:

```html
<lvl0/>
<lvl1/>
<lvl2/<
<lvl3/>
<any class='category'>cat1</any>
<p/> // cat1
<p/> // cat1
<p/> // cat1
...
<any class='category'>cat2</any>
<p/> // cat2
<p/> // cat2
<p/> // cat2
...
````
Every element pick up before the first tag cat1 will be with the default value, that is to say "all" thanks to these changes on your config. Element after the cat1 can be found by filtering over cat1 while the ones after cat2 will have the same attribute set to cat2.

On the front side, the snippet would need these changes;
```html

...
algoliaOptions: { 'facetFilters': ["category:$CATEGORY"] },
}); 
</script> 
```
- Replace $CATEGORY with the category you want to search on. The list of possible category depends of the position of the record according to the strategy explained above. So as of today you have: all

For example if you want to refine the search to the category "all" just specify:
```js
'facetFilters': ["category:all"] 
```

Cc @JoelMarcey

### DIFF
--- a/configs/pytorch.json
+++ b/configs/pytorch.json
@@ -49,7 +49,12 @@
       "lvl3": "article h3",
       "lvl4": "article h4",
       "lvl5": "article h5",
-      "text": "article p, article li"
+      "text": "article p, article li",
+      "category": {
+        "selector": ".category",
+        "global": true,
+        "default_value": "all"
+      }
     },
     "index": {
       "lvl0": {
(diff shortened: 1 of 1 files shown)


s-pace:
This PR will not fix the issue

PR #950, opened 2019-08-01
Update expo.json (add v34.0.0 docs, remove v29.0.0 and older)

ide:
Expo SDK 34 docs with URLs like https://docs.expo.io/versions/v34.0.0/ have been added, and docs for versions 29 and older have been removed. Also added the base domain as a start URL.

### DIFF
--- a/configs/expo.json
+++ b/configs/expo.json
@@ -1,18 +1,16 @@
 {
   "index_name": "expo",
   "start_urls": [
+    "https://docs.expo.io/",
     {
       "url": "https://docs.expo.io/versions/(?P<version>.*?)/",
       "variables": {
         "version": [
-          "v26.0.0",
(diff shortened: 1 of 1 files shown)


s-pace:
Actually this PR will break the search for 24h since the first start_url that matches apply. Fixed by latest commit.
(Sorry for the wrong commit. I am off and using my phone is not the best way)

ide:
My apologies -- I didn't test this. As a feature request, it might be nice to support patterns for variables so we could just write `v\d+\.0\.0` and match all future versions. This is probably a really specific request and I don't know how the scraper works internally so maybe it's not realistic :)

s-pace:
Thanks for the feedback. By experience I can tell that it is better to use DocSearch meta tags instead of relying on URLs :)

PR #958, opened 2019-08-06
Update config

YiniXu9506:
# Pull request motivation(s)
Update start url

Return items, whose URLs start with `https://pingcap.com/docs/dev`

@s-pace PTAL


s-pace:
Deployed. Is it fine?

YiniXu9506:
Yes, the Dev version has been crawled and returned. But the order does not match the page_rank I set. I opened a new issue, would you PTAL? Thanks.

PR #960, opened 2019-08-06
Update the page rank

YiniXu9506:
# Pull request motivation(s)
Update the page rank. 

The order of version is not we expected.

The order of return version should be v3.0, v2.1 and then dev.

@s-pace PTAL. Thanks


YiniXu9506:
@s-pace Conflicts removed! Thank you!

YiniXu9506:
@s-pace If you search query "tidb operator" or "google" or something else, you will see the version `dev` is higher than version `2.1`. However, I set higher value to `v2.1`. I don't know why this happen.

s-pace:
It is not deployed yet

s-pace:
The page_rank is only here to separate hits with same global ranking

YiniXu9506:
@s-pace I believe the global rankings of `dev` and `v2.1` are the same if I don't set any params to them. So I think `v2.1` will be higher than `dev` when I set the higher value to `v2.1`.

It seems most return results are ordered by 3.0, 2.1 and then dev.  But some of them are not, `dev` still higher than `v2.0`.

s-pace:
This ranking is only effective when the global one is a tie. This PR ia not deployed yet. Let’s see it tomorrow. I will show you the ranking for a specific query once it is deployed. Let me know the query you want to check

YiniXu9506:
Got it. Would you please help me to check "tidb operator". Please let me know once it deployed. Thank you so much!

s-pace:
We will

s-pace:
Deployed. Is it better now?

YiniXu9506:
I am sorry, I still get `dev` before `v2.1` from the results if I search query `tidb operator` or `google`.

s-pace:
Are you sure? I have tried with these queries on https://pingcap.com/en/ and v3 was first

s-pace:
Dev is before v2 but with hits that are not the same

s-pace:
Page_rank is only useful when there is a tie

YiniXu9506:
Okay I see what you mean! Thanks @s-pace

Sylvain Pace <notifications@github.com>于2019年8月7日 周三下午7:32写道：

Issue #744, opened 2019-02-25
[dgraph] Duplicate search results (pages with trailing slash)


danielmai:
# Do you want to request a **feature** or report a **bug**?

### If it is a DocSearch index issue, what is the related `index_name` ?

`index_name`=dgraph

## What is the **current** behaviour?

*If the current behaviour is a bug, please provide all the steps to reproduce and screenshots with context.*

When searching in https://docs.dgraph.io, many pages return duplicate results, e.g., searching for "dgraph" returns these two links:

- `https://docs.dgraph.io/get-started#dgraph`
- `https://docs.dgraph.io/get-started/#dgraph`

![2019-02-25-140006_3440x1440_scrot](https://user-images.githubusercontent.com/2251820/53371694-ce48ad80-3905-11e9-936b-1cdc5097904e.png)

## What is the **expected** behaviour?

Only a single result is returned (ideally the link without the trailing slash.

#### Any quick clues?
I read the ["duplicate results" FAQ](https://community.algolia.com/docsearch/faq.html#why-do-i-have-duplicate-content-in-my-results%3F) and it looks like it requires a change to the dgraph.json config in this repo.

s-pace:
👋 @danielmai 

You needed to PR a configuration in order to fix this issue. It has been solved [by this commit](https://github.com/algolia/docsearch-configs/commit/6f6c5c26f678f6d0f189d5eeb7652ff023e097de) 

Please note that adding a redirection status `3XX` on one of the URL would have also fixed this issue.

danielmai:
Thanks for fixing our config, @s-pace. Cheers!

Thanks for the tip about the 3XX behavior.

PR #961, opened 2019-08-07
Avoid crawling page with URL not finishing by a /

kenfdev:
Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

# Pull request motivation(s)

The Open Policy Agent official docs is planning to use docsearch for their search feature in the docs. Currently, the docsearch search result looks like it has duplicate content. This is probably the same issue as the following:

https://github.com/algolia/docsearch-configs/issues/744

The preview can be seen from the following URL

https://deploy-preview-1513--openpolicyagent.netlify.com/docs/v0.12.2/

When searching `regex`, you can see three results.

image

They are:

```
https://www.openpolicyagent.org/docs/v0.12.2/language-reference/#regex
https://www.openpolicyagent.org/docs/latest/language-reference#regex
https://www.openpolicyagent.org/docs/latest/language-reference/#regex
```

This PR should fix the results to be as follows:

```
https://www.openpolicyagent.org/docs/v0.12.2/language-reference/#regex
https://www.openpolicyagent.org/docs/latest/language-reference/#regex
```

The `latest` and `v0.12.2` duplication remains, but this should be related to `meta` tags as mentioned here:

https://github.com/open-policy-agent/opa/pull/1513#issuecomment-508811002

### DIFF
--- a/configs/openpolicyagent.json
+++ b/configs/openpolicyagent.json
@@ -3,7 +3,9 @@
   "start_urls": [
     "https://www.openpolicyagent.org/docs"
   ],
-  "stop_urls": [],
+  "stop_urls": [
+    ".*(?<!/)$"
+  ],
   "sitemap_urls": [
     "https://www.openpolicyagent.org/sitemap.xml"
   ],


kenfdev:
Thank you @s-pace for pointing the correct solution. Will close this PR.

PR #962, opened 2019-08-08
Update workers-cloudflare.json

signalnerve:
# Pull request motivation(s)

Hi all! We (the Cloudflare Workers team - hello!) are going to be adding a pretty expansive set of content to [our docs](https://workers.cloudflare.com/docs) that we'd like to exclude from search, under `/docs/archive`. This PR should (I hope! please let me know if I've done this wrong) exclude all of that content using the `stop_urls` key.

n/a - our content has not been merged in yet, so search is working as expected.

Any content under `/docs/archive` (e.g. `/docs/archive` itself, `/docs/archive/recipes`, etc) should be hidden from the Docsearch index.

n/a

thank you again for this service, so rad!

### DIFF
--- a/configs/workers-cloudflare.json
+++ b/configs/workers-cloudflare.json
@@ -3,7 +3,9 @@
   "start_urls": [
     "https://workers.cloudflare.com/docs"
   ],
-  "stop_urls": [],
+  "stop_urls": [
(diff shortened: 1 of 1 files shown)


signalnerve:
Thanks for the review @s-pace, updated :)

PR #963, opened 2019-08-08
cockroachlabs: do not scrape start urls

j-low:
# Pull request motivation(s)
* addresses: https://github.com/cockroachdb/docs/issues/5179

Docs landing page being returned in search results that include search term which happens to appear in link on landing page

@jseldess


j-low:
@s-pace prettifier changes reverted

PR #964, opened 2019-08-09
nus-cs2103-ay1920s1.json: tweak start_urls

damithc:
Provide a better starting URL for the tag 'schedule'


damithc:
The page added in this PR (`https://nus-cs2103-ay1920s1.github.io/website/schedule/timeline.html`) doesn't seem to be indexed yet. Is there a delay in crawling?

image

PR #965, opened 2019-08-09
Phalconphp - Added 3.4 (version) and specific languages

niden:
Added 3.4 (version) and specific languages

Please review to ensure that the config is correct :)


configs/phalconphp.json
```diff
@@ -1,7 +1,58 @@
...
+          "id-id",
+          "it-it",
+          "ja-jp",
+          "pl-pl",
+          "ru-ru",
+          "tr-tr",
+          "uk-ua",
+          "ur-in",
+          "zh-cn"
+        ]
```
s-pace: Please add an extra_attributes for the version here just like https://github.com/algolia/docsearch-configs/blob/f3b40e5435fa5f289cfbd758eb5f4e4286b3f6a6/configs/brew_all.json#L10

configs/phalconphp.json
```diff
@@ -33,4 +84,5 @@
     "761639486"
   ],
   "nb_hits": 499439
-}
```
s-pace: Avoid this change please

niden:
@s-pace I think I resolved the issues. Can you double check please?

configs/phalconphp.json
niden: Removed newline

configs/phalconphp.json
niden: Added the extra attributes

PR #967, opened 2019-08-09
[Refactor] Naming update

DarkShadeKnigh:
# Pull request motivation(s)

### DIFF
--- a/configs/bllFlow.json
+++ b/configs/bllFlow.json
@@ -1,14 +1,14 @@
 {
-  "index_name": "bllFlow",
+  "index_name": "bllflow",
   "start_urls": [
-    "https://big-life-lab.github.io/bllFlow/"
+    "https://big-life-lab.github.io/bllflow/"
   ],
   "sitemap_urls": [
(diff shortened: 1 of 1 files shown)


DarkShadeKnigh:
Had a name change this caused issues with the search bar not being updated

configs/bllFlow.json
```diff
@@ -1,14 +1,14 @@
 {
-  "index_name": "bllFlow",
+  "index_name": "bllflow",
```
s-pace: You can’t change it this way. We will have to provide you new credentials

DougManuel: We'd appreciate the new credentials. Do we make a request by email? Or do you send them to the original email (they were originally sent to me).

s-pace: Please apply and reply with a link to this thread. I will handle it asap when I am back from PTO in 2 weeks

s-pace:
@DarkShadeKnigh solved. New indexName: bllflow, new ApiKey:ebce673983a3972ea16324f81b978cdc.

PR #972, opened 2019-08-16
[Edit site config] Added tables to selectors.

kestarumper:
# Pull request motivation(s)
[uniforms](https://uniforms.tools/) is an open source tool capable of generating any given form for which we provide documentation that uses algolia as search bar.
Algolia don't include tables content in search results. We improve this by changing our site's algolia config.

Algolia only uses paragraphs, enumerations and headings.

Table's content will be available in search bar.

@radekmie @s-pace


s-pace:
Will be effective at the next crawl tomorrow

PR #974, opened 2019-08-20
Update pomerium.json

desimone:
# Pull request motivation(s)

Only /docs/ is indexed

/community and /recipes should also be indexeed


s-pace:
👋 @desimone,

DocSearch is only provided for documentation pages. What are the missing documentation pages that you would like to be indexed?

desimone:
https://www.pomerium.io/recipes/
https://www.pomerium.io/community/

Thanks!

PR #975, opened 2019-08-20
Update hackerone.json

jobertabma:
Our documentation site isn't using `name` or `id` attributes. To avoid appending the wrong anchor, we're disabling it altogether for the time being.

### DIFF
--- a/configs/hackerone.json
+++ b/configs/hackerone.json
@@ -33,5 +33,6 @@
   "conversation_id": [
     "535135855"
   ],
-  "nb_hits": 2588
+  "nb_hits": 2588,
+  "use_anchors": false
 }


configs/hackerone.json
```diff
@@ -33,5 +33,6 @@
   "conversation_id": [
     "535135855"
   ],
-  "nb_hits": 2588
+  "nb_hits": 2588,
+  "use_anchors": false
```
s-pace: This parameter is here to help at the crawling time. It defines if the anchor of the page scraped matters in the computation of the uniqueness of the page. It is useful for websites which require the anchor for the routing.

Extracting the `id` or `name` attribute is mandatory. If you want to avoid it, I would recommend you to remove the anchor from the hit returned by algolia on the front side. You can do it by using the [transformData method](https://community.algolia.com/docsearch/behavior.html#transformdata).

Please note that defining these attribute is a good practice.

s-pace:
Closing since it doesn't solve the issue raised. You can [find a bootstrap example here](https://jsfiddle.net/s_pace/oqL5xch6/)

s-pace:
Feel free to reopen if needed

jobertabma:
> Closing since it doesn't solve the issue raised. You can [find a bootstrap example here](https://jsfiddle.net/s_pace/oqL5xch6/)

Thanks, that'll work!

PR #976, opened 2019-08-21
Update kirupa.json

kirupa:
# Pull request motivation(s)


s-pace:
Will be deployed in 24h :)

PR #978, opened 2019-08-22
Make the handsontable docs parsable for version higher than 7.0.0

budnix:
# Pull request motivation(s)

I am currently implementing `docsearch` for https://handsontable.com/docs project and I spotted that with current config after new docs release we would have a problem with the non-existent indexed version? the new version would have to be added to "version" key (through PR)? Looking at existing configs it seems that `Jest` documentation works like ours. They don't have hardcoded docs versions so it is resolved automatically by crawler? Correct me if I'm wrong.

I expect that I don't need to change this configuration file every time the Handsontable version is released. So after deploying a new version of the documentation index will be fulfilled with the content.

And I was instructed to use `faceFilters`.
```
// ...
// ...
```

feature

### DIFF
--- a/configs/handsontable.json
+++ b/configs/handsontable.json
@@ -1,17 +1,37 @@
 {
   "index_name": "handsontable",
   "start_urls": [
-    {
-      "url": "https://handsontable.com/docs/(?P<version>.*?)/",
-      "variables": {
-        "version": [
-          "7.1.1",
-          "6.0.0"
-        ]
-      }
-    }
+    "https://handsontable.com/docs/"
   ],
-  "stop_urls": [],
+  "stop_urls": [
+    "/docs/1.14.0",
+    "/docs/1.14.1",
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @budnix 

This PR will not fixe you issue. Jest is using docusaurus which embedded the version as specific meta DocSearch tags. If you want to do the same, you can follow [this recommendation](https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags).

Feel free to reopen if needed.

budnix:
Thanks! I've already included the meta tags to the two latest versions of the docs. This from now on will be updated on every new release.

budnix:
Can we reopen the issue? I can't do it by my self (don't have a "Reopen PR" button). The Handsontable documentation (https://handsontable.com/docs) now contains the required HTML meta tags, so I believe the proposed config will work?

PR #980, opened 2019-08-23
handsontable: Filter over version

s-pace:
This PR solves #978
cc @budnix

Which version do you want us to index? Which one should not be indexed?

So far, the meta tags are only available on 7.1.1. Is it expected? Could you [provide us a sitemap as explained here](https://community.algolia.com/docsearch/tips.html#use-a-sitemapxml) please?

### DIFF
--- a/configs/handsontable.json
+++ b/configs/handsontable.json
@@ -1,15 +1,7 @@
 {
   "index_name": "handsontable",
   "start_urls": [
-    {
-      "url": "https://handsontable.com/docs/(?P<version>.*?)/",
-      "variables": {
-        "version": [
-          "7.1.1",
-          "6.0.0"
-        ]
-      }
-    }
+     "https://handsontable.com/docs/"
   ],
   "stop_urls": [],
   "sitemap_urls": [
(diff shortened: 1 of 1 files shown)


budnix:
We want to make https://handsontable.com/docs parsable only from version 7.1.0 and later. I've already added the HTML meta tag `docsearch:version` for https://handsontable.com/docs/7.1.1 and https://handsontable.com/docs/7.1.0. We won't change search algorithm for older docs versions so the crawler should ignore version listen in [here](https://github.com/algolia/docsearch-configs/pull/978/files#diff-5c2e4317b63a4b9611d5c7923e368a68R6).

The sitemap is available at https://handsontable.com/docs/sitemap.xml.

s-pace:
👋 @budnix 

Solved, you can now use the facetFilters according to the versions:

```html

  apiKey: '6b1f6a9141059860765e9ab6b9cabd0b',
  indexName: 'handsontable',
});
</script>
```

- Replace `$VERSION` with the version you want to search on. The list of possible version is extracted from your met tags.
   So as of today you have: `7.1.0`, `7.1.1`

- Replace `$LANGUAGE` with the language you want to search on. So as of today you have: `en`

budnix:
Thanks @s-pace 🥇 

I have one more question. What happens when we release a new version of documentation saying `8.0.0`? I understand that there will be a moment where the results are empty (before the crawler re-parses a new content)? Is there a recommended way to handle this issue?

s-pace:
You are welcome.

You will only need to wait 24h to be sure a crawl has run

PR #986, opened 2019-08-28
Add Studio tag and URL to twilio.json

negomi:
# Pull request motivation(s)
We have a use case where we need to search within just the Studio docs.

To do this, we need to apply a custom tag to the Studio section: https://community.algolia.com/docsearch/config-file.html#using-custom-tags

It is not possible to scope the search to a single section.

We can search within the Studio section only.

Feature

### DIFF
--- a/configs/twilio.json
+++ b/configs/twilio.json
@@ -1,7 +1,13 @@
 {
   "index_name": "twilio",
   "start_urls": [
-    "https://www.twilio.com/docs"
+    {
+      "url": "https://www.twilio.com/docs/studio",
+      "tags": ["studio"]
+    },
+    {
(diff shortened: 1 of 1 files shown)


configs/twilio.json
```diff
@@ -1,7 +1,13 @@
...
   "index_name": "twilio",
   "start_urls": [
-    "https://www.twilio.com/docs"
+    {
+      "url": "https://www.twilio.com/docs/studio",
+      "tags": ["studio"]
+    },
+    {
+      "url": "https://www.twilio.com/docs"
+    },
```
s-pace: Please remove this last coma to make it a valide JSON. You do not need to wrap this URL instead an object btw

PR #1000, opened 2019-09-04
docs.bitsrc.io moved to a new domain docs.bit.dev

itaymendel:
# Pull request motivation(s)

We've migrated from docs.bitsrc.io to docs.bit.dev, so i want to update the crawler.


s-pace:
#1000 🎉

PR #1011, opened 2019-09-10
Update hub index

Pat878:
# Pull request motivation(s)

Multiple hub index results are being returned for the same record.

Only 1 record should be returned for each hub result. Preferably, just the record's `lvl1` and `lvl2` info should appear in the results.

N/A

N/A


s-pace:
This PR will not address the issue. Continuing conversation by PM

PR #1020, opened 2019-09-17
Update sitemap url to sitemapindex.xml

aivus:
Use `sitemapindex.xml` for all versions instead of `sitemap.xml` only for current version.

Thanks :shipit:


s-pace:
You can try it out. It is live @aivus

PR #1025, opened 2019-09-18
Updated webflow config

tbell511:
updated webflow config

### DIFF
--- a/configs/webflow.json
+++ b/configs/webflow.json
@@ -16,7 +16,8 @@
       "attributes", 
       "links",
       "carrd",
-      "hide-carrd"
+      "hide-carrd",
+      "all"
     ]
   },
   "conversation_id": [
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @tbell511 

You are [misusing the `custom_settings` parameter](https://community.algolia.com/docsearch/config-file.html#custom_settings-optional). This key should only include Algolia settings. The items of the array `attributesForFaceting` are attributes from records build by the scraper. You can [find more details here](https://www.algolia.com/doc/api-reference/api-parameters/attributesForFaceting/).

Please provide more details regarding what you want to achieve if you want us to better assist you. You should follow [the default pull request template](https://github.com/algolia/docsearch-configs/blob/master/.github/PULL_REQUEST_TEMPLATE)

tbell511:
Hi @s-pace,

I was following this page when I added the `attributesForFaceting` https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags

Is there another way to do this?

PR #1034, opened 2019-09-23
add variable for facetfilters

YiniXu9506:
# Pull request motivation(s)
Add variables to start_urls for facetfilters

I want to use the version variable to filter the returned results

Request a **feature**

No
@s-pace PTAL, thanks!


configs/pingcap.json
```diff
@@ -4,50 +4,68 @@
     {
       "url": "https://pingcap.com/docs-cn/v3.0/",
       "selectors_key": "docs_cn",
+      "variables": {
```
s-pace: Variables are only used to dynamically inject values within the URL. Please use `extra_attribute` instead. You [can find an example here.](https://github.com/algolia/docsearch-configs/blob/master/configs/brew_all.json)

YiniXu9506: @s-pace Thanks!  Can I use the extra_attrobutes like below to filter results tag with cn and belong to v3.0?

```
index.search(
  {
    query: 'query string',
    facetFilters: ['tags: cn', 'version: v3.0'],
    hitsPerPage: 50,
  },
  (err, { hits } = {}) => {
    if (err) throw err;

    console.log(hits);
  }
);
```

s-pace: Yes indeed, please [mind the format of the facet filters](https://www.algolia.com/doc/api-reference/api-parameters/filters/#facet-filters). You should not include a space. 
```
    facetFilters: ['tags: cn', 'version:v3.0'],
```

YiniXu9506: Got it! Thanks! I have updated the extra_attributes, PTAL!

PR #1035, opened 2019-09-24
update kintohub.json

nandiheath:
# Pull request motivation(s)

Click on the search content is go to legacy pages and shows 404

Should go to the latest pages.

### DIFF
--- a/configs/kintohub.json
+++ b/configs/kintohub.json
@@ -1,8 +1,7 @@
 {
   "index_name": "kintohub",
   "start_urls": [
-    "https://docs.kintohub.com/docs/getting-started.html",
-    "https://docs.kintohub.com/docs/"
+    "https://docs.kintohub.com/docs/introduction/getting-started"
   ],
   "stop_urls": [],
   "selectors": {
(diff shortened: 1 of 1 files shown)


configs/kintohub.json
```diff
@@ -1,8 +1,7 @@
 {
   "index_name": "kintohub",
   "start_urls": [
-    "https://docs.kintohub.com/docs/getting-started.html",
-    "https://docs.kintohub.com/docs/"
+    "https://docs.kintohub.com/docs/introduction/getting-started"
```
s-pace: https://docs.kintohub.com/docs/introduction/getting-started will be the only page crawled and thus available from the search.

nandiheath: thank you for the advice. and i added the sitemap as well.

configs/kintohub.json
```diff
@@ -1,9 +1,9 @@
...
```
s-pace:
```suggestion
    "https://docs.kintohub.com/docs/"
```

`start_urls` will be used as a regex to match the sitemap in this context. The broader it is, the better

s-pace: image
I can't do it on my own.

nandiheath: ok, just fixed it

PR #1038, opened 2019-09-25
Changed fibaro_manuals config to support languages

pdymar:
# Pull request motivation(s)
Filtering results by language.

Search shows all language versions of pages.

Search shows only pages in current language


s-pace:
Just check and you do. Deploying it!

PR #1039, opened 2019-09-26
Change to production URL

potatodepaulo:
# Pull request motivation(s)
Changing to production URL for site docs.nearprotocol.com. I didn't understand that this would be a manual process on your part. I appreciate y'all setting this up :). If I understand correctly, this should update from crawler running in about 24 hours or so automatically.

Links in docs search take user to the staging url which I provided in application. 

 I'd like the results to stay within the docs portal domain.

### DIFF
--- a/configs/near.json
+++ b/configs/near.json
@@ -1,10 +1,10 @@
 {
   "index_name": "near",
   "start_urls": [
-    "https://docs-3bgv.onrender.com/"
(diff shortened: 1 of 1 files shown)


s-pace:
Happy to help @potatodepaulo 

Index updated! you can now retrieve https://docs.nearprotocol.com/

PR #1042, opened 2019-09-26
Prefect: Add a stop_url regex for preventing indexing archived API documentation

cicdw:
# Pull request motivation(s)
We archive our historical API documentation, which presents a large number of duplicate matches in our search.  This PR should prevent the indexing of our old versioned API docs.

Duplicate search results.

Not to crawl, e.g., https://docs.prefect.io/api/0.6.5/

Thanks!

### DIFF
--- a/configs/prefect.json
+++ b/configs/prefect.json
@@ -3,7 +3,7 @@
   "start_urls": [
     "http://docs.prefect.io"
   ],
-  "stop_urls": ["/api\/([0-9]+\.[0-9]+\.[0-9]+)\/.*"],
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @cicdw 

Deployed, I have also add `/unreleased/` to the `stop_urls` to avoid having duplicate from this version.

Feel free to PR if it doesn't suit your need.

cicdw:
Hi @s-pace - thank you! I believe including `"/unreleased/"` along will versioned APIs will prevent all pages in our API documentation from being indexed though - I'd just like the unreleased documentation indexed, as it is always the most up-to-date.  I'll PR a fix.

PR #1053, opened 2019-09-30
Fix start urls

stancl:
I removed the `getting-started` because it might change in the future. Hope Algolia is ok with a meta redirect to `getting-started/` on the new URLs.

Also made the first start URL specifically `v1` so that when I change the dynamic redirect, both versions will get indexed.


stancl:
Ah I see, the config has a wrong URL to the sitemap. Will fix in a second.

PR #1058, opened 2019-10-02
make bokeh config more specific

bryevdv:
# Pull request motivation(s)

This is a first PR to specialize the configuration for the particulars of the Bokeh documentation. 

Does not well-distinguish between User Guide, Ref Guide, and Dev Guides

Bokeh docs sections and facets are better defined. 

feature

Is there a way to factor out the repeated variables from the start_urls sections?

Is using a "#junk" target for lvl0 with a default value that should be used everywhere necessary/reasonable? I want to just specify that any pages that match the each selector key have a stipulated lvl0, always.

### DIFF
--- a/configs/bokeh.json
+++ b/configs/bokeh.json
@@ -1,35 +1,88 @@
 {
   "index_name": "bokeh",
   "start_urls": [
-    "https://docs.bokeh.org"
+    {
+      "url": "https://docs.bokeh.org/(?P<lang>.*?)/(?P<version>.*?)/user_guide/",
+      "selectors_key": "user_guide",
+      "page_range": 3,
+      "variables": {
+        "lang": ["en"],
+        "version": ["dev", "latest", "1.3.4", "1.2.0", "1.1.0", "1.0.4", "0.12.16", "0.12.15", "0.12.14", "0.11.0"]
+      }
+    },
+    {
+      "url": "https://docs.bokeh.org/(?P<lang>.*?)/(?P<version>.*?)/reference/",
+      "selectors_key": "reference",
+      "page_range": 2,
+      "variables": {
+        "lang": ["en"],
+        "version": ["dev", "latest", "1.3.4", "1.2.0", "1.1.0", "1.0.4", "0.12.16", "0.12.15", "0.12.14", "0.11.0"]
+      }
+    },
+    {
+      "url": "https://docs.bokeh.org/(?P<lang>.*?)/(?P<version>.*?)/dev_guide/",
+      "selectors_key": "dev_guide",
+      "page_range": 1,
+      "variables": {
+        "lang": ["en"],
+        "version": ["dev", "latest", "1.3.4", "1.2.0", "1.1.0", "1.0.4", "0.12.16", "0.12.15", "0.12.14", "0.11.0"]
+      }
+    }
   ],
   "stop_urls": [
     "/_",
     "/docs/gallery/"
   ],
   "selectors": {
-    "lvl0": {
-      "selector": ".current .toctree-l1:nth-child(2)",
-      "global": true,
-      "default_value": "Documentation"
+    "strip_chars": "¶",
+    "default": {
+      "lvl0": {
+        "selector": ".current .toctree-l1:nth-child(2)",
+        "global": true,
+        "default_value": "Documentation"
+      },
+      "lvl1": "title",
+      "lvl2": ".content h2, .class > dt .descname",
+      "lvl3": ".content h3, .method .descname",
+      "text": ".content p, .content li"
     },
-    "lvl1": ".content h1",
-    "lvl2": ".content h2, .class > dt .descname",
-    "lvl3": ".content h3, .method .descname",
-    "lvl4": ".content h4",
-    "lvl5": ".content h5",
-    "lvl6": ".content h6",
-    "text": ".content p, .content li",
-    "lang": {
-      "selector": "/html/@lang",
-      "type": "xpath",
-      "global": true
+    "user_guide": {
+      "lvl0": {
+        "selector": "#junk",
+        "global": true,
+        "default_value": "User Guide"
+      },
+      "lvl1": "h1",
+      "lvl2": ".content h2",
+      "lvl3": ".content h3",
+      "lvl3": ".content h4",
+      "text": ".content p, .content li"
     },
-    "version": {
-      "selector": ".version-link:first-child a",
-      "global": true
+    "reference": {
+      "lvl0": {
+        "selector": "#junk",
+        "global": true,
+        "default_value": "Reference Guide"
+      },
+      "lvl1": "h1",
+      "lvl2": ".content h2, .class > dt .descname, .function > dt .descname, .data > dt .descname",
+      "lvl3": ".content h3, .method .descname",
+      "text": ".content p, .content li"
+    },
+    "dev_guide": {
+      "lvl0": {
+        "selector": "#junk",
+        "global": true,
+        "default_value": "Developer's Guide"
+      },
+      "lvl1": "h1",
+      "lvl2": ".content h2",
+      "lvl3": ".content h3",
+      "lvl3": ".content h4",
+      "text": ".content p, .content li"
     }
   },
+  "selectors_exclude": [".toc", ".header", ".footer", ".second-nav"],
   "keep_tags": [
     "code"
   ],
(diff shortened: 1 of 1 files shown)


configs/bokeh.json
```diff
@@ -1,35 +1,88 @@
 {
   "index_name": "bokeh",
   "start_urls": [
-    "https://docs.bokeh.org"
+    {
+      "url": "https://docs.bokeh.org/(?P<lang>.*?)/(?P<version>.*?)/user_guide/",
+      "selectors_key": "user_guide",
+      "page_range": 3,
```
s-pace:
```suggestion
      "page_rank": 3,
```
Did you mean this parameter?

configs/bokeh.json
```diff
@@ -1,35 +1,88 @@
...
+      "variables": {
+        "lang": ["en"],
+        "version": ["dev", "latest", "1.3.4", "1.2.0", "1.1.0", "1.0.4", "0.12.16", "0.12.15", "0.12.14", "0.11.0"]
+      }
```
s-pace: Why not relying on the selectors as previously done?

s-pace: Because of this, you will need to update your configuration every time you release a new version

s-pace: If the issue is that the `version` selector does not extract a "latest" one, please use the DocSearch meta tags => https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags

configs/bokeh.json
```diff
@@ -1,35 +1,88 @@
...
+        "default_value": "Documentation"
+      },
+      "lvl1": "title",
+      "lvl2": ".content h2, .class > dt .descname",
+      "lvl3": ".content h3, .method .descname",
+      "text": ".content p, .content li"
+    },
+    "user_guide": {
+      "lvl0": {
+        "selector": "#junk",
```
s-pace: Good guess. Using an empty selector will be more explicit
```suggestion
        "selector": "",
```

configs/bokeh.json
bryevdv: @s-pace The initial email I got mentioned only that there was 1.3.4 indexed now, and at least the experiments I tried seemed to indicate there were no other facets other then 1.3.4 currently. I don't seem to have permission to access the actual index on the algolia account. Can you confirm if other versions have been picked up? And if not, what changes are necessary to have them be picked up? Thanks!

bryevdv: Also apologies but I don't see how to apply the meta-tag to get a "latest" version. In our current scheme the files (all on S3) under "latest" are just a copy of whatever files are under the actual latest version, e.g. "1.3.4" They won't differ (by a meta tag or anything else)

s-pace: @bryevdv Could [you provide a compliant sitemap](https://community.algolia.com/docsearch/tips.html#use-a-sitemapxml)? It will help to make sure we do crawl every version

s-pace: If you are fine with filtering with the numerical version e.g. `1.3.4`, that's fine then. We do extract this value from your DOM according to the selector `.version-link:first-child a`. Make sure the facetFilter is related to this value and not `latest` since it is not present within the DOM.

Side note, we will not redirect to the `1.3.4` URLs but `latest` instead due to the meta tags `rel=canonical`

bryevdv: @s-pace we currently generate `sitemap.txt` for each release, e.g 

http://docs.bokeh.org/en/1.3.4/sitemap.txt

I'm not familiar with an XML sitemap and also I am not sure how we could best keep a single "running" sitemap that has everything for all the releases offhand. We can certainly look into it but it's not something I can provide in the immediate future.

s-pace: Alright so please add these sitemaps and you will need to update your configuration at every release.

You can also define a sitemap index. [See here for reference](https://www.sitemaps.org/protocol.html#index).

s-pace:
@bryevdv I will do it.

s-pace:
Actually your sitemap is not compliant at all with our crawler. I didn't notice the `.txt`.

bryevdv:
@s-pace Thank you for all your help. I am definitely interested in interacting and  improving and especially in automating things as much as possible over time. We just have two very important releases to get out before the end of the year and adding back search is actually one of the longest-awaited features so I am willing to tolerate a little manual updating in the short term. 

> Actually your sitemap is not compliant at all with our crawler. I didn't notice the .txt.

I can see about writing a script to convert the existing .txt sitemaps to .xml ones and uploading by hand. In the mean time, even if it is just for 1.3.4 I am very keen to see if the selector keys and selectors for the different doc sections are working as well as possible. Is it possible to merge an update including just those?

s-pace:
On it. I think have a quick win

s-pace:
Thanks to 787603fdb386002c39369755226d6bc446203252 you now have the following versions available: 0.10.0, 0.12.14, 0.12.15, 0.12.16, 1.0.4, 1.1.0, 1.2.0, 1.3.4

bryevdv:
@s-pace awesome! Should I make a new PR with the selector key changes?

s-pace:
Please do, you should only update start_url and selectors due to 1bc152085a9388f0097c774f92c36c62fd47b457

bryevdv:
@s-pace how does specifying the separate selector_key blocks work with a list of string urls in `start_urls` work? Can I mix pattern blocks without variables, like 
```
    {
      "url": "https://docs.bokeh.org/(?P<lang>.*?)/(?P<version>.*?)/user_guide/",
      "selectors_key": "user_guide",
      "page_range": 3,
    },

if with the plain strings? Or do I need to multiply each versioned URL times all the pattern blocks to make blocks with explicit URLS?

**Comment by bryevdv** [comment]
@s-pace can you confirm that these are the correct sitemap and sitemap index formats that you expect? If so I can get all of them uploaded today

https://gist.github.com/bryevdv/6369e7862f9e30b6f9334af6904187f1

**Comment by bryevdv** [comment]
@s-pace sorry one more question :) I am trying to set up a more sophisticated search page with `instantsearch.js`. Without any extra, search now returns results for all versions. How can I set an initial version facet to use? I gather `facetFilters` is not available on `instantsearch`. I have seen maybe a half a dozen various answers to this question on SO and on algolia issues or support forums, but none of them worked. The most promising seemed perhaps to be:
```
    search.addWidget({
      init: function(args) {
        args.helper.addFacetRefinement('version', '1.2.0');
      }
    });
```
but that raises a JS error * version is not defined in the facets attribute of the helper configuration* and if I add the facet in `searchParameters` then it "works" but just returns no results at all.

s-pace:
> sitemap

I have tried to make the sitemap index available on a local server. It works but we would need the sitemaps to be available to make sure they are compliant. They were correctly parsed so please proceed and make them available.

If we have this sitemap, it will be much easier to create the dedicated `start_urls`. Let's way for the sitemap to be available to update your config.

You can also create an element having a class `lvl0-active` class that we will extract as `lvl0` instead of hardcoding it from the config. This class must be set from the server-side

s-pace:
> instantsearch.js

Please email docsearch@algolia.com for better assistance and better track of this request

bryevdv:
@s-pace OK I can upload them today, but I got this note via email:

> We would rather avoid crawling unused versions since we need to do a full crawl and indexing every time. If the crawl is too long, it will be evicted.

Old docs are always left up to be accessible to anyone who might need them. There will be an "archive" page that links all of them as well. How do I instruct docsearch what *not* to crawl in this case? Is it sufficient to leave them them out of the sitemap? Do they need to be explicitly called out in the config somehow? I guess that would need to be manually updated every release then.

bryevdv:
@s-pace I have uploaded all the sitemaps, e.g.

http://docs.bokeh.org/sitemap_index.xml
http://docs.bokeh.org/en/1.3.4/sitemap.xml

s-pace:
> Old docs are always left up to be accessible to anyone who might need them. There will be an "archive" page that links all of them as well. How do I instruct docsearch what not to crawl in this case? Is it sufficient to leave them them out of the sitemap? Or can I set the update frequency in the site map to "annual" or "never"? Do they need to be explicitly called out for exclusion in the config? I guess that would need to be manually updated every release then. Or maybe just excluding the archive page with the full set of links...

The crawl do not follow update frequency rules from sitemap.xml's and we do not provide a different frequency for the crawls except the daily basis. If you want to have a finer granularity for this, you will need to[ run the crawler on your own](https://community.algolia.com/docsearch/run-your-own.html). 

> @s-pace I have uploaded all the sitemaps, e.g.
Great, your DocSearch index is now synced with your documentation. The sitemap_index is correct and help to optimise the crawl. You can check it out.

I have also sent you an email. Let's continue by email so that we do not loose track of your help.

PR #1062, opened 2019-10-06
custom selectors for Bokeh sections

bryevdv:
# Pull request motivation(s)

Allow Bokeh docs search results to be grouped and prioritized according to major section of the docs (User Guide, Ref Guide, etc)

All results lumped into on unordered "Documentation" lvl0

Results grouped by major section

feature

### DIFF
--- a/configs/bokeh.json
+++ b/configs/bokeh.json
@@ -1,7 +1,39 @@
 {
   "index_name": "bokeh",
-  "start_urls": [
-    "https://docs.bokeh.org"
+    "start_urls": [
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/user_guide.html",
+      "selectors_key": "user_guide",
+      "page_rank": 3
+    },
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/user_guide/",
+      "selectors_key": "user_guide",
+      "page_rank": 3
+    },
+
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/reference.html",
+      "selectors_key": "reference",
+      "page_range": 2
+    },
+
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/reference/",
+      "selectors_key": "reference",
+      "page_range": 2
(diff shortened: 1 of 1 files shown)


configs/bokeh.json
```diff
@@ -1,7 +1,39 @@
...
-  "start_urls": [
-    "https://docs.bokeh.org"
+    "start_urls": [
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/user_guide.html",
+      "selectors_key": "user_guide",
+      "page_rank": 3
+    },
+    {
+      "url": "https://docs.bokeh.org/en/(?P<version>.*?)/docs/user_guide/",
```
s-pace: Having 2 times the same page might introduces duplicates

configs/bokeh.json
```diff
@@ -46,7 +139,6 @@
   "custom_settings": {
     "separatorsToIndex": "_",
     "attributesForFaceting": [
-      "lang",
```
s-pace: You will not be able to filter on `lang` if you remove it from the `attributesForFaceting`.

bryevdv:
@s-pace I did try to run the crawler myself and it blew through quotas on a Community account just testing out experiments even restricted to only a single version. Is there a way to run without uploading to algolia? Otherwise I am not sure what you are suggesting I should do.

configs/bokeh.json
bryevdv: I must've misunderstood an earlier comment, I thought the only facet was version and so that removing was appropriate.

configs/bokeh.json
bryevdv: It's unclear from the docs whether these are URLs to craw from or patterns to match against whenever happening to crawl any URL (or both)

bryevdv:
As an example, running just now on a new Community index, I see:

> DocSearch: https://docs.bokeh.org/en/1.3.4/docs/reference/plotting.html 4148 records)

This is surprising to me, but I have no idea why or how this one file has more than 10x the records of any other file. Any pointers to documentation (or again, especially any way to run in a "dry-run" mode that does not actually upload to algolia) would be very appreciated.

PR #1063, opened 2019-10-07
Added search for tables

dominikamachura:
# Pull request motivation(s)


s-pace:
For the reference: `table` is an element which is too big to be correctly indexed. Selectors must only target small paragraph of textual content.

PR #1065, opened 2019-10-07
Bryanv/simplify

bryevdv:
@s-pace I have simplified and tested as best I can under quota against CFWXQJJG8W and it performs to my expectations.

### DIFF
--- a/configs/bokeh.json
+++ b/configs/bokeh.json
@@ -1,37 +1,116 @@
 {
   "index_name": "bokeh",
-  "start_urls": [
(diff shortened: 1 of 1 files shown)


bryevdv:
Well, after following your advice to run the scraper manually, merely running four experimental scapes ***and doing nothing else*** was a enough to generate 64000 operations?? And now the account is suspended... How is anyone supposed to test their configuration if the mere act of generating and uploading an index counts against this budget so severely?

PR #1066, opened 2019-10-08
update config

YiniXu9506:
# Pull request motivation(s)
Build a custom UI for search results

We use dropdown UI to display the search results.

I want to create a UI page to display the search results. I remove the `page_rank` for each version update lvl0 selectors for `docs-cn` and `docs`.

feature

@s-pace PTAL, thanks!

### DIFF
--- a/configs/pingcap.json
+++ b/configs/pingcap.json
@@ -11,8 +11,7 @@
       },
       "tags": [
         "cn"
-      ],
-      "page_rank": 3
+      ]
     },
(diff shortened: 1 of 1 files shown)


s-pace:
You might want to implement DocSearch meta tags in such regards.
https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags

Merging it and deploying in the meantime

PR #1071, opened 2019-10-11
Update Cloudflare Workers docs URL

signalnerve:
# Pull request motivation(s)

Hi there! We've updated the location of our docs - this PR updates the URLs in our docsearch config.

Requests go to our old docs location, and are immediately redirected to the new location.

The URLs in docsearch are updated, skipping the need to go to the original URLs and redirect to the new location.

n/a

Quick question – we've had some issues with our docsearch where it's referencing outdated versions of pages when searching. An example of this is searching `cf`, where the third result points at text that is no longer on the associated page. I originally started working on this config again to try and resolve that, before realizing that the URLs were out-of-date. Will changing the URLs cause a re-index to remove the old content? Included a picture to show the behavior I'm talking about (notice that the content shown in search isn't on the current page, which is the associated search result page)

Screen Shot 2019-10-11 at 9 56 28 AM

### DIFF
--- a/configs/workers-cloudflare.json
+++ b/configs/workers-cloudflare.json
@@ -1,13 +1,13 @@
 {
   "index_name": "workers-cloudflare",
   "start_urls": [
-    "https://workers.cloudflare.com/docs"
+    "https://developers.cloudflare.com/workers"
   ],
   "stop_urls": [
-    "https://workers.cloudflare.com/docs/archive"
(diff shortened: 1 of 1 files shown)


s-pace:
@signalnerve the outdates result should be removed thanks to the new crawl I have run.

Can you confirm it solves the issue you have raised please?

PR #1080, opened 2019-10-23
Update DITA-OT config to index more text elements

infotexture:
# Pull request motivation(s)

**DITA Open Toolkit docs:** https://www.dita-ot.org/dev/

As a command-line tool, many of the parameter listings and other
options in our docs are rendered as definition lists, with terms in
`<dt>` elements, and descriptions in `<dd>` elements, but this content
is apparently not indexed at all.

For example, searching for the `args.figurelink.style` parameter yields
no results, though the term is present, and includes a dedicated ID:

https://www.dita-ot.org/dev/parameters/parameters-base.html#base__args.figurelink.style

At the moment, it looks like the index only “sees” paragraphs and list
items within the `<main>` element as text.

This pull requests updates the list of elements in the `text` key to
index  additional content on each page.

Signed-off-by: Roger Sheen <roger@infotexture.net>


infotexture:
🙏 Thanks @s-pace.

PR #1086, opened 2019-10-29
update guanmai config

BigBugaboo:
I need change this config, because data of search has some trouble, I had examine config in local.

### DIFF
--- a/configs/guanmai.json
+++ b/configs/guanmai.json
@@ -1,21 +1,17 @@
 {
   "index_name": "guanmai",
+  "js_render": true,
   "start_urls": [
     "https://document.guanmai.cn/"
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": {
-      "selector": "p.sidebar-heading.open",
-      "global": true,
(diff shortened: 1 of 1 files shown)


configs/guanmai.json
```diff
@@ -1,21 +1,17 @@
 {
   "index_name": "guanmai",
+  "js_render": true,
```
s-pace: Can you avoid to rely on client side rendering please?

BigBugaboo: sorry, can no.I have many document and so I use this way to develop for lesse time.if i use other way that  I will be weast more time.

BigBugaboo: Why is this not recommended this method?

BigBugaboo: if i do not add this,I can't get the exact path.

s-pace: Do you have a clear example please?
Did you run the crawl on your machine?

BigBugaboo:
> I need change this config, because data of search has some trouble, I had examine config in local.

PR #1096, opened 2019-11-06
Add faceting by language to bold_

CarlosBonetti:
Change configuration to consider the `<meta name="docsearch:language" ... />` tags


s-pace:
@CarlosBonetti Deployed, language facet is now available

PR #1109, opened 2019-11-14
Update clarity_design config

Mavericker-1996:
# Pull request motivation(s)

recover sitemap.xml field


Mavericker-1996:
Why does nb-hits didn't change when I recovered sitemap.xml? 
Does it update every 24h ?

s-pace:
[`nb_hits` is only updated when we run the crawl on purpose](https://community.algolia.com/docsearch/config-file.html#nb_hits-special). 

https://github.com/algolia/docsearch-configs/commit/55b64ed8e852f1ba89fdc436faa5ff2641eefe51 does it @Mavericker-1996. It grows.

Mavericker-1996:
thx, so next time when I update my website, will it change automatically or should I tell to run the crawl?

Mavericker-1996:
And I  found it missed  several search  results compared to the last update, will it change automatically ?

s-pace:
> `nb_hits` is the number of records that were extracted and indexed by DocSearch. We check this key internally to keep track of any unintended spike or drop that could reveal a misconfiguration.

> nb_hits is updated automatically each time you run DocSearch on your config. If the term is a tty, DocSearch will prompt you before updating the field. To avoid being prompted, set the UPDATE_NB_HITS environment variable to true (to enable) or false (to disable). This variable can be set in the .env file alongside APPLICATION_ID and API_KEY.

> You don't have to edit this field. We're documenting it here in case you were wondering what it's all about.

You do not have to update it yourself except if you have run the crawl on your own. [See here for more informations](https://community.algolia.com/docsearch/run-your-own.html)


> And I found it missed several search results compared to the last update, will it change automatically ?
 Could you [please fill out this template](https://github.com/algolia/docsearch-configs/blob/master/.github/ISSUE_TEMPLATE) in order to let us know which results are missing please?

Thanks

Mavericker-1996:
# Do you want to request a **feature** or report a **bug**?
**bug**

`index_name`= "alibaba-inc_clarity_design

The missed urls are:
- https://at-design.alibaba-inc.com/dev-doc/quick-start
- https://at-design.alibaba-inc.com/dev-doc/i18n

add these two missed urls

#### What have you tried to solve it?
update sitemap.xml

#### Any quick clues?



### Any other feedback / questions ?

PR #1117, opened 2019-11-18
Remove directory path

phatboyg:
So, when I enable the search, my results are linking to:

`http://masstransit-project.com/MassTransit/MassTransit/getting-started.html`

Which is a double path link. The generated content for VuePress is in /MassTransit, so I'm taking off the extension to hopefully? resolve the issue. If there is another setting to fix it, I haven't found it yet.

# Pull request motivation(s)

### DIFF
--- a/configs/masstransit.json
+++ b/configs/masstransit.json
@@ -1,7 +1,7 @@
 {
   "index_name": "masstransit",
   "start_urls": [
-    "http://masstransit-project.com/MassTransit/"
+    "http://masstransit-project.com/"
   ],
   "stop_urls": [],
   "selectors": {


s-pace:
Unfortunately this PR will not solve this issue which is due to vuepress itself cf https://github.com/vuejs/vuepress/issues/1802

We do crawl URLs without the `MassTransit/` repetition.  You can try to inject the whole JS without using the vuepress integration by following these steps:

```html

  apiKey: 'e458b7be70837c0e85b6b229c4e26664',
  indexName: 'masstransit',
  algoliaOptions: { 'facetFilters': ["lang:$LANG"] },
});
</script>
```

- Replace $LANG with the lang you want to search on. So as of today you have: en-US

  For example if you want to refine the search to the lang "en-US" just specify:
```js
'facetFilters': ["lang:en-US"]
```

phatboyg:
Right on, makes sense. I actually moved the entire site up a level, so now it's just at `http://masstransit-project.com/` without the path, so I need to update the index builder to scan from the top down, and then I can enable the search and all should work :)

s-pace:
👌 @phatboyg 

Your config is updated and the crawl already run. You should be all set. 🚀

PR #1125, opened 2019-11-21
Update kubesphere.json

leoendless:
Add a new domain for kubesphere


s-pace:
@leoendless We are receiving 403 on start_urls. Is the new domain available? If not, please PR your configuration only when your website is live please. Thanks

leoendless:
Sorry, it was my mistake. No need to add a new domain. But the versions need to be updated. See #1128

PR #1126, opened 2019-11-21
Update theodocs.json

MathiasRobert:
We just move to docusaurus v2 and need to change the config file to get our searchbar up to work again.


s-pace:
Deployed @MathiasRobert 

Thanks for the PR, works like a charm

PR #1131, opened 2019-11-21
Update itential_doc.json

michaelyfan:
# Pull request motivation(s)
Update config to dynamically pull documentation release versions instead of hardcoding it. Add these versions as attributes so our search can filter on them.

Config works currently, but will be out-of-date as soon as we launch a new release and the documentation is updated.

Config is able to adapt to new releases of documentation so that we don't have to update it every time there's a new release. Config is able to distinguish between release versions while scraping so that these can be attributes that our search dropdown can filter against.

N/A

N/A

### DIFF
--- a/configs/itential_doc.json
+++ b/configs/itential_doc.json
@@ -5,85 +5,84 @@
   ],
   "start_urls": [
     {
-      "url": "https://docs.itential.io/2019.2/admin/",
-      "page_rank": 3
+      "url": "https://docs.itential.io/(?P<version>.*?)/admin/",
+      "page_rank": 3,
+      "variables": {
+        "version": [
+          "2019.3",
+          "2019.2",
+          "2019.1",
+          "2018.3"
+        ]
+      }
     },
     {
-      "url": "https://docs.itential.io/2019.2/api/",
+      "url": "https://docs.itential.io/(?P<version>.*?)/api/",
       "selectors_key": "api",
-      "page_rank": 3
+      "page_rank": 3,
+      "variables": {
+        "version": [
+          "2019.3",
+          "2019.2",
+          "2019.1",
+          "2018.3"
+        ]
+      }
     },
     {
-      "url": "https://docs.itential.io/2019.2/changelogs/",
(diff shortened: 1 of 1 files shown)


michaelyfan:
We've been avoiding sitemaps because we still need to assign `selectors_key` based on the format of the URL. For example, URLs that match regular expression `\docs\.itential\.io\\|(?:.*\\)admin` should get `selectors_key: guide`. We've been using variables as a workaround since `start_urls` doesn't seem to like regexes:
```
{
      "url": "docs\.itential\.io\\|(?:.*\\)admin",
      "selectors_key": "guide",
      "page_rank": 3
}
```
gives me `ERROR: Error while obtaining start requests` when I test against the scraper.

Any way around this? If not, I'd rather this PR get merged; search is a pressing need for us right now.

michaelyfan:
I've updated the config with the selector change you mentioned, as well as split our use cases into several `start_url`s to illustrate our situation. I also went ahead an included a sitemap for your information.

It's not ideal since the releases are hardcoded into the configuration; we want this configuration to be able to adapt to any new release, hence my earlier regex, so that we don't have to keep coming back with an MR every couple of months for 1-line changes. This isn't resolved by a sitemap, since with one we'd still have to assign explicit release versions to `selector_keys`. What's the best course of action here? Are regular follow-ups inevitable?

s-pace:
Thank you for the details @michaelyfan 

I have made a PR #1134 to address the exposed issue. It should help to make sure the crawl extract the right version without the need to edit the configuration at each release 🚀 

The current sitemap you have provided is not compliant with our crawler. Could you make sure the `loc` tag contains the full URL of the page please?
[image](https://docs.itential.io/sitemap.xml)

[As stated by sitemap.org,](https://www.sitemaps.org/protocol.html)  the `loc` tag must be the `URL of the page. This URL must begin with the protocol (such as http) and end with a trailing slash, if your web server requires it. This value must be less than 2,048 characters.`

Please ping #1134 once you have updated the sitemap and that these changes are live.

PR #1133, opened 2019-11-25
Update and rename 3playground.json to three-playground.json

derekmei:
# Pull request motivation(s)

I believe there is a naming issue right now where the request to match the index name will not recognize 3playground because it starts with a number. I'd like to request this pull request be accepted so I can test and make sure that this is not the reason why my results are not showing up

**Screenshot of request - index doesn't match the name **
Screen Shot 2019-11-25 at 1 23 50 PM

Expected behavior is that the index in the request matches the index name in the configuration file.

Yes, would like to report a bug

### DIFF
--- a/configs/three-playground.json
+++ b/configs/three-playground.json
@@ -1,5 +1,5 @@
 {
-  "index_name": "3playground",
+  "index_name": "three-playground",
   "start_urls": [
     "https://www.3playground.com/"
   ],


s-pace:
Thanks for the complete details @derekmei 

You can use these credentials without any issue [as you can see on this jsfiddle](https://jsfiddle.net/s_pace/v20txsha/) 

Please note that the apiKey is meant to be shared publicly. It is a searchOnly one.

PR #1135, opened 2019-11-26
Update modyo locale configs

juak-modyo:
Updates the lang settings of modyo documentation indexes.

### DIFF
--- a/configs/modyo.json
+++ b/configs/modyo.json
@@ -1,40 +1,48 @@
 {
-  "index_name": "modyo",
-  "start_urls": [
-    "https://docs.modyo.com/"
-  ],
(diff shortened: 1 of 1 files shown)


juak-modyo:
> As a vuepress, the lang are extracted from the DOM according to the lang selector. You do not need to bother.
> 
> This PR is not helping as it wrongly uses variable. Please [comply with the PR template ](https://github.com/algolia/docsearch-configs/blob/master/.github/PULL_REQUEST_TEMPLATE)and [refer to the dedicated documentation](https://community.algolia.com/docsearch/config-file.html)

Thanks for the reply. I'll close it as it is not needed.

PR #1136, opened 2019-11-27
Update and rename salestimhelpcenter.json to salestimdeveloperhub.json

guillaumemeyer:
moving our site from help.Salestim.com to developers.salestim.com


s-pace:
You can not rename your indexName with a PR. [Please apply with the new website](https://community.algolia.com/docsearch/apply.html). We will remove the old index once we will provide your new credentials. Thanks

PR #1143, opened 2019-11-27
Coderello Docs Update

hivokas:
# Pull request motivation(s)

- change the `index_name` and config file name because I need to index docs for all packages, not only `laravel-nova-lang`
- remove URLs from `start_urls` because I think that it is better to use `sitemap.xml` to get the list of pages
- update selectors (use `docsearch-content` class as the root one for the content) 
- add `attributesForFaceting` in order to have an ability to filter by language, version, package


hivokas:
Unfortunately, checks have failed because of the change of the filename. 
Isn't it allowed?

configs/coderello_docs.json
```diff
@@ -0,0 +1,26 @@
+{
+  "index_name": "coderello_docs",
+  "start_urls": [],
```
s-pace: Please keep start_urls so that we have a better idea of what are the documentation pages you want to index

configs/coderello_docs.json
```diff
@@ -0,0 +1,26 @@
...
+    "lvl0": ".docsearch-content h1",
+    "lvl1": ".docsearch-content h2",
+    "lvl2": ".docsearch-content h3",
+    "lvl3": ".docsearch-content h4",
+    "lvl4": ".docsearch-content h5",
+    "lvl5": ".docsearch-content h6",
+    "text": ".docsearch-content p, .docsearch-content li"
+  },
+  "custom_settings": {
+    "attributesForFaceting": ["language", "version", "package", "page"]
```
s-pace: Did you make sure to have 4 DocSearch meta tags in every page?

configs/coderello_docs.json
```diff
@@ -0,0 +1,26 @@
+{
+  "index_name": "coderello_docs",
```
s-pace: Changing the name of the configuration will require additional steps on our side. The reason why you want to change the name of it are legit, we will update it once deploying this PR. You can disregard the circle ci check.

configs/coderello_docs.json
hivokas: I have these tags on every documentation page, yeah.

hivokas: For example:

```
<meta name="docsearch:version" content="1.0">
<meta name="docsearch:package" content="laravel-nova-lang">
<meta name="docsearch:page" content="introduction">
```

For this page: https://coderello.com/docs/laravel-nova-lang/1.0/introduction

hivokas: However, the main page of the website doesn't have ones, because it should not be indexed.
So it doesn't have either the root class `.docsearch-content` and `docsearch:*` meta tags.

hivokas: I'm also planning to add a blog to this website. But it should not be indexed because blog ≠ docs 🙂
How could we handle it? Will the excess of the root class and meta tags command the parser to skip these pages?

configs/coderello_docs.json
hivokas: Thanks.

configs/coderello_docs.json
hivokas: Could it be the link to the main page of the website?

hivokas: I thought that if I the config file doesn't have `start_urls` then the parser will use the `sitemap.xml`. Did I get it wrong?

configs/coderello_docs.json
hivokas: *not excess, but absense

hivokas:
@s-pace waiting for your feedback 🙂

configs/coderello_docs.json
s-pace: We do not usually index blog section.  However we might still encompass this part of your documentation if it does help the community with enlightening content. If so, please create a dedicated start_url item with a dedicated `selector_key`.

Please note that meta tags are retrieved on every page despite of the selector_key

configs/coderello_docs.json
s-pace: > Could it be the link to the main page of the website?

Yes that's correct, however you will need to remove the whole .start_url key. Anyway, please use the main page of the website. 

You can use   `"scrap_start_urls": false`, to avoid indexing the first page

s-pace:
Please address the last commit. I will merge later on this day so that you can start integrating it 🚀

hivokas:
I've updated the PR.

s-pace:
You should be all set @hivokas. Same apiKey, you only need to change the indexName as following:

​apiKey: 'b82c2509a4f7c906241f15a97c28a173',
indexName: 'coderello_docs',

hivokas:
Thanks!

Could I somehow access an `Overview` tab?

Now I see this error:
```
Access restricted

Please ask the owner of this application (accounts+documentation-search@algolia.com) to grant you necessary permissions.
```

s-pace:
No, you can not. You can only have this access by [running DocSearch on your own](https://community.algolia.com/docsearch/run-your-own.html).

hivokas:
Okay, thanks a lot for the assistance with the docsearch setup 🙂

PR #1151, opened 2019-12-05
patternfly

jschuler:
Update selectors
For level 0, Instead of grabbing "Documentation" from the top tab, it should grab the framework (React/HTML) and display that instead. 
Also remove the previous level 2 selector as it was just duplicated in the main body header


jschuler:
@s-pace do you know why the test on CI is failing?

s-pace:
Issue on our end sorry. All ✅

PR #1167, opened 2019-12-08
Update wpcontentcrawler.json

turgutsaricam:
# Pull request motivation(s)
The search results are displayed from all versions of the documentation regardless of the currently viewed version, causing same contents from different versions appear in the results. I wanted to display the results just for the currently viewed version of the documentation. After this is merged, I will add facet filters using the JavaScript snippet I added to the site, to show results just for the current version.

Current behavior is not a bug.

Showing the results just for the current version of the documentation.

I am not sure if I updated the file correctly. In this case, `(?P<version>.*?)` will only match `master` and `1.9` and disregard other matches, right? Also, `/master` and `/` actually points to the same version. Could you please tell me if it will be enough to define the same facet filter for both of these URLs, e.g `'facetFilters': ["version:master"]`?

### DIFF
--- a/configs/wpcontentcrawler.json
+++ b/configs/wpcontentcrawler.json
@@ -1,6 +1,12 @@
 {
   "index_name": "wpcontentcrawler",
   "start_urls": [
+    {
+      "url": "https://docs.wpcontentcrawler.com/(?P<version>.*?)/",
+      "variables": {
+        "version": ["master", "1.9"]
+      }
+    },
     "https://docs.wpcontentcrawler.com/"
   ],
   "stop_urls": [


s-pace:
Thanks for the PR.

> I am not sure if I updated the file correctly. In this case, (?P<version>.*?) will only match master and 1.9 and disregard other matches, right? Also, /master and / actually points to the same version. Could you please tell me if it will be enough to define the same facet filter for both of these URLs, e.g 'facetFilters': ["version:master"]?

Yes, the scraper will only index the versions defined from `variables`.

`/` will not be indexed since the pages are not available under pages defined from `start_urls`.

PR #1186, opened 2019-12-14
CodeceptJS: Updated search index

DavertMik:
We have updated our site, indexes should be updated too

# Pull request motivation(s)

### DIFF
--- a/configs/codecept.json
+++ b/configs/codecept.json
@@ -12,8 +12,8 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "//*[contains(@class,'navGroups')]//*[contains(@class,'navListItemActive')]/preceding::h3[1]",
-      "type": "xpath",
+      "selector": ".article-title",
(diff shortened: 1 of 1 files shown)


configs/codecept.json
```diff
@@ -12,8 +12,8 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "//*[contains(@class,'navGroups')]//*[contains(@class,'navListItemActive')]/preceding::h3[1]",
-      "type": "xpath",
+      "selector": ".article-title",
+      "type": "css",
```
s-pace:
```suggestion
```

DavertMik:
Done!

DavertMik:
How is the index updated? Does it require some time after this PR to rerun a crawler, or not?
I didn't test it on algolia locally, but all selectors are valid

s-pace:
Deployed and fixed

PR #1198, opened 2019-12-16
Update realdeposits.json

xadjek:
# Pull request motivation(s)

### DIFF
--- a/configs/realdeposits.json
+++ b/configs/realdeposits.json
@@ -1,7 +1,14 @@
 {
   "index_name": "realdeposits",
   "start_urls": [
-    "https://doc.realdeposits.com/integration_docs/latest"
+    {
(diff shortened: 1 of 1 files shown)


s-pace:
We do not want to add unnecessary version. Please remove the filter to fix the search experience.

The recommended way to handle version is available [from this part of our documentation](https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags).

xadjek:
I dont have any filters on my page. I have only one version of doc - https://doc.realdeposits.com/integration_docs/latest, but I cant  remove "latest" word in url because of doc engine. Please help here

s-pace:
@xadjek you do filter on version on your page as you can see on the following screenshot made on the network requests made by your website:
image

Please follow the instructions previously provided by email and copy the JS snippet which does not filter on non-existing filter.

Thanks

PR #1199, opened 2019-12-16
Update workato.json

th0r:
We've shipped a new version of Workato docs based on VuePress (https://docs.workato.com) so we need to update the config.

New settings are based on [this config](https://github.com/algolia/docsearch-configs/blob/master/configs/workato_vuepress.json) which was used on the test website (https://th0r.github.io/workato-docs/).


configs/workato.json
```diff
@@ -4,17 +4,37 @@
...
+  ],
+  "strip_chars": " .,;:#",
+  "custom_settings": {
+    "attributesForFaceting": [
+      "lang"
+    ]
   },
   "conversation_id": [
-    "952339719"
+    "1019001521"
```
s-pace:
```suggestion
    "952339719"
```
Please do not edit this part of the configuration.

th0r: Fixed

th0r:
> In order to avoid wasting ressources, can we remove the `workato_vuepress` config and the underlying crawl?

Removed `workato_vuepress.json`

PR #1201, opened 2019-12-17
Updates for silverstripe.json

unclecheese:
* Add new URL
* Add new facet

### DIFF
--- a/configs/silverstripe.json
+++ b/configs/silverstripe.json
@@ -1,7 +1,8 @@
 {
   "index_name": "silverstripe",
   "start_urls": [
-    "https://docs.silverstripe.org"
+    "https://docs.silverstripe.org",
+    "https://ss-userhelp.netlify.com"
   ],
   "stop_urls": [],
   "selectors": {
@@ -19,7 +20,8 @@
   "nb_hits": 21730,
   "custom_settings": {
     "attributesForFaceting": [
-      "version"
+      "version",
+      "context"
     ]
   }
 }


configs/silverstripe.json
```diff
@@ -1,7 +1,8 @@
 {
   "index_name": "silverstripe",
   "start_urls": [
-    "https://docs.silverstripe.org"
+    "https://docs.silverstripe.org",
+    "https://ss-userhelp.netlify.com"
```
s-pace: We would like to stick to one domain only per config. What is the difference between both domain?
You will need to apply for the new URL

configs/silverstripe.json
```diff
@@ -19,7 +20,8 @@
   "nb_hits": 21730,
   "custom_settings": {
     "attributesForFaceting": [
-      "version"
+      "version",
+      "context"
```
s-pace: We could not find the following meta in all pages of your website.
```html
<meta data-react-helmet="true" name="docsearch:context" content="TBD">
```

Could you make sure that this meta is available on every page please?

configs/silverstripe.json
unclecheese: We have two documentation sites that are running of the same app. The content it displays is determined by an environment variable. If I have to apply for a new domain, that's fine, I guess. It would remove the need for the `context` facet.

configs/silverstripe.json
unclecheese: Sorry, where did you see this `TBD` tag? Could you provide a link to a page that does not have it? Should be a value of `user` or `docs` depending on the domain.

Again, probably irrelevant if we're doing a new application.

s-pace: TBD here stands for To Be Defined. Since I couldn't find the meta I have created this value. Sorry if it was not clear!

This is indeed irrelevant and it does not work like this. Please refer [to the dedicated documentation](https://community.algolia.com/docsearch/config-file.html) before submitting a PR. Let us know if some part of the documentation are not clear, we will be happy to correct it.

s-pace:
Closing this PR since adding a brand new domain will require a new application. We have already acknowledged the application and we will provide you the credentials later on today in the dedicated email thread.

PR #1585, opened 2020-01-06
Fix url of result pages

znat:
Currently, a `/docs` element is added to the result page. If the page url is `https://botfront.io/docs/rasa/nlu/training-adding-data/#add-training-data`, clicking on a result will link to `https://botfront.io/docs/docs/rasa/nlu/training-adding-data/#add-training-data`.
I suppose it's because the start_url appends a `/docs` to the hostname

Also, I removed the `docs.botfront.io` site which does not exist anymore

# Pull request motivation(s)

### DIFF
--- a/configs/botfront.json
+++ b/configs/botfront.json
@@ -1,8 +1,7 @@
 {
   "index_name": "botfront",
   "start_urls": [
-    "https://docs.botfront.io",
-    "https://botfront.io/docs"
+    "https://botfront.io"
   ],
   "stop_urls": [],
   "selectors": {


znat:
Hi @s-pace,
Seems the changes are not in effect:

The link seems correct when hovering:

Capture d’écran 2020-01-20 à 09 19 46

But it gets to the wrong page (it adds a `docs` to the path):

Capture d’écran 2020-01-20 à 09 20 00

Any idea on how to fix that?
Thanks

s-pace:
Links are correctly from the index but I think the issue is due to the JS snippet.

znat:
Thanks @s-pace 
> but I think the issue is due to the JS snippet.
Can you elaborate a bit?

PR #1601, opened 2020-01-16
Moleculer: fix version selection in the config

icebob:
# Pull request motivation(s)
The Moleculer 0.14 version documentation is not available with Algoliasearch. This PR fixes it. Using the ˙<option>` values instead of parsing the version from the option text.

### DIFF
--- a/configs/moleculer.json
+++ b/configs/moleculer.json
@@ -6,7 +6,7 @@
       "variables": {
         "version": {
           "url": "https://moleculer.services/docs/",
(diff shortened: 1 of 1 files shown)


s-pace:
Failling PR, can you double check pls

icebob:
Should I need something else that Algolia search works with 0.14 version as well?

image

s-pace:
Yes this PR does not work. It breaks the configuration. [Could you use meta tags instead please?](https://community.algolia.com/docsearch/required-configuration.html#introduces-global-information-as-meta-tags)

icebob:
Ok, thanks. I've added meta tags. What's the next step? Or it will work after some time?

image

icebob:
@s-pace It doesn't still work.

PR #1605, opened 2020-01-22
Set page rank for more important pages

jackieros:
Our less important pages seem to show up more often or higher than the more important ones (perhaps because the more important ones are longer & have more sections in them? ) Setting rank for our more important pages.

### DIFF
--- a/configs/cloudinary.json
+++ b/configs/cloudinary.json
@@ -1,7 +1,83 @@
 {
   "index_name": "cloudinary",
   "start_urls": [
-    "https://cloudinary.com/documentation"
+    "https://cloudinary.com/documentation",
+    {
+      "url": "https://cloudinary.com/documentation/image_transformations",
+      "page_rank": 3
+    },
+    {
+      "url": "https://cloudinary.com/documentation/video_manipulation_and_delivery",
+      "page_rank": 3
+    },
+    {
+      "url": "https://cloudinary.com/documentation/upload_images",
+      "page_rank": 3
+    },
+    {
+      "url": "https://cloudinary.com/documentation/upload_videos",
+      "page_rank": 3
+    },    
+    {
+      "url": "https://cloudinary.com/documentation/image_upload_api_reference",
+      "page_rank": 3
+    },
+    {
+      "url": "https://cloudinary.com/documentation/image_transformation_reference",
+      "page_rank": 3
+    },
+    {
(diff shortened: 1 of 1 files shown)


configs/cloudinary.json
```diff
@@ -1,7 +1,83 @@
 {
   "index_name": "cloudinary",
   "start_urls": [
-    "https://cloudinary.com/documentation"
+    "https://cloudinary.com/documentation",
```
s-pace: Should be the last item of the array since the first which matches apply

jackieros:
@s-pace - Is this the right way to do it? 
A few other questions that I didn't see addressed in the docs: 

1. Within pages that are assigned the same rank, does the listed order in the array matter from a priority-perspective, or will all '3's be equal? 

2. Can you confirm that this page rank is just part of what determines the overall ranking in the results?  For example, if a lower-ranked page has a search term hit within an H2 at the top of the page, can it still win out over a higher ranked page in which the search term is only included once within regular paragraph text near the bottom of that page? 

3. Are the rank numbers only evaluated by order, or is the number actually treated as a weight - i.e. If only 2 pages are defined with page_rank values, is there any difference between setting the ranks as 1 & 2 or 1 & 8 ? 

Thx.

s-pace:
> Within pages that are assigned the same rank, does the listed order in the array matter from a priority-perspective, or will all '3's be equal?

Position matter yes. Level two. The higher it is, the higher its rank will be.

> Can you confirm that this page rank is just part of what determines the overall ranking in the results? For example, if a lower-ranked page has a search term hit within an H2 at the top of the page, can it still win out over a higher ranked page in which the search term is only included once within regular paragraph text near the bottom of that page?

Yes it can. It [depends of these criteria](https://www.algolia.com/doc/guides/getting-started/how-algolia-works/#relevance).

> Are the rank numbers only evaluated by order, or is the number actually treated as a weight - i.e. If only 2 pages are defined with page_rank values, is there any difference between setting the ranks as 1 & 2 or 1 & 8 ?

The value has an impact yes.

> I assume the page_rank only applies to the specific page listed and the ranking does not filter down to all pages linked from that page (or "below" that page in terms of the level configs)?

All pages matching the URL will be impacted.

jackieros:
@s-pace - Thanks Sylvain.
If you are OK with the PR, please go ahead and merge.

PR #1612, opened 2020-01-30
Update qctrl.json

robertlove:
Updating config as HTML structure has changed


robertlove:
> We need to rely on a main container of your textual content [as you can see here](https://community.algolia.com/docsearch/required-configuration.html#use-the-right-classes-as-selectors)

Thanks. Will fix asap

PR #1616, opened 2020-01-31
Update ClickHouse docs crawler config

blinkov:
# Pull request motivation

1. Documentation moved to new domain
2. Non-default version docs are hidden from normal search engines, but also need some in-site search

### DIFF
--- a/configs/yandex_clickhouse.json
+++ b/configs/yandex_clickhouse.json
@@ -43,6 +43,5 @@
   "conversation_id": [
     "1058861561"
   ],
-  "nb_hits": 50105,
-  "ignoreRobotsTxtRules": true
+  "nb_hits": 50105
 }
--- a/configs/yandex_clickhouse.json
+++ b/configs/yandex_clickhouse.json
@@ -1,7 +1,7 @@
 {
   "index_name": "yandex_clickhouse",
   "start_urls": [
-    "https://clickhouse.yandex/docs/"
+    "https://clickhouse.tech/docs/"
   ],
   "stop_urls": [],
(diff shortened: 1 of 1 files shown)


configs/yandex_clickhouse.json
```diff
@@ -20,16 +20,29 @@
...
   },
   "min_indexed_level": 2,
   "conversation_id": [
     "1058861561"
   ],
-  "nb_hits": 50105
-}
\ No newline at end of file
+  "nb_hits": 50105,
+  "ignoreRobotsTxtRules": true
```
s-pace: What is the use of this parameter? It is not supported from our scraper IIRC

blinkov: @s-pace I've found it over here: https://www.algolia.com/doc/api-reference/crawler/configuration/ignore-robots-txt-rules/

s-pace: Thanks for the details. The URL you are referring to is the enterprise crawler offered by Algolia. This is another tool completely different from the DocSearch Crawler. [You can find out more information on this page](https://www.algolia.com/products/crawler/).

The documentation related to the DocSearch crawler is available [from this dedicated documentation](https://community.algolia.com/docsearch/config-file.html). The DocSearch crawler does not rely on the robots.txt so no worries.

s-pace: Please remove this attribute from the configuration so that I can deploy.

blinkov: @s-pace done, thanks

PR #1621, opened 2020-02-04
keep: create webiny-js

Shipow:
HUB generated configuration

### DIFF
--- a/configs/webiny-js.json
+++ b/configs/webiny-js.json
@@ -1,0 +1,36 @@
+{
+  "index_name": "webiny-js",
+  "start_urls": [
+    "https://docs.webiny.com/"
+  ],
+  "stop_urls": [],
+  "selectors": {
+    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li",
+    "lvl1": "[class^='docItemContainer_'] h1",
+    "lvl2": "[class^='docItemContainer_'] h2",
+    "lvl3": "[class^='docItemContainer_'] h3",
+    "lvl4": "[class^='docItemContainer_'] h4",
+    "lvl5": "[class^='docItemContainer_'] h5",
+    "lvl6": "[class^='docItemContainer_'] h6",
+    "lvl0": {
+      "selector": ".menu__link--sublist.menu__link--active",
+      "global": true,
+      "default_value": "Documentation"
+    }
+  },
+  "selectors_exclude": [
+    ".hash-link"
+  ],
+  "sitemap_urls": [
+    "https://docs.webiny.com/sitemap.xml"
+  ],
+  "sitemap_alternate_links": true,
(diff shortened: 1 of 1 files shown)


configs/webiny-js.json
```diff
@@ -0,0 +1,36 @@
...
+    "https://docs.webiny.com/sitemap.xml"
+  ],
+  "sitemap_alternate_links": true,
+  "custom_settings": {
+    "attributesForFaceting": [
+      "language",
+      "version"
+    ]
+  },
+  "min_indexed_level": 0,
```
s-pace: Can we make sure to not define this value to 0 since it is its default value?

configs/webiny-js.json
```diff
@@ -0,0 +1,36 @@
...
+  "stop_urls": [],
+  "selectors": {
+    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li",
+    "lvl1": "[class^='docItemContainer_'] h1",
+    "lvl2": "[class^='docItemContainer_'] h2",
+    "lvl3": "[class^='docItemContainer_'] h3",
+    "lvl4": "[class^='docItemContainer_'] h4",
+    "lvl5": "[class^='docItemContainer_'] h5",
+    "lvl6": "[class^='docItemContainer_'] h6",
+    "lvl0": {
```
s-pace: Can we make sure to order correctly the lvl in an increasing order please?

configs/webiny-js.json
```diff
@@ -0,0 +1,36 @@
...
+      "selector": ".menu__link--sublist.menu__link--active",
+      "global": true,
+      "default_value": "Documentation"
+    }
+  },
+  "selectors_exclude": [
+    ".hash-link"
+  ],
+  "sitemap_urls": [
```
s-pace: Should be defined after `start_urls `

configs/webiny-js.json
```diff
@@ -0,0 +1,36 @@
...
+    "https://docs.webiny.com/sitemap.xml"
+  ],
+  "sitemap_alternate_links": true,
```
s-pace: Should be defined after `sitemap_urls`

Shipow: before ?

configs/webiny-js.json
Shipow: It's your own code, I didn't change the class function. Please PR it as this should be easier.

Shipow: should we stop generating the demo meanwhile?

s-pace: Will do, let's create some this way, I will fix them up later on then.

configs/webiny-js.json
s-pace: No afterward

PR #1633, opened 2020-02-05
Move/refine index to reflect production changes

hkfoster:
We’ve launched a new version of our main product, and as a result have moved our documentation to a subdomain. The requested changes are simply account for that move.

### DIFF
--- a/configs/fly.json
+++ b/configs/fly.json
@@ -1,20 +1,20 @@
 {
-  "index_name": "fly",
+  "index_name": "js-fly",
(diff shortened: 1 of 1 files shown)


s-pace:
Thanks for the PR @hkfoster 

This is not possible to rename a configuration. I have updated fly.json instead since it is already used on your website. If you want a new `index_name`, please [apply again from this URL](https://community.algolia.com/docsearch/apply.html)

PR #1739, opened 2020-02-25
Update flo-scss.json

matttelliott-flosports:
update configuration for flo-scss


configs/flo-scss.json
```diff
@@ -5,13 +5,13 @@
...
   "stop_urls": [],
   "selectors": {
-    "text": ".p-5 p, .p-5 li",
-    "lvl0": ".p-5 h1",
-    "lvl1": ".p-5 h2",
-    "lvl2": ".p-5 h3",
-    "lvl3": ".p-5 h4",
-    "lvl4": ".p-5 h5",
-    "lvl5": ".p-5 h6"
+    "text": ".flo-scss-content p, .p-5 li",
```
Shipow: ^ the parent of the list element needs to be updated as well

matttelliott-flosports: fixed

PR #1743, opened 2020-02-26
Update monolithic config

VibroAxe:
# Pull request motivation(s)
Documentation not indexing correctly. Fixed based on jekylrb (same site theme and implementation

No index result's

Index result's ;)

#NB
Config Update


VibroAxe:
Updated min-indexed-level to ensure we have a real title. Ready to merge

PR #1752, opened 2020-02-28
Document code tags

paulcsmith:
### Pull request motivation(s)

There are a number of people searching for specific calls, but they don't show up since it is in a code tag. We'd love to have those indexed as well

Code tags are not indexed

Would love to have them indexed

**Thanks for taking a look!**


s-pace:
FYI https://docsearch.algolia.com/docs/faq/#can-you-index-code-samples

PR #1782, opened 2020-03-11
updating FamilySearch config with new location of docs

joeycozza:
We made some changes to where our docs are located, so updating the start_urls and made sure to include the stop_urls for endpoints we don't want searched.


s-pace:
It seems this new URL required a login. It will not be handled by DocSearch then.

PR #1792, opened 2020-03-17
Add language and version attributes for faceting

tmkasun:
# Pull request motivation(s)

Can't search from different versions of the documentation

We currently have API Manager (APIM) 3.0.0[1], 3.1.0[2] docs spaces and the /next[3] always points to the bleeding edge version documentation, when we do a release, /next -> becomes the released version (i:e 3.2.0 )  and /next will continue to have latest docs content.

If I switch from APIM 3.0.0 to APIM 3.1.0 doc space, and search something from that page, I should only see the results from APIM 3.1.0 document space.

Configuration change

As per the instruction in here https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags , to address https://github.com/wso2/docs-apim/issues/838 issue

### DIFF
--- a/configs/wso2.json
+++ b/configs/wso2.json
@@ -19,5 +19,11 @@
   "conversation_id": [
     "1070576193"
   ],
+  "custom_settings": {
+    "attributesForFaceting": [
+      "language",
+      "version"
+    ]
+  },
   "nb_hits": 15044
 }


s-pace:
I think you have forgotten to include the docsearch meta tags @tmkasun

tmkasun:
Hi @s-pace ,

Ahh yes thanks, I will update our doc page metatags soon

PR #1823, opened 2020-03-28
Reverting recent Bubble Manual change

ajyang818:
Sylvain just helped us update the config files for the Bubble Manual, but unfortunately this has caused the entire navigation of the site to break. We tried debugging it, but there might be problems stemming from the fact that we're on legacy Gitbooks + using a 3rd party plugin to get Algolia Docsearch. So, we'd like to try reverting everything to how it was earlier today to see if the problem is fixed. (The original issue we reached out to Support with is minor enough that we're willing to go without fixing it.)

# Pull request motivation(s)

### DIFF
--- a/configs/bubble_manual.json
+++ b/configs/bubble_manual.json
@@ -1,7 +1,7 @@
 {
   "index_name": "bubble_manual",
   "start_urls": [
-    "https://manual.bubble.io/"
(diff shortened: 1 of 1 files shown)


ajyang818:
Actually everything seems to be fixed now, so it isn't this config change...closing out, please ignore!

PR #1824, opened 2020-03-28
Update caddyserver.json

mholt:
# Pull request motivation(s)

Add indexing for definition lists. Some of our docs use HTML definition lists (`<dl>`, `<dt>`, and `<dd>` tags) for structure. For example: https://caddyserver.com/docs/json/#field-list-header

They don't appear to be indexed.

I am hoping they can be indexed. :)

Neither -- just an enhancement.

Loving the search integration so far, thank you!!


mholt:
Thank you!!

PR #1832, opened 2020-03-31
jenkins x config: enable search by language

MarckK:
# Pull request motivation(s)

We would like the [Jenkins X documentation site](https://jenkins-x.io/) to be able to search by language, as the site is translated into 3 languages.

The [Jenkins X documentation site](https://jenkins-x.io/) is not currently searching by language.

For example, if I search Prow, the top results include the main Prow section within other translations. This is true if I search within the English translation or within the Spanish translation. 

Screenshot 2020-03-31 at 16 39 00

Screenshot 2020-03-31 at 16 40 20

We would prefer if the search results were prioritised for results matching the language translation of the user.

For example, searches in Spanish would filter for results in Spanish (or at least prioritise them within the search results list)

The current PR should enable the desired search functionality for non-English translations. However, the site is currently structured so that English is the default language, therefore there is not a `jenkins-x.io/en/` directory. 

For searches with users on the English translation, I am not sure the PR will change the current functionality. How would we ensure that searches in English are filtered for English language results?

### DIFF
--- a/configs/jenkins_x.json
+++ b/configs/jenkins_x.json
@@ -1,7 +1,15 @@
 {
   "index_name": "jenkins_x",
   "start_urls": [
-    "https://jenkins-x.io/"
+    {
+      "url": "http://www.example.com/(?P<lang>.*?)/",
+      "variables": {
+        "lang": ["es", "zh"]
+      }
+    },
+    {
+      "url": "https://jenkins-x.io/"
+    }
   ],
   "sitemap_urls": [
     "https://jenkins-x.io/sitemap.xml"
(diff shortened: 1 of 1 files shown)


configs/jenkins_x.json
```diff
@@ -1,7 +1,15 @@
...
   "start_urls": [
-    "https://jenkins-x.io/"
+    {
+      "url": "http://www.example.com/(?P<lang>.*?)/",
+      "variables": {
+        "lang": ["es", "zh"]
+      }
+    },
+    {
+      "url": "https://jenkins-x.io/"
```
s-pace:
```suggestion
      "url": "https://jenkins-x.io/",
      "extra_attributes": {
        "lang": [
          "en"
        ]
        }
```

PR #1856, opened 2020-04-13
fix: update knorm configs

joelmukuthu:
# Pull request motivation(s)

Knorm's documentation site is now generated with Docusaurus v2, having moved from vuepress.

Searches point to non-existent URLs

Searches should point to correct URLs

N/A

N/A


joelmukuthu:
Hi, I'm not sure if this is the right config for a docusarus v2 site but I copied it off of other docusaurus 2 site configs. Please let me know if I should fix something or please do so if you have the time. 

Best,
Joel

PR #1859, opened 2020-04-15
Update steamspeak.json

dalexhd:
# Pull request motivation(s)

It doesn't index my documentation.

No

I don't know if removing that last `/docs/` part fixes my problem, since this returns 404 error. I saw another configurations like [https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json](docusaurus-2.json) and it has the same url system as mine, but it also returns 404 error.  Why my documentation is not updating?

### DIFF
--- a/configs/steamspeak.json
+++ b/configs/steamspeak.json
@@ -1,7 +1,7 @@
 {
   "index_name": "steamspeak",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


dalexhd:
Fixed thanks ^^

PR #1880, opened 2020-04-23
update: bootstrap-vue.js.org to bootstrap-vue.org

tmorehouse:
# Pull request motivation(s)

We have just obtained our own domain `bootstrap-vue.org`, and will be 301 redirecting `bootstrap-vue.js.org`  to the new domain (both domains are currently hosting the same documentation site)

Working fine

Working as is (but crawling the new URL)

Neither

If you have any questions, just ask.

/cc @jackmu95 @pi0


tmorehouse:
Thanks!

PR #1901, opened 2020-05-01
fix: proper content selection & use sitemap instead of crawling from root

harrysolovay:


configs/aws_amplify_new.json
```diff
@@ -1,28 +1,22 @@
...
+    "lvl5": "amplify-toc-contents h4, amplify-toc-content th",
+    "lvl6": "amplify-toc-contents h5",
+    "text": "amplify-toc-contents p, amplify-toc-contents li, amplify-toc-content td"
   },
   "js_render": true,
-  "js_wait": 2,
-  "conversation_id": [
-    "1115766863"
-  ],
+  "js_wait": 10,
```
s-pace, line 19: Way too big. Please do server side rendering instead

configs/aws_amplify_new.json
```diff
@@ -1,28 +1,22 @@
 {
   "index_name": "aws_amplify_new",
-  "start_urls": [
-    "https://docs.amplify.aws/",
```
s-pace, line 4: Please keep it

configs/aws_amplify_new.json
harrysolovay: Reverted it to 2

configs/aws_amplify_new.json
harrysolovay: Put it back in place

harrysolovay:
@s-pace please let me know if there're any other changes that need to be made. Thank you!

PR #1906, opened 2020-05-04
Fix #1898

### LINKED ISSUE #1898: Arrow-kt.io results have been reduced
# Do you want to request a feature or report a bug? A bug

`index_name`=arrow

Some pages aren't being found now. For instance https://arrow-kt.io/docs/0.10/apidocs/arrow-core-data/arrow.core/-option/index.html when searching by `Option`.

Previous page was found before.

#### Any quick clues?

The searchable content and the number of records that are extracted and indexed seem to be explicitly reduced as part of a recent change: https://github.com/algolia/docsearch-configs/commit/44ec7e22dd822afd9477c59a9ab7d75967217080 

Please, could these links be replaced?
```
    "https://arrow-kt.io/docs/0.10/",
    "https://arrow-kt.io/docs/0.10/core/"
```
by:
```
    "https://arrow-kt.io",
```

### Any other feedback / questions ?

Thank you very much.

rachelcarmena:
# Pull request motivation(s)

Fix #1898
- Remove hard-coded version `0.10`
- Add `h1` header

https://arrow-kt.io/sitemap.xml has been updated to have `/current/` as well instead of `/0.10/` on URLs.

`Option` on a header with level `1` is not being found.

Getting `Option` results when looking for `Option`. Just `Optional` and `OptionT` now.

##### NB2: Any other feedback

Thank you so much @s-pace for his kind assistance :+1:

### DIFF
--- a/configs/arrow.json
+++ b/configs/arrow.json
@@ -2,14 +2,14 @@
   "index_name": "arrow",
   "start_urls": [
     {
-      "url": "https://arrow-kt.io/docs/apidocs/",
+      "url": "https://arrow-kt.io/docs/current/apidocs/",
       "tags": [
         "api"
       ],
       "selectors_key": "api"
     },
-    "https://arrow-kt.io/docs/0.10/",
-    "https://arrow-kt.io/docs/0.10/core/"
+    "https://arrow-kt.io/docs/current/",
+    "https://arrow-kt.io/docs/current/core/"
   ],
   "stop_urls": [
     "https://arrow-kt.io/docs/arrow/data/nonemptylist/"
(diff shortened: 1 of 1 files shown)


s-pace:
This will not fix the issue since https://arrow-kt.io/docs/current/ is not available. Can you. make it happen please?

rachelcarmena:
> This will not fix the issue since https://arrow-kt.io/docs/current/ is not available. Can you. make it happen please?

Sorry, it's not available. I thought about a different behavior because of having `https://arrow-kt.io/docs/0.10/` in that file before. I proposed a simplification here: #1907

PR #1909, opened 2020-05-05
Update Tachiyomi config (w/ questions)

Soitora:
# Pull request motivation(s)

It has gone a couple of days and some updated entries has not updated.

I would expect an update every 24 hours as I've been explained is gonna happen.

N/A

First of all, [here's a link to the repo](https://github.com/tachiyomiorg/website/graphs/contributors) that show that I'm clearly a maintainer.
Second, I got some questions before we proceed:

- What does "`scrap_start_urls`" do? I only find info about "`scrape_start_urls`" which by definition should do entirely different things. I do want my things scraped decently frequenly considering how often our docs are updated.

- Regarding "`selectors_exclude`", how exactly does it work? For example:
```
# Application theme <span class="badge default-indicator">Main theme</span>
```
This makes Algolia present it as:
> Application theme Main theme

image

image

To remove just the "**Main theme**" part, should I add `.badge` to the `selectors_exclude` list or will that remove the entire "**Application theme**" entry too?

----

Thank you for providing the Algolia Docsearch, it's a really good resource for our website for the users to more easily find information they require.

### DIFF
--- a/configs/tachiyomi.json
+++ b/configs/tachiyomi.json
@@ -1,9 +1,17 @@
 {
   "index_name": "tachiyomi",
   "start_urls": [
-    "https://tachiyomi.org/help/"
+    "https://tachiyomi.org/help/",
+    "https://tachiyomi.org/help/faq/",
+    "https://tachiyomi.org/help/guides/",
+    "https://tachiyomi.org/help/contribution/",
+    "https://tachiyomi.org/forks/",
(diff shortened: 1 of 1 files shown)


Soitora:
Are you able to answer the questions I asked?

s-pace:
Selector_exclude removes all element matching the selector and only these elements. It should do the trick on your documentation yes

PR #1931, opened 2020-05-09
Update start_urls

beutlich:
Update start_urls to filter for library, tool and version.

FYI @pylypko


beutlich:
Hm, number of search hits decreased from 387614 to 1042. This was not intentional.

s-pace:
What are the missing URLs?

PR #1932, opened 2020-05-09
Add sitemap_urls

beutlich:
Add sitemap_urls. In #1931 the number of hits decreased by accident.

FYI @pylypko

### DIFF
--- a/configs/modelica.json
+++ b/configs/modelica.json
@@ -160,6 +160,9 @@
       }
     }
   ],
+  "sitemap_urls": [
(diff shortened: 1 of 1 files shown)


beutlich:
The number of hits did not increase again as expected. Do I need to wait for the next crawl or is there something messed up in the config?

s-pace:
It is not related to a new crawl but to the start_urls you have set. I will try to make the crawl broader now

beutlich:
> I will try to make the crawl broader now

Thanks a lot. It works again, except the small typo to be fixed by #1933.

PR #1934, opened 2020-05-10
Fix url

beutlich:
FYI @pylypko

### DIFF
--- a/configs/modelica.json
+++ b/configs/modelica.json
@@ -41,55 +41,14 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


beutlich:
Do you have an idea why filtered search with `"facetFilters": ["tool:WSM", "version:3.2.3", "library:ModelicaServices"]` for "exit" returns zero matches though there is https://doc.modelica.org/Modelica%203.2.3/Resources/helpWSM/ModelicaServices/ModelicaServices.System.exit.html?

beutlich:
Thanks. This fixed the "tool:WSM" filter. But need to think for the "tool:OM" filter since "library:ModelicaServices" or "library:ModelicaReference" are already patern-matched by "library:Modelica".

PR #1936, opened 2020-05-10
Fix url pattern to force non-greedy matches

beutlich:
FYI @pylypko

Hopefully, this fixes:

Modelica.html -> library:Modelica
Modelica.Blocks.html -> library:Modelica
ModelicaServices.html -> library:ModelicaServices
etc

### DIFF
--- a/configs/modelica.json
+++ b/configs/modelica.json
@@ -22,7 +22,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


beutlich:
That seems not to have been effective, since `"facetFilters": ["tool:OM", "library:ModelicaServices"]` results in zero matches. Can you please support?

PR #1942, opened 2020-05-11
Set reachable start_urls

beutlich:
One more try.


beutlich:
I also reversed the URL sorting in the sitemap.xml.

PR #1960, opened 2020-05-14
update btcpayserver

dennisreimann:
# Pull request motivation

We've migrated [our docs](https://docs.btcpayserver.org/) from GitBook to Vuepress and the selectors need to be changed.

Right now our old GitBook pages are indexed.

The new Vuepress pages should get indexed.


dennisreimann:
Hey @s-pace, thanks for merging this so quickly 🚀

When will the site get reindexed?

PR #1968, opened 2020-05-18
update mxnet website config - python api

ys2843:
Add stop url and refined selectors for Python API site, tested locally.


ys2843:
@s-pace Hi pace, could you please take a look why CI failed? The JSON format should be valid. Thanks.

PR #1972, opened 2020-05-19
Index more doc content element types

juliusv:
Best would maybe be to index anything under ".doc-content" regardless of sub-tag, but I'm not sure if that would cause any problems.

# Pull request motivation(s)

Certain relevant documentation text elements (in divs, tables, etc.) are not indexed, and thus cannot be found via Docsearch.

For those elements to be found :)

### DIFF
--- a/configs/prometheus.json
+++ b/configs/prometheus.json
@@ -11,7 +11,7 @@
     "lvl3": ".doc-content h4",
     "lvl4": ".doc-content h5",
     "lvl5": ".doc-content h6",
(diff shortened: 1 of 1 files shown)


juliusv:
@s-space Hmm, how does the linked commit relate to this PR and close it?

s-pace:
My bad. https://github.com/algolia/docsearch-configs/commit/b590647b460e2fd735a6be5c092a44315b559b4a#diff-faa65298ccae9f682c816a36a3e79a02 resolves it

PR #1982, opened 2020-05-21
Adding tables to content that is indexed

nathlaroche:
content from table does not show up in search results. Trying to add table element to text selector to see if they will be picked up.

### DIFF
--- a/configs/nuance-mix.json
+++ b/configs/nuance-mix.json
@@ -13,7 +13,7 @@
     "https://mix.nuance.com/v3/documentation/mix-datapack/v6_x/"
   ],
   "selectors": {
-    "text": ".content p, .content li",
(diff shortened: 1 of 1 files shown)


configs/nuance-mix.json
```diff
@@ -13,7 +13,7 @@
     "https://mix.nuance.com/v3/documentation/mix-datapack/v6_x/"
   ],
   "selectors": {
-    "text": ".content p, .content li",
+    "text": ".content p, .content li, .content table",
```
s-pace, line 16:
```suggestion
    "text": ".content p, .content li, .content td",
```

s-pace: Please use the smallest textual element possible

PR #1994, opened 2020-05-27
revert changes to demisto.json

csestito:
# Pull request motivation(s)

Our indexing broke after recent changes, reverting to the last working version 

For algolia to work again on our site

**bug**

### DIFF
--- a/configs/demisto.json
+++ b/configs/demisto.json
@@ -14,14 +14,13 @@
       "global": true,
       "default_value": "Documentation"
     },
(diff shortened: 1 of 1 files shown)


csestito:
xsoar.pan.dev is still not showing results. The fix submitted to our repo at https://github.com/demisto/content-docs/pull/143 wasn't successful in fixing the issue. The only thing that changed overnight is this file, and I suspect has to be the culprit of the breakage.

PR #2003, opened 2020-06-01
Update lvl0 selector for new docs version in projectcalico.json

rymarchikbot:
I would like to update project calico crawler configuration to support new versions of docs.


s-pace:
It is already deployed by ec86c3b

s-pace:
Is the current class set on the server side or thanks to JS?

PR #2004, opened 2020-06-01
Temporarily block new SDK reference sub-folder

jackieros:
We are releasing a new SDK reference under the folder: documentation/sdks/php/....
(In the future, we'll release other SDK references under documentation/sdks/<language>/ )

The contents of this reference will be displayed in a separate tab from the main docs, and will have a different layout & CSS.

In order to properly handle this new content in our search, we'll need to define selector_keys for that output, and we'll also have to add some code so that any results from that reference will open in a new tab and not in the main doc page tab.

Until we are ready to do that, I think it's best that we just stop the crawler from crawling these subfolders at all.

### DIFF
--- a/configs/cloudinary.json
+++ b/configs/cloudinary.json
@@ -87,6 +87,7 @@
   ],
   "stop_urls": [
     "https://cloudinary.com/documentation/upload_widget_1",
+    "https://cloudinary.com/documentation/sdks/*",
     "\\?",
     "\\.txt$"
   ],


jackieros:
@s-pace - Thanks for the merge. Just double-checking that the line I added is the correct regular expression to block the crawler from anything under that sub-folder? 

And a couple additional questions: 
Currently, the stop_url reg expression I defined is for all pages under: `https://cloudinary.com/documentation/sdks, but currently that URL itself returns a 404 as it's just an s3 container. I'm waiting on my devops team to define a redirect for it. 
1. Will the regular expression referencing all pages under that folder currently work even though that specific URL currently returns 404?
2. The plan is for documentation/sdks URL to redirect back to a page under documentation/
(https://docs-dev.cloudinary.com/documentation/cloudinary_sdks).  If it redirects to a page that the crawler IS supposed to crawl, will that cause problems for using it as a stop_url?

s-pace:
Yes it is.

1.
All URL matching this pattern will not be indexed. So these pages will not be indexed.

2.
Since the page is not on the same domain, the crawler will not follow

PR #2024, opened 2020-06-08
Update domain

stancl:
Fixed.


stancl:
@s-pace Thanks for merging. Is there anything we can do to have better "headings" in the search widget? For example here, the `TenancyServiceProvider` is under the `Event system` page, but it shows `Concepts` because that's the parent category.

Screen Shot 2020-06-09 at 13 25 29
Screen Shot 2020-06-09 at 13 25 33
Screen Shot 2020-06-09 at 13 25 41

PR #2025, opened 2020-06-09
Update hail_is.json

akotlar:
Attempt at indexing method names

We have sphinx doc structure that looks like this (you may check https://hail.is/docs/0.2/functions/index.html for more examples).
Screenshot 2020-06-08 23 00 43

Namely, in a .section, we may find

```
.section
   h1/h2/h3/h4
   dl
     dt
```
Children of `<dt>` contain the nearest anchor tag to the relevant results, most of the time.

Currently, the config only indexes the h1/h2/h3/h4, so matches always link to the nearest h1/h2/h3/h4, which reduces the relevance of the link. I would like to link to the nearest `dt`.

This config is untested; in particular, not sure if the selector should read `dl > dt` or the `dl > dt > .descname` (where the .descname element is the one that contains the method name; I think we want the selector that can be linked to via an id, which is `dt`)

Links to relevant matches are to the nearest h1/h2/h3/h4

Should link to the nearest method block.

So for instance, instead of linking to https://hail.is/docs/0.2/hail.Table.html#Table when searching for something that matches "add_index", I want to link to https://hail.is/docs/0.2/hail.Table.html#hail.Table.add_index

Feature

### DIFF
--- a/configs/hail_is.json
+++ b/configs/hail_is.json
@@ -12,7 +12,7 @@
     "lvl2": ".section h3",
     "lvl3": ".section h4",
     "lvl4": ".section h5",
-    "lvl5": ".section dl dt .descname",
+    "lvl5": ".section dl dt",
     "text": ".section p, .section li"
   },
   "strip_chars": " .,;:#",
--- a/configs/hail_is.json
+++ b/configs/hail_is.json
@@ -12,7 +12,7 @@
     "lvl2": ".section h3",
     "lvl3": ".section h4",
     "lvl4": ".section h5",
-    "lvl5": ".section h6",
+    "lvl5": ".section dl dt .descname",
     "text": ".section p, .section li"
   },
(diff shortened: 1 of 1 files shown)


akotlar:
Relevant to this PR: I would like to also index the method name listed in `dl > dt > .descname`, exactly (preserving underscores), and prioritize hits including this term (for instance, I would like to be able to turn use `exactOnSingleWordQuery`). At the same time, I also want to be able to match those method names that include the underscore, if I fail to type the underscore. So basically, I want to index both the exact term (say `method_with_underscore`) and the term split on underscore (such that I can match `method with underscore`).

You can see the current behavior here: http://34.207.246.132

When I search "import_vcf" I should get the "import_vcf" function as the first hit, rather that "Import GVCFs".

If you could assist with this, would be great, thank you!

PR #2027, opened 2020-06-09
custom selectors for overview page

taapsa-visma:


configs/maventa.json
```diff
@@ -1,22 +1,32 @@
 {
   "index_name": "maventa",
   "start_urls": [
-    "https://documentation.maventa.com/"
+    "https://documentation.maventa.com/",
```
s-pace, line 4: Should be the last as it will always match first all URLs

configs/maventa.json
```diff
@@ -1,22 +1,32 @@
...
+      "lvl1": ".col-sm-9 h2",
+      "lvl2": ".col-sm-9 h3",
+      "lvl3": ".col-sm-9 h4",
+      "lvl4": ".col-sm-9 h5, .col-sm-9 summary",
+      "lvl5": ".col-sm-9 h6",
+      "text": ".col-sm-9 p, .col-sm-9 li"
+    },
+    "overview": {
+      "lvl0": ".col-sm-9 h1",
+      "lvl1": ".col-sm-9 h2"
```
s-pace, line 25: You need to define all selectors within a set. Please add other lvls

configs/maventa.json
taapsa-visma: moved to last

configs/maventa.json
taapsa-visma: page contains only h1 and h2 and text. text added

PR #2036, opened 2020-06-11
Update hail_is.json

akotlar:
# Pull request motivation(s)
Remove impact of change_log.html.

Changelog often comes up in top results

Changelog should be substantially de-prioritized.

Feature

Better way to handle this?


akotlar:
decided to remove change_log.html from index until we have a better solution for downweighing its hits

PR #2045, opened 2020-06-14
Update custom faceting attributes

juliusv:
We want to use the "include" facet (generated by meta tags) for deciding which pages to include in the search index.
The "prometheus-version" tag needs to be kept for the transition, but can be removed later, once we switch the site to using "include"
(after it is indexed).


juliusv:
@s-pace

PR #2047, opened 2020-06-16
Update growingio config

changvvb:
# Pull request motivation(s)

Update growignio config

Crawl more two growingio doc urls.


changvvb:
Thanks

PR #2079, opened 2020-07-01
Update data-driven-forms.json to wait

rvsia:
# Pull request motivation(s)

- DDF documentation page is based on next.js, most of pages are server-side rendered, however there are examples using query parameters to change the content of the page. The query parameter is not available on the first render in next.js so we have to wait until the page is updated.

Algolia loads wrong data

Algolia should load updated data

cc @Hyperkid123

### DIFF
--- a/configs/data-driven-forms.json
+++ b/configs/data-driven-forms.json
@@ -19,5 +19,5 @@
   ],
   "nb_hits": 3573,
   "js_render": true,
-  "js_wait": 2
+  "js_wait": 1
 }
--- a/configs/data-driven-forms.json
+++ b/configs/data-driven-forms.json
@@ -17,5 +17,7 @@
   "conversation_id": [
     "952910917"
   ],
-  "nb_hits": 3573
+  "nb_hits": 3573,
+  "js_render": true,
+  "js_wait": 2
 }


configs/data-driven-forms.json
```diff
@@ -17,5 +17,7 @@
   "conversation_id": [
     "952910917"
   ],
-  "nb_hits": 3573
+  "nb_hits": 3573,
+  "js_render": true,
+  "js_wait": 2
```
s-pace, line 22: Would it be possible to reduce as much as possible the time to wait?

rvsia: I would like to, is there any way how to determine it? Or should I try some smaller value?

Hyperkid123: The query params should be available almost instantly. The issue is that the router needs to just access the window property and set query values. It should take less than a second IMO.

s-pace:
```suggestion
  "js_wait": 1
```
Let's give it a try with 1.

s-pace:
Crawl done. Can you please check that the result is as expected?

Hyperkid123:
@s-pace seems to be working. We are getting the search results.

PR #2081, opened 2020-07-03
Update nervos_ckbd.json

xxuejie:
# Pull request motivation(s)

### DIFF
--- a/configs/nervos_ckbd.json
+++ b/configs/nervos_ckbd.json
@@ -1,10 +1,10 @@
 {
   "index_name": "nervos_ckbd",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


s-pace:
http://docs.nervos.org/docs/ is not available and still redirects to the docs-new domain. This PR only frozes the crawl.

@xxuejie can you make sure http://docs.nervos.org/docs/ is available and comply with our PR template please?

xxuejie:
@s-pace I think there has been a mistake in the original template, our doc site should just live at `docs.nervos.org`(which now redirects to `docs-new.nervos.org` since we want to keep it still running while we are doing the updates here).

I will make a different PR that updates the template.

PR #2084, opened 2020-07-04
Update intuit_auto.json

hipstersmoothie:
# Pull request motivation(s)

1. Fix lvl0 to match sidebar headings. looked at tailwind's config for reference
2. separate selectors for blog (referenced gatsby config)
3. add sitemap
4. change page rank so blog is at bottom and plugin authoring docs are lower than all other docs

current behavior treats page h1 as lvl0 which makes the docs search slightly off.

Sidebar title are lvl0 making docs more searchable.

### DIFF
--- a/configs/intuit_auto.json
+++ b/configs/intuit_auto.json
@@ -1,24 +1,55 @@
 {
   "index_name": "intuit_auto",
   "start_urls": [
-    "https://intuit.github.io/auto/"
+    {
+      "url": "https://intuit.github.io/auto/",
+      "page_rank": 5,
+      "tags": ["docs"]
+    },
+    {
+      "url": "https://intuit.github.io/auto/plugins",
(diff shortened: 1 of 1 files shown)


configs/intuit_auto.json
```diff
@@ -1,24 +1,55 @@
 {
   "index_name": "intuit_auto",
   "start_urls": [
-    "https://intuit.github.io/auto/"
+    {
+      "url": "https://intuit.github.io/auto/",
+      "page_rank": 5,
+      "tags": ["docs"]
+    },
```
s-pace, line 8: Should be the last item as the first matching prevail

hipstersmoothie: fixed1

hipstersmoothie:
@s-pace should be good now

PR #2091, opened 2020-07-10
update mdanalysis.json

orbeckst:
# Pull request motivation(s)

We added more documentation to our site https://www.mdanalysis.org and also now use dedicated subdomains https://docs.mdanalysis.org/ and https://userguide.mdanalysis.org/. To keep the configuration simple, we created a sitemapindex as starting point for the Algolia crawler that we would like to activate (see https://github.com/MDAnalysis/MDAnalysis.github.io/issues/79)

## Detailed changes
- use sitemapindex.xml instead of start_urls
  - www.mdanalysis.org 
  - docs.mdanalysis.org
  - userguide.mdanalysis.org
- stop indexing old docs (www.mdanalysis.org/docs and www.mdanalysis.org/mdanalysis)
- stop indexing generated docs for the three domains (not sure if we can use regular expressions for subdomains - to be safe, I repeated them)
- completes https://github.com/MDAnalysis/MDAnalysis.github.io/issues/79

Only a small subset of our site is indexed. Furthermore, these parts are now outdated due to site reorganization.

Current docs appear in search.


configs/mdanalysis.json
```diff
@@ -1,17 +1,33 @@
...
+      "variables": {
+        "generated": ["_modules", "_sources"]
+      }
+    },
+    {
+      "url": "https://userguide.mdanalysis.org/.*?/(?P<generated>.*?)/",
+      "variables": {
+        "generated": ["_modules", "_sources"]
+      }
+    }
```
s-pace, line 30:
```suggestion
    "https://www.mdanalysis.org/docs",
    "\\/_"
```

You can not use variables on `stop_urls`. The suggested pattern is easier to catch and help to avoid crawling auto generated pages.

orbeckst:
Thank you @s-pace ! I added your change and merged master.

orbeckst:
Thank you for the quick turn-around, much appreciated!

PR #2097, opened 2020-07-15
Update apache_camel.json

aashnajena:
@zregvart is a maintainer of the [Camel Website repository](https://github.com/apache/camel-website) and can approve these changes.

These changes are made for the following reasons :
- To include custom selectors for different start_urls
- To avoid duplicate records
- To index only content-level records

### DIFF
--- a/configs/apache_camel.json
+++ b/configs/apache_camel.json
@@ -1,17 +1,55 @@
 {
   "index_name": "apache_camel",
-  "stop_urls": [],
+  "only_content_level": true,
+  "stop_urls": [
+    "https://camel.apache.org/categories/",
+    "https://camel.apache.org/blog/page/",
+    "index.html"
+  ],
+  "start_urls": [
+    { "url": "https://camel.apache.org/blog/", "selectors_key": "blog" },
+    { "url": "https://camel.apache.org/camel-k/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/camel-quarkus/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/camel-kafka-connector/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/camel-spring-boot/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/camel-karaf/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/components/", "selectors_key": "antora-pages" },
+    { "url": "https://camel.apache.org/manual/", "selectors_key": "antora-pages" },
+    "https://camel.apache.org/download/",
+    "https://camel.apache.org/community/books",
+    "https://camel.apache.org/community/articles",
+    "https://camel.apache.org/docs/"
+  ],
   "sitemap_urls": [
     "https://camel.apache.org/sitemap.xml"
   ],
   "selectors": {
-    "lvl0": "article h1",
-    "lvl1": "article h2",
-    "lvl2": "article h3",
-    "lvl3": "article h4",
-    "lvl4": "article h5",
-    "lvl5": "article h6",
-    "text": "article p, article li"
+    "default": {
+      "lvl0": { "selector": "", "default_value": null },
+      "lvl1": "article h1",
+      "lvl2": "article h2",
+      "lvl3": "article h3",
+      "text": "article p, article li"
+    },
+    "antora-pages": {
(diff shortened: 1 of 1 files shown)


configs/apache_camel.json
```diff
@@ -1,20 +1,58 @@
...
   "selectors": {
-    "lvl0": "article h1",
-    "lvl1": "article h2",
-    "lvl2": "article h3",
-    "lvl3": "article h4",
-    "lvl4": "article h5",
-    "lvl5": "article h6",
-    "text": "article p, article li"
+    "default": {
+      "lvl0": { "selector": "", "default_value": null },
```
s-pace:
```suggestion
      "lvl0": { "selector": "", "default_value": "" },
```
default_value can not be `null`

configs/apache_camel.json
```diff
@@ -1,20 +1,58 @@
...
+      "lvl1": "div.toolbar ul li:nth-child(1) a",
+      "lvl2": "div.toolbar ul li:nth-child(2) a",
+      "lvl3": "div.toolbar ul li:nth-child(3) a",
+      "lvl4": "div.toolbar ul li:nth-child(4) a",
+      "lvl5": "article h2",
+      "lvl6": "article h3",
+      "text": "article p, article li"
+    },
+    "blog": {
+      "lvl0": { "selector": "", "default_value": null },
```
s-pace: Same feedback here. Why don't you want to define the `lvl0`? It helps to provide more context to your end users

aashnajena: I have changed this to include a selector here.

configs/apache_camel.json
```diff
@@ -1,20 +1,85 @@
...
+      "lvl2": "div.toolbar ul li:nth-child(2) a",
+      "lvl3": "div.toolbar ul li:nth-child(3) a",
+      "lvl4": "div.toolbar ul li:nth-child(4) a",
+      "lvl5": "article h2",
+      "lvl6": "article h3",
+      "text": "article p, article li"
+    },
+    "blog": {
+      "lvl0": "div.nav-panel-explore .version.is-current a",
+      "lvl1": {
```
s-pace, line 70: It is not recommended to define a lvl0 & lvl2 while not defining a lvl1.

aashnajena: We want the lvl1 entry to be "Blog". That's why we have specified an empty selector and a default value. Is there another way to do this?

s-pace: I would recommend you[ to set both lvl0 and lvl1 global then](https://docsearch.algolia.com/docs/config-file/#using-global-selectors).

aashnajena: We don't have the title "Blog" mentioned in any of our blog posts, so we cannot retrieve it from HTML markup.

s-pace: Yes understood, this is why we recommend to use global selectors so that all hits contain this level

aashnajena: Is this better now?

configs/apache_camel.json
```diff
@@ -1,20 +1,89 @@
...
+      "lvl4": "div.toolbar ul li:nth-child(4) a",
+      "lvl5": "article h2",
+      "lvl6": "article h3",
+      "text": "article p, article li"
+    },
+    "blog": {
+      "lvl0": {
+        "selector": "div.nav-panel-explore .version.is-current a",
+        "global": true
+      }
```
s-pace, line 72:
```suggestion
      },
```

s-pace:
@aashnajena & @zregvart I have updated the configuration to enhance the search experience.

We're happy to provide DocSearch free of charges for any documentation website. This project is built to help the community. [As part of our attribution policy](https://docsearch.algolia.com/docs/styling/#attribution), we do ask our user to display our logo from the dropdown. Can you make sure to display the logo from the dropdown please? You [can find the assets here](https://www.algolia.com/press/?section=brand-guidelines). If you are using [InstantSearch under the hood, you can leverage its API quickly](https://www.algolia.com/doc/api-reference/widgets/powered-by/js/)

zregvart:
@s-pace thank you, we indeed use our own presentation for the search results, and the reason for this is that when we initially implemented the search we did not use the Algolia's JavaScript package as it nearly tripled the page size. So we implemented our own AJAX and UI on top of the Algolia API, we did this in quite some haste omitting the attribution. We're now working with two interns from Outreachy (@aashnajena being one of them) to reimplement the search and one of tasks is to add the Algolia attribution.

PR #2104, opened 2020-07-16
fix: wait for JavaScript render and use anchors

sarahdayan:
# Pull request motivation(s)

The goal of this PR is to indicate that the ESLint website renders anchor information at runtime, and for it to wait until JavaScript is done rendering before scraping the pages.

The ESLint website uses JavaScript to generate anchors at runtime for headings. This results in DocSearch not currently indexing all anchors properly, and defaulting to `#top` (the header's id).

Capture d’écran 2020-07-16 à 15 30 20

DocSearch should wait until JavaScript is done executing, so that the HTML contains proper anchors for headings before scraping it.

cc @nzakas

### DIFF
--- a/configs/eslint.json
+++ b/configs/eslint.json
@@ -51,6 +51,5 @@
   },
   "scrap_start_urls": false,
   "js_render": true,
-  "use_anchors": true,
   "nb_hits": 17888
 }
--- a/configs/eslint.json
+++ b/configs/eslint.json
@@ -50,5 +50,7 @@
     }
   },
   "scrap_start_urls": false,
+  "js_render": true,
+  "use_anchors": true,
   "nb_hits": 17888
 }


configs/eslint.json
```diff
@@ -50,5 +50,7 @@
     }
   },
   "scrap_start_urls": false,
+  "js_render": true,
+  "use_anchors": true,
```
s-pace, line 54:
```suggestion
```

This setting's name is misleading. It lets know the scrapper that it should crawl pages with different anchors as different page. It is required for some website where the anchors define the routing such as https://nsfi.github.io/ppfish-components/#/home

We always retrieve the closest anchor from an element extracted. You can remove this parameter

sarahdayan:
Hey @s-pace did you deploy? I still see the issue on ESLint's website.

nzakas:
Thanks for this. We are also looking at doing this server-side, it appears this was broken when we migrated from Jekyll to Eleventy.

s-pace:
It seems that the DOM does not contain any `id` or` name` attribute which can be extracted and use as anchors even when rendered thanks to my favorite browser.

I have removed the usage of the headless chrome as it does not help here and uses superfluous resources. I made sure the crawler removes the DOM elements introducing these meaningless anchors. Changes are live. You can check it out?

If you want to enhance the accuracy of the redirection, please [implement this recommendation from the eslint website](https://docsearch.algolia.com/docs/tips#add-anchors-to-headings)

nzakas:
Thanks for looking into this. I'm fairly certain we just have a misconfiguration somewhere that I can fix.

PR #2105, opened 2020-07-16
update crawler

Giwi:
# Pull request motivation(s)
We wanted to improve the site search index by splitting content and code documentation.

The content is correctly indexed but for code documentation, search results do not provide the right anchor link.

We want the code doc to be indexed only by the H1, but the content must remain as what we have today.

### DIFF
--- a/configs/warp10.json
+++ b/configs/warp10.json
@@ -1,19 +1,33 @@
 {
   "index_name": "warp10",
   "start_urls": [
-    "https://www.warp10.io"
+
+    {
+      "url": "https://www.warp10.io/doc/functionList",
+      "selectors_key": "doc"
+    },
+    {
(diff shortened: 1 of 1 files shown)


s-pace:
👋 @Giwi 

Unfortunately this PR introduces changes that are not correctly handled by the scraper.

Can you make your website server side rendered? It seems that your website requires JS to correctly render. I have tried to use the headless chrome emulation the scraper provides and wait for up to 3 seconds but the crawl is still not working (`js_render=true` & `js_wait=3`). 

An other solution would be to serve only the content matching the selectors depending on the `user_agent` of the crawler. The HTML served would need the textual information along with the appropriate container specifying the correct anchor to extract via an `id` or `name` attribute.

In the meantime, your DocSearch index will not be updated. Let us know.

PR #2106, opened 2020-07-17
Update and rename configs/tencent_wechatpay.json to 吐露/tencent_wechat…

akiraTencent:
up


s-pace:
Please comply with the template of the PR so that we can understand and better assist you.

It is not possible to rename the `configs` folder as it is shared among the community. If you want to rename your DocSearch index, please rename the file and change the `index_name` attribute of the configuration.

PR #2108, opened 2020-07-17
Remove 🔗 character from index

danyelf:
When we added link icons, these made it into the index.

# Pull request motivation(s)

Our site recently changed to use 🔗  as an anchor link. This is useful for users; not so much for search. Let's take it out! Heres' an example of how it looks today:

image

Current behavior previews the icon. We're able to hide it on the site by wrapping in a span <display=hidden> but that gets eaten by algolia

Don't render that icon. Easiest way is to just skip the character in our config

### DIFF
--- a/configs/honeycomb.json
+++ b/configs/honeycomb.json
@@ -20,5 +20,6 @@
   "conversation_id": [
     "428594718"
   ],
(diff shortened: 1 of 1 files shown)


danyelf:
Thank you!

PR #2124, opened 2020-07-27
Adding another attribute for filtering

mattbrailsford:
Need to be able to do an OR filtered search based on package AND version so created a computed attribute to hold a combined value


mattbrailsford:
In case it helps with performance, I've removed the other two attributes for now as the uvid one combines them so there is probably no need in having the extra indexes

PR #2127, opened 2020-07-28
Update and rename configs/tencent_wechatpay.json to 吐露/tencent_wechat…

akiraTencent:
Modify the search file directory


s-pace:
[As described from my previous comment](https://github.com/algolia/docsearch-configs/pull/2106#issuecomment-664249552), this is not possible to edit this directory. Why do you want to edit such directory? It does not impact your DocSearch index. I would recommend you [to check out our documentation to better understand how DocSearch works](https://docsearch.algolia.com/docs/how-does-it-work/). Thanks.

PR #2128, opened 2020-07-28
Update fishd-mobile.json

hangaoke1:
# Pull request motivation(s)

### DIFF
--- a/configs/fishd-mobile.json
+++ b/configs/fishd-mobile.json
@@ -1,7 +1,15 @@
 {
   "index_name": "fishd-mobile",
   "start_urls": [
-    "https://nsfi.github.io/fishd-mobile-site"
+    {
+      "url": "https://nsfi.github.io/fishd-mobile-site/#/(?P<lang>.*?)/",
+      "variables": {
+        "lang": [
+          "en-US",
+          "zh-CN"
+        ]
+      }
+    }
   ],
   "selectors": {
     "lvl0": ".markdown h1",


hangaoke1:
Thanks for reply，I found that the Chinese content of my site was not crawled，so I change the config。[MyDocSite](https://nsfi.github.io/fishd-mobile-site/#/zh-CN/components/quickStart)

s-pace:
This PR doesn't not help. Can you [provide a sitemap](https://docsearch.algolia.com/docs/tips/#use-a-sitemapxml) and [introduce the language via meta tags](https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags) please? Thanks

hangaoke1:
I haven't used sitemap, but I will add facetFilters to my web site to search for documents in the corresponding language. I see other multilingual documents do the same, I don’t know if it is feasible。

s-pace:
We mainly need a sitemap since the scrapper does not browse the Chinese pages.

hangaoke1:
ok，do i need to create a new pr ？

s-pace:
Yes to let us know once it is ready. Thanks

hangaoke1:
hi，[vantConfig](https://github.com/algolia/docsearch-configs/blob/75717a708c7cbf34272369eb3f1e6445c7eb7c9d/configs/vant.json)，
I am very confused. In this configuration, he did not configure the sitemap, but the crawler crawled the Chinese content normally. I don’t know why my site did not take effect。


In addition, the sitemap generated by the tool looks like this
```xml
<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
<url>
<loc>https://nsfi.github.io/fishd-mobile-site/index.html#/zh-CN/components</loc>
<priority>1.0</priority>
<lastmod>2020-07-28</lastmod>
<changefreq>hourly</changefreq>
</url>
</urlset>
```

s-pace:
I do confirm that we crawl both english and chinese pages from the `vant` website while only crawling the english ones from `fishd-mobile`. It seems that the two website are not entirely similar. Can you check what is different between both websites?

As for the sitemap, we can give it a try. You must precise the `sitemap_urls` to let the crawler know. Please submit a PR on your website as soon as it is ready.

Thanks

hangaoke1:
I think I know the problem. I removed the a tag for switching languages ​​from my site, so when the crawler is on the English page, it will never crawl to the Chinese content. I will update my site and observe，thank u very much

PR #2131, opened 2020-07-30
Update maximousblk_piracy.json

maximousblk:
# Pull request motivation(s)

I switched to Docusaurus from Vuepress

The element IDs changed so the links are all broken

The links in the search results direct to the correct page

no

What is `nb_hits` field for?

### DIFF
--- a/configs/maximousblk_piracy.json
+++ b/configs/maximousblk_piracy.json
@@ -27,16 +27,7 @@
     "separatorsToIndex": "_",
     "attributesForFaceting": [
       "language",
-      "version",
-      "type"
-    ],
-    "attributesToRetrieve": [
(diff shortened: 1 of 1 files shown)


configs/maximousblk_piracy.json
```diff
@@ -3,38 +3,45 @@
...
+      "version",
+      "type"
+    ],
+    "attributesToRetrieve": [
+      "hierarchy",
+      "content",
+      "anchor",
+      "url",
+      "url_without_anchor",
+      "type"
```
s-pace, line 39:
```suggestion
      "version"
```
Only needed when you use the latest version of docusaurus v2 featuring the DocSearch v3

PR #2281, opened 2020-08-05
Update unbounded_adapt.json to adaptjs.json

mterrel:
# Pull request motivation(s)

The search index for our open source project site does not appear to be updating and I believe it is likely due to the outdated primary URL in this config file.

The primary URL for the Adapt open source project has moved from https://adapt.unbounded.systems to https://adaptjs.org. This can be verified by simply visiting https://adapt.unbounded.systems and noticing it redirects to https://adaptjs.org .

The config is pointing at the wrong primary URL for the Adapt project.

The config should point to https://adaptjs.org and this will hopefully allow search indexing to update.

I went ahead and changed the name of the configuration file to match the new URL. If that is not correct, let me know. I do notice that the CI check fails, probably because of the renamed file.

### DIFF
--- a/configs/unbounded_adapt.json
+++ b/configs/unbounded_adapt.json
@@ -1,10 +1,10 @@
 {
   "index_name": "unbounded_adapt",
   "start_urls": [
-    "https://adapt.unbounded.systems/docs/"
+    "https://adaptjs.org/docs/"
   ],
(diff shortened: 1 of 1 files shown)


mterrel:
@s-pace I changed the filename back and rebased to current master. Let me know if there's anything else you need.

PR #2287, opened 2020-08-10
Added back .html to stop URL to avoid duplicates

jkelvie:
# Pull request motivation(s)


s-pace:
👋 @jkelvie 

There is no need to add such stop_urls as the website does not reference any page with .html page now. Furthermore the regex `https:\/\/read.bespoken.io\/.+[^h][^t][^m][^l]$` is not correctly built as this regex will regex will match all pages with an URL starting by `https://read.bespoken.io/` due to `.+`. `+` is a greedy quantifier and it will match as many time as possible. The latest element of the regex are ineffective.

PR #2309, opened 2020-08-18
edit config 

1ceb3rg:
Changed so that the last child (the current section is shown). @s-pace If possible I would like to show the combined text of all links/text in the .sectionBreadcrumb dev. If that is not possible then I would prefer to use the last-child.


s-pace:
The change submitted on selector has no effect since both selectors target the same element as [you can see on this page](https://docs.testarchitect.com/using-the-testarchitect-help-system/quick-start-tab/).

I would not recommend to use the deepest element from the breadcrumb anyway since the UI cluster results by same `lvl0`. Using the deepest element will avoid such feature since this value is not consistent. For example result from these pages will not have the same lvl0 while they should since they belongs to the same category `Using the TestArchitect Help system `
https://docs.testarchitect.com/using-the-testarchitect-help-system/desktop-help/searching-for-information-in-desktop-help/
https://docs.testarchitect.com/using-the-testarchitect-help-system/desktop-help/using-local-help-documentation-in-offline-mode-on-windows/

1ceb3rg:
@s-pace When we search for Click there's four pages with that title, all under the parent section "Automation Guide", but they are part of different child sections https://docs.testarchitect.com/automation-guide/action-based-testing-language/testarchitect-automation-classes/automation-classes/abtbutton/click/ https://docs.testarchitect.com/automation-guide/action-based-testing-language/testarchitect-automation-classes/automation-classes/abtentity/click/  etc So I think the results might look better if it showed abtEntity and abtButton as global sections. In this case last-child would target abtentity or abtbutton, correct?  Thanks!

PR #2323, opened 2020-08-24
Update mlflow.json to index dt tags

apurva-koti:
# Pull request motivation(s)

`dt` tags are ignored in MLflow's search indexing. This hampers our documentation as it becomes difficult to find API references for functions and classes, which use are contained within `dt` tags. This similar to the approach taken by `requests3`.

Index `dt` tags, which is not happening right now.


apurva-koti:
@smurching, @ankitmathur-db

apurva-koti:
@s-pace

PR #2300, opened 2020-08-17
Add new Cloudflare Workers search index

signalnerve:
Hi there! We're updating the Cloudflare Workers documentation this week, and will be replacing the existing URLs (https://developers.cloudflare.com/workers) with a totally new structure. 

This PR addresses that issue by making a new config/index, called `workers-cloudflare-v2`. We'll ultimately deprecate the current index, so I'll probably be back in a week or so to remove that file with another PR.

I'm unsure if we're allowed to add files without going through the submission flow on the site, but because we're planning to release our docs tomorrow, I'm hoping that this is OK!

cc-ing some of my coworkers to review this work – @adamschwartz @davidtsong @rita3ko

PR #2328, opened 2020-08-25
Update CF workers docs URL

signalnerve:
This PR updates the CF Workers docs URL as discussed in #2300 to our final production URL.

@s-pace I noticed that a version of this JSON config was moved into the `disabled` directory — would love to get this merged in before we potentially lose our search behavior!1

### DIFF
--- a/configs/cloudflare-workers-v2.json
+++ b/configs/cloudflare-workers-v2.json
@@ -1,11 +1,11 @@
 {
   "index_name": "cloudflare-workers-v2",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


s-pace:
> @s-pace I noticed that a version of this JSON config was moved into the disabled directory — would love to get this merged in before we potentially lose our search behavior!1

This configuration was a duplicated and it has been moved. I have double checked and you are using the right credentials from https://github.com/cloudflare/workers-docs-engine/blob/master/src/components/docs-search.js#L22-L23

PR #2329, opened 2020-08-25
Remove dead link for QuestDB

mpsq:
This PR removes a dead link "https://questdb.io/docs/".


s-pace:
This Page is not crawled. Removing it will narrow down the scope of the crawl to the introduction page only

PR #2331, opened 2020-08-26
Add sitemap URL

Crampichon:
# Pull request motivation(s)
Zendesk articles are available through different URLs.
This causes the DocSearch to display duplicates in the search results
I'm hoping to fix this adding this sitemap with only one URL per article.

### DIFF
--- a/configs/upslide.json
+++ b/configs/upslide.json
@@ -19,5 +19,6 @@
   "conversation_id": [
     "1156149394"
   ],
-  "nb_hits": 4092
+  "nb_hits": 4092,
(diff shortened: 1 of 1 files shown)


s-pace:
The issue with duplicates content is that the same page is able from 2 different URLs. https://github.com/algolia/docsearch-configs/commit/99ef3e5bf863a94b050cee39dbb466e9f1ec4d42 solves this. For example, both following URLs refer to the same page:
- https://support.upslide.net/hc/en-us/articles/360015692380
- https://support.upslide.net/hc/en-us/articles/360015692380-Inserting-Cross-references

PR #2349, opened 2020-09-01
chore(twilio_paste): adjust selectors to exclude data

richbachman:
- Adjusted selectors to remove data
- Adjusted urls to include scoped indexing

### DIFF
--- a/configs/twilio_paste.json
+++ b/configs/twilio_paste.json
@@ -26,22 +26,6 @@
       "selectors_key": "content"
     },
     {
-      "url": "https://paste.twilio.design/content/voice-and-tone/",
-      "selectors_key": "content"
-    },
-    {
-      "url": "https://paste.twilio.design/content/product-style-guide/",
-      "selectors_key": "content"
-    },
-    {
-      "url": "https://paste.twilio.design/content/writing-for-components/",
-      "selectors_key": "content"
-    },
-    {
-      "url": "https://paste.twilio.design/content/get-support/",
-      "selectors_key": "content"
-    },
-    {
       "url": "https://paste.twilio.design/libraries/",
       "selectors_key": "content"
     },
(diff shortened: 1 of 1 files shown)


configs/twilio_paste.json
```diff
@@ -1,27 +1,88 @@
 {
   "index_name": "twilio_paste",
   "start_urls": [
-    "https://paste.twilio.design/"
+    {
+      "url": "https://paste.twilio.design/"
+    },
```
s-pace, line 6:
```suggestion
```
The first `start_url` matching applies. This URL will always match and the other item will have no effect.

configs/twilio_paste.json
```diff
@@ -1,27 +1,88 @@
...
+      "selectors_key": "content"
+    },
+    {
+      "url": "https://paste.twilio.design/content/writing-for-components/",
+      "selectors_key": "content"
+    },
+    {
+      "url": "https://paste.twilio.design/content/get-support/",
+      "selectors_key": "content"
+    },
```
s-pace, line 46:
```suggestion
```
`start_url` https://paste.twilio.design/content/ is enough

configs/twilio_paste.json
```diff
@@ -1,27 +1,88 @@
...
+    {
+      "url": "https://paste.twilio.design/libraries/",
+      "selectors_key": "content"
+    },
+    {
+      "url": "https://paste.twilio.design/roadmap/",
+      "selectors_key": "content"
+    }
```
s-pace, line 54:
```suggestion
    },
    "https://paste.twilio.design/"
```

s-pace (changes requested):

s-pace:
👋 @richbachman 

I have updated the configuration to avoid duplicates. It seems that some pages are available from 2 different URLs. For example:
https://paste.twilio.design/components/spinner/
https://paste.twilio.design/components/spinner

Both URLs are different but still available. Please note the trailing `/`. You might want to add a redirection or make sure only one URL is available.

PR #2352, opened 2020-09-02
fix: doc website not found

andregaldino:
The main website documentation return page not found  [https://docsearch.algolia.com/docs](https://docsearch.algolia.com/docs), I believe that the correct page should be [https://docsearch.algolia.com/](https://docsearch.algolia.com/)


s-pace:
Thank you for your contribution

PR #2367, opened 2020-09-09
Update MLflow search config

apurva-koti:
We are updating our search config to differentiate between the `h2` tags on our R API and REST API pages, and the `h2` tags on all other pages. The former should be equally weighted with `dt` and `h3` tags, and the latter should take priority.

# Pull request motivation(s)

Currently, `h2` tags in the R API and REST API pages take priority despite the page-rank setup, because the `h2` selector level is higher than the `dt` and `h3` tags, which correspond to the docstrings from the Python API (which should always be first).

`h2` tags from R/REST APIs be treated equally to `dt`, `h3` from Python API; all other `h2`s take precedence

### DIFF
--- a/configs/mlflow.json
+++ b/configs/mlflow.json
@@ -2,18 +2,25 @@
   "index_name": "mlflow",
   "start_urls": [
     {
-      "url": "https://mlflow.org/docs/latest/R-api.html",
+      "url": "https://mlflow.org/docs/latest/rest-api.html",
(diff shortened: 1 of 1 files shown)


apurva-koti:
@ankitmathur-db, @smurching

apurva-koti:
Hi @s-pace, could you take a look? Thanks :)

PR #2368, opened 2020-09-10
disable typo tolerance 

1ceb3rg:
# Pull request motivation(s)

Currently results for iOS, SAP, and WPF are not great

When the user searches for SAP results containing SAP should be first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ DocSearch is made of 3 repositories:
   a Docker image
 
 
-[1]: https://docsearch.algolia.com/docs/
+[1]: https://docsearch.algolia.com/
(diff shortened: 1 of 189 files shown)


s-pace:
👋 @1ceb3rg 
PR deployed. As part of our attribution policy, we do ask to display our logo from the search UI. We wouldn't find it on your website as seen on the following screenshot:
image

Can you please make sure to display it?

1ceb3rg:
@s-pace thanks! Yes, I noticed that yesterday. I think it's a side effect of something one of the web devs was working on. It has been noted and should be fixed soon.

PR #2369, opened 2020-09-10
config: add top level selector as a facet

pa-eps:
Added the `lvl0` selector as one of the facets.

The following forum thread helped to figure out what kind of
changes were needed in the config file. If I have missed anything,
please let me know.

https://discourse.algolia.com/t/how-to-configure-a-new-facet/10776

# Pull request motivation(s)

At Elastic Path, we are building a custom Search UI for our docs
site that is integrated with Algolia DocSearch service. In our
custom search UI, we need to add an option to be able to filter
search results based on results "Category". This "Category" is
same or equivalent of `lvl0` selector.

Currently I have 2 facets that are configured (i.e. language, version). In my custom search UI built using algolia react instantsearch library, I am able to easily perform filters on those 2 facets. But, I need an addition facet.

Once the new facet is added into the configuration and my site has been indexed, I expect to be able to add an additional filter option in my custom search UI.

n/a

n/a

### DIFF
--- a/configs/elasticpath_commerce-cloud.json
+++ b/configs/elasticpath_commerce-cloud.json
@@ -28,7 +28,8 @@
   "custom_settings": {
     "attributesForFaceting": [
       "language",
-      "version"
+      "version",
+      "hierarchy.lvl0"
     ]
   },
   "scrape_start_urls": false,


pa-eps:
@s-pace - Sorry for tagging you directly. I don't know all the maintainers of this repo or if it has any notification set when new PRs are opened for config changes. Would you be able to review the changes in this PR. Please let me know if you see any potential issues due to the changes.

PR #2420, opened 2020-09-24
Update rasa docsearch

m-vdb:
# Pull request motivation(s)

Hi there, my teammate @lunelson and I have been in contact with Sylvain via email to update [Rasa](https://rasa.com/)'s DocSearch index. We've updated our documentation from Sphinx to Docusaurus (v2, the future 🚀 ) and would like to configure an index for 3 different sites:
- https://rasa.com/docs/rasa
- https://rasa.com/docs/rasa-x
- https://rasa.com/docs/action-server

We've been told that we could use [custom tags](https://docsearch.algolia.com/docs/config-file/#using-custom-tags) to achieve this, so here is an attempt at updating our config. I'm preparing PRs on our end to update the frontend widget config to hit the right facet on the right site.

Can you let met know if my updates are right? (I don't have prior knowledge about DocSearch, but I know Algolia products a bit)

### DIFF
--- a/configs/rasa.json
+++ b/configs/rasa.json
@@ -16,10 +16,8 @@
   ],
   "stop_urls": [
     "http://rasa.com/docs/.*?/\\d",
-    "http://rasa.com/docs/.*?/master",
-    ".rst.txt$",
+    "http://rasa.com/docs/.*?/next",
     "/_",
-    "/dev/",
     "[^/]$"
   ],
   "selectors": {
--- a/configs/rasa.json
+++ b/configs/rasa.json
@@ -1,7 +1,18 @@
 {
   "index_name": "rasa",
   "start_urls": [
-    "https://rasa.com/docs"
+    {
+      "url": "https://rasa.com/docs/rasa",
+      "tags": ["rasa"]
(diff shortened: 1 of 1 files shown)


Shipow (approved):

configs/rasa.json
```diff
@@ -1,32 +1,40 @@
...
+    "lvl1": ".main h1",
+    "lvl2": ".main h2",
+    "lvl3": ".main h3",
+    "lvl4": ".main h4",
+    "lvl5": ".main h5",
+    "text": ".main p, .main li"
   },
   "selectors_exclude": [
-    ".scv-banner"
+    ".announce-bar"
```
m-vdb, line 37: @lunelson would you mind double checking this config? the PR is already merged, but if there are issues I think we should address them sooner than later 🙏

Shipow: Sure, we'll have a look.

m-vdb: I've made [this second PR](https://github.com/algolia/docsearch-configs/pull/2422) where I also mentioned you @Shipow :)

PR #2433, opened 2020-09-27
Update sap_spartacus.json

gladius-mtl:
Seems I didn't change my config correctly in my previous update, and the search results are now all pointing to only one part of my site (which happens to be the archived docs, not the current docs, gaaah!), so I am reverting my previous change. My apologies for the inconvenience, and I will follow up through another channel to see if I can get a bit of help 😊

# Pull request motivation(s)

### DIFF
--- a/configs/sap_spartacus.json
+++ b/configs/sap_spartacus.json
@@ -2,11 +2,7 @@
   "index_name": "sap_spartacus",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


gladius-mtl:
I made an error in my PR -- I forgot the closing quotes around my site URL. Not sure how to exist an open PR through the web UI, so I will just close and resubmit.

PR #2434, opened 2020-09-27
Update sap_spartacus.json

gladius-mtl:
Seems I didn't change my config correctly in my previous update, and the search results are now all pointing to only one part of my site (which happens to be the archived docs, not the current docs, gaaah!), so I am reverting my previous change to bring my config back to its original state (when it was first set up). My apologies for the inconvenience.

# Pull request motivation(s)


gladius-mtl:
Again, made a mistake 🙁 
Will create a new PR

PR #2468, opened 2020-10-02
fix(rasa): Force crawling sitemap, bump nbHits

PLNech:
# Pull request motivation(s)
Fixing Rasa's docs crawling.

Crawling `rasa` only results in very few hits.

Getting all matching pages according to the config, most of which are coming from the sitemap.

Bug (Fix)


m-vdb:
thanks @PLNech  🙏

PR #2477, opened 2020-10-05
Update httpie.json to reflect docs URL change

jakubroztocil:
# Pull request motivation(s)

HTTPie documentation has been moved from `httpie.org/doc` to `httpie.org/docs`.  

---

After the URL change, `stop_urls` in the config are ignored, and all docs versions (`https://httpie.org/docs` + `https://httpie.org/docs/*`) are searched simultaneously:

---

We want to search only in the latest stable version of the docs, which is exactly at `https://httpie.org/docs`

### DIFF
--- a/configs/httpie.json
+++ b/configs/httpie.json
@@ -1,20 +1,17 @@
 {
   "index_name": "httpie",
-  "start_urls": [
-    "http://httpie.org/docs"
-  ],
-  "stop_urls": [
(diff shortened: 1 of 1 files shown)


KLMcreator:
The update should only scrape the latest version. Please let me know if there's any issue.

jakubroztocil:
Thanks! 🍰

PR #2495, opened 2020-10-08
Adds the tsconfig reference to the typescript search db

orta:
Hi folks, I'm trying to get the [TSConfig reference](https://www.typescriptlang.org/tsconfig/) added into the search, 

I wondered if I am doing something wrong here by explicitly adding it, but it shows in the sitemap and I made the DOM structure conform to the existing queries, and figure that maybe the page needs to be allow listed this way instead!

### DIFF
--- a/configs/typescriptlang.json
+++ b/configs/typescriptlang.json
@@ -12,6 +12,10 @@
     {
       "url": "https://www.typescriptlang.org/docs/",
       "page_rank": 1
+    },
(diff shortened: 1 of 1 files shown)


Shipow (approved):

Shipow:
Let us know how it goes, we might want to force the sitemap as well. (this merge will trigger a new crawl and index update).

PR #2500, opened 2020-10-09
Bug: dunglas_mercure.json

maxime-urbanski:
"start_urls" not including the specification section of mercure's project . This part is accessible with the url: https://mercure.rocks/spec. At the moment, this section is not accessible for the crawler.


shortcuts:
Updated the **start_urls** with the **/spec** endpoint

PR #2502, opened 2020-10-09
Update nuxtjs_content.json

benjamincanac:
# Pull request motivation(s)

When searching, the lvl0 is always Documentation, that's why the selector is added.

Also, the documents doesn't seem to be indexed with the language.


shortcuts:
Hi, 

Have you tried playing with the [**meta tags**](https://docsearch.algolia.com/docs/required-configuration#introduce-global-information-as-meta-tags) for the language?

Let us know if the problem persists

Atinux (approved):

PR #2628, opened 2020-10-23
Added config file for lucky-byte

TomKlotzPro:
# Pull request motivation(s)


lucky-xiaohu:
hi there,

Today I  search my site for a very common word "星收付“, but  get `No result for "星收付"`，
I  fill the apiKey and indexName to my Docusaurus v2 site 2 days ago，my index name is `lucky-byte`，any suggest?

PR #2654, opened 2020-10-27
Update timescale config

solugebefola:
# Pull request motivation(s)

### DIFF
--- a/configs/timescale.json
+++ b/configs/timescale.json
@@ -1,5 +1,6 @@
 {
   "index_name": "timescale",
+  "sitemap_urls": ["https://docs.timescale.com/docs-sitemap.xml"],
   "start_urls": [
     {
       "url": "https://docs.timescale.com/(?P<version>.*?)/getting-started/installation",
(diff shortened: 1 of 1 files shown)


solugebefola:
@shortcuts not sure why tests failed?

shortcuts:
Hi @solugebefola,

Was the **min_indexed_level** working before? We only accept integers from this key.

So, you should change
https://github.com/algolia/docsearch-configs/blob/e87ab8ff85e10895252bf5bb34963c4432e7bd06/configs/timescale.json#L70-L73

to
```json
"min_indexed_level": 1,
```

solugebefola:
It was working before.  The last time we updated the config was many months ago, so maybe the format changed?  Or the tests were updated?

shortcuts:
@solugebefola It must have been updated.

Looking at your previous records, no errors were detected on your config but it didn't take count of `min_indexed_level` key, from what I'm seeing.

Feel free to @ me if there's any issue with the new configuration, it should be effective in the next 24 hours.

PR #2683, opened 2020-10-31
Move buf index from buf.build/docs to docs.buf.build

bufdev:
We have moved the documentation for buf from https://buf.build/docs to https://docs.buf.build - this updates the configuration as such. Thanks for your continued service here!


bufdev:
Hey, this is still broken for our website - is there some deployment mechanism that needs to happen so that this configuration is live? Thanks for your help.

PR #2698, opened 2020-11-03
Fix configuration

pascalberger:
# Pull request motivation(s)
There's an error in the current configuration for community section.

Results from extensions and community are mixed

Groups should have a unique name

Bug


shortcuts:
@pascalberger Let me know if you need any help/if the problem persists!

PR #2730, opened 2020-11-06
Set js_render to true, add page rank

astuteape:
# Pull request motivation(s)

The content at `sendgrid.api-docs.io` is client-side rendered and is not being indexed by Algolia. I believe the `"js_render": true` is set at the right level, but let me know if this should be defined at the level of each `start_url` or somewhere else in the config.

Our older docs were being returned as results before our more recent docs. Our goal is to have the pages with a lower `page_rank` surfaced last in our search results.

1. Our v2 docs at the `start_url` `twilio.com/docs/sendgrid are shown in our search results before our more current documentation at `sendgrid.com/docs` and `sendgrid.api-docs.io`.
2. Our docs at `sendgrid.api-docs.io` are not being indexed.

1. Search results from `sendgrid.com/docs` and `sendgrid.api-docs.io` will be returned before results from `www.twilio.com/docs/sendgrid/`
2. Our internal search powered by Algolia will display results from https://sendgrid.api-docs.io/v3.0/


shortcuts:
Hi @astuteape,

I've updated your config with a selector for the `v3` which was the reason why it wasn't retrieving any results, it should be working as expected.

Let me know if there's any other issues related to it, the changes should be effective in the next 24 hours.

astuteape:
Hi @shortcuts 

Thank you for the help. I am already seeing the changes, which is good news. However, I think the page rank may still not be what I was after. I'm now seeing both sendgrid.api-docs.io (v3) and twilio.com/docs/sendgrid (v2) show in the results before the content at sendgrid.com/docs. Is there a way I can get the order to be from first results to last results:

1. sendgrid.com/docs
2. sendgrid.api-docs.io
3. twilio.com/docs/sendgrid

Thanks so much for your help and time.

shortcuts:
Hi @astuteape, 

The hits are "weighted" based on the `page_rank` but also a lot of other criteria.

You can see the `weight` of a query by searching it from the dashboard or analyzing the hits returned by the client.

For example with the query: `IP Access Management`
- The API V3 is displayed first with a weight of `90`
- The doc is displayed second with a weight of `80`

Maybe playing around with ranks like setting it at `20` for the doc could help, feel free to submit the PR and I'll merge it asap!

Have a nice day

astuteape:
Thanks @shortcuts 

That's super helpful. I'm going to open another PR with some updated page ranks.

PR #2736, opened 2020-11-09
Fix icon selector

shortcuts:
# Pull request motivation(s)

### DIFF
--- a/configs/bootstrap-icons.json
+++ b/configs/bootstrap-icons.json
@@ -4,28 +4,40 @@
     {
       "url": "https://icons.getbootstrap.com/index.html",
       "selectors_key": "icons-docs",
-      "tags": ["icons-docs"],
+      "tags": [
+        "icons-docs"
+      ],
       "page_rank": 2,
       "extra_attributes": {
-        "language": ["en"]
+        "language": [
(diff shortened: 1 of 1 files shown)


XhmikosR:
I was wondering, what is the point of having getbootstrap.com specific stuff in this config? icons and the main site are different projects?

shortcuts:
This config was only made for the POC (https://github.com/twbs/bootstrap/pull/32098) of having getbootstrap.com doc and the icons under the same search.

XhmikosR:
Oh, OK. Because these are different projects thus it doesn't make sense to keep the same config.

shortcuts:
Sure, we thought that it was a good idea to merge them with the new DocSearch v3, but now that it's clear, I'll separate them and send you the details asap.

XhmikosR:
Thanks :) Let me know the details and I'll try to work on it ASAP.

BTW is v1.0.0 stable enough? Also, do note that for v4, which is what getbootstrap.com currently uses, we support old browsers so we might have compatibility issues with newer JS. For v5, we can make the switch and when we hit stable/beta and move to the main domain, it shouldn't matter anymore :)

PR #2769, opened 2020-11-15
Preact: add language and version facets

developit:
Currently, documentation results for Preact 8 and 10 are intermixed in results - I'm hoping this configuration addresses that, and brings search to the other languages.

I tried to based this off some of the other definitions from this repo, but I don't actually know if it is correct.

Fixes preactjs/preact-www#713.

### DIFF
--- a/configs/buf.json
+++ b/configs/buf.json
@@ -10,7 +10,7 @@
   "stop_urls": [],
   "selectors": {
      "lvl0": {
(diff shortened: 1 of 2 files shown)


shortcuts:
Hi @developit, 

I've updated your config, it's now getting both `v8` and `v10` and retrieving the `lang` from the `html` tag.

You should see the changes in the next 24 hours, feel free to @ me if there's still issues with it!

Have a nice day

PR #2786, opened 2020-11-17
holistics.json: Change start_urls to cover other sections

nvquanghuy:
One issue I face is that the index only crawled in the `/docs/` folder, whereas in our docs, we have 3 other sections that are also part of the document.

* https://docs.holistics.io/guides/
* https://docs.holistics.io/faqs/
* https://docs.holistics.io/api/

This PR changes the start_urls so that it will crawl through all the relevant sub-sections.

### DIFF
--- a/configs/holistics.json
+++ b/configs/holistics.json
@@ -7,7 +7,7 @@
     "https://docs.holistics.io/sitemap.xml"
   ],
   "sitemap_alternate_links": true,
-  "stop_urls": [],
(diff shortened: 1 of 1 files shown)


shortcuts:
It should be all good now! Let me know if there's any issues with this update.

The changes should be effective in the next 24 hours

nvquanghuy:
Awesome thank you for the quick turnaround. Appreciate it!!

PR #2868, opened 2020-11-23
add version support

shortcuts:
# Pull request motivation(s)

### DIFF
--- a/configs/apache_spark.json
+++ b/configs/apache_spark.json
@@ -1,15 +1,15 @@
 {
   "index_name": "apache_spark",
   "start_urls": [
-    "https://spark.apache.org/docs/latest/"
+    {
+      "url": "https://spark.apache.org/docs/(?P<version>.*?)/",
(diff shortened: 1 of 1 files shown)


configs/apache_spark.json
```diff
@@ -1,15 +1,15 @@
...
-  "stop_urls": [
-    "/api",
-    "/index.html"
-  ],
-  "sitemap_urls": [
-    "https://spark.apache.org/sitemap.xml"
+    {
+      "url": "https://spark.apache.org/docs/(?P<version>.*?)/",
+      "variables": {
+        "version": ["latest"]
```
shortcuts, line 7: You can add your versions to this array

shortcuts:
Make sure to update your DocSearch instance with the [**facetFilters**](https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags) to filter the search based on the you are

PR #2878, opened 2020-11-23
create redkube-otomi

Shipow:
HUB generated configuration


shortcuts:
[Duplicated config](https://github.com/algolia/docsearch-configs/blob/master/configs/redkubes_otomi.json)

PR #2907, opened 2020-11-25
add sitemap for https://socket.io

darrachequesne:
# Pull request motivation(s)

Add a sitemap for the Socket.IO website: https://socket.io/sitemap.xml

N/A

N/A

I also increased the `nb_hits` value, as it seems our latest updates were not reflected in the search.

From what I read here: https://docsearch.algolia.com/docs/config-file/#nb_hits-special

> nb_hits is updated automatically each time you run DocSearch on your config.

Is this the expected process?

### DIFF
--- a/configs/socket_io.json
+++ b/configs/socket_io.json
@@ -3,6 +3,9 @@
   "start_urls": [
     "https://socket.io/docs/"
   ],
+  "sitemap_urls": [
+    "https://socket.io/sitemap.xml"
+  ],
   "stop_urls": [],
   "selectors": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @darrachequesne,

Hits are updated automatically on your indices but you can't see the changes on the files, only on the dashboard.

Could you please confirm that you also asked an access for analytics for this index? 

Thanks!

darrachequesne:
Thanks!

> Could you please confirm that you also asked an access for analytics for this index?

Yes indeed.

PR #2908, opened 2020-11-25
Update ory search config and problems with search index and empty results

aeneasr:
# Pull request motivation(s)

We are currently encountering several issues related to search when using docusaurus `2.0.0-alpha.69`.

**Search result previews are empty / `null`**

Bildschirmfoto 2020-11-25 um 15 24 59

This result happens without facet filters. For example, search for `login` on https://www.ory.sh/hydra/docs/5min-tutorial/. We have facet filters disabled on those subpages currently (due to another bug.. ;) )

**FacetFilters not working**

When adding facetFilters to the queries, the search results are simply empty. I checked the meta tags and they seem to be included so I am not sure what is happening. You can check this on: https://www.ory.sh/oathkeeper/docs/ (search for e.g. "access rule").

Here are the facetFilters we are using: `["language:en",["docusaurus_tag:default","docusaurus_tag:docs-default-v0.38"],["tags:oathkeeper","tags:docs"]]`

I believe the problem is the `docusaurus_tag` facets but they seem to be rendered correctly when I check the website's source.

See above

We would expect snippets to show previews, and `docusaurus_tag` filters not to cause issues.

We hope that these changes to the index fix the problems mentioned but it is a lot of trial and error figuring out what's going on.

It would also be cool if there was some way of debugging this stuff faster and more efficient :)

### DIFF
--- a/configs/ory.json
+++ b/configs/ory.json
@@ -4,13 +4,7 @@
     {
       "url": "https://www.ory.sh/docs",
       "tags": [
-        "cloud"
-      ]
-    },
-    {
-      "url": "https://www.ory.sh/docs/ecosystem",
-      "tags": [
-        "ecosystem"
+        "docs"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi!

Your new config should cover bugs you had, feel free to @ me if you need help with an update.

For now, you can use [DocSearch Scraper](https://github.com/algolia/docsearch-scraper) to run your config [on your own](https://docsearch.algolia.com/docs/run-your-own/) for debugging.

aeneasr:
Nice, indeed it did! Thank you!

PR #2911, opened 2020-11-26
Update yiddishe-kop config

Yiddishe-Kop:
### Update Docsearch config

The docsearch on https://laravel-commerce.yiddishe-kop.com/ is not returning any results.
I hope this fixes it.

Thanks.

### DIFF
--- a/configs/yiddishe-kop.json
+++ b/configs/yiddishe-kop.json
@@ -13,7 +13,7 @@
     "lvl2": "div.flex div article.prose h2",
     "lvl3": "div.flex div article.prose h3",
     "lvl4": "div.flex div article.prose h4",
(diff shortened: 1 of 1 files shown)


Yiddishe-Kop:
Why is the CI saying that it's `missing nb_hits key`, when it's not missing?

shortcuts:
> Why is the CI saying that it's `missing nb_hits key`, when it's not missing?

Hi, there was a trailing comma on the last object which was not supposed to be here

Yiddishe-Kop:
How long does it take until the new config goes into effect?

shortcuts:
It should be effective in the next 24 hours max!

PR #2951, opened 2020-12-01
Add docsearch configs for Diem developer sites

JoelMarcey:
# Pull request motivation(s)

Add configs for the new Diem developer sites

No search results

Search results on the sites

### DIFF
--- a/configs/diem_developer_website.json
+++ b/configs/diem_developer_website.json
@@ -4,14 +4,12 @@
     "https://developers.diem.com/",
     "https://developers.diem.com/docs/welcome-to-diem"
   ],
-  "sitemap_urls": [
-    "https://developers.diem.com/sitemap.xml"
-  ],
+  "sitemap_urls": ["https://developers.diem.com/sitemap.xml"],
   "sitemap_alternate_links": true,
   "stop_urls": [],
   "selectors": {
     "lvl0": {
-      "selector": ".menu__link--sublist.menu__link--active",
+      "selector": ".menu__link[class*='active_']",
       "global": true,
       "default_value": "Documentation"
     },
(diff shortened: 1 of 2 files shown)


configs/diem_developer_website.json
```diff
@@ -0,0 +1,32 @@
...
+    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li"
+  },
+  "selectors_exclude": [
+    ".hash-link"
+  ],
+  "conversation_id": [
+    "1176378229"
+  ],
+  "nb_hits": 989
+}
```
JoelMarcey, line 32: I use the same values here as another config - https://github.com/algolia/docsearch-configs/blob/master/configs/libra-website.json

I hope that is ok?

configs/diem_improvement_proposals.json
```diff
@@ -0,0 +1,28 @@
...
+    "text": "article p, article li"
+  },
+  "selectors_exclude": [
+    ".hash-link"
+  ],
+  "conversation_id": [
+    "1129563849"
+  ],
+  "nb_hits": 187
+}
```
JoelMarcey, line 28: I use the same values here as another config - https://github.com/algolia/docsearch-configs/blob/master/configs/libra-lips.json

I hope that is ok?

configs/diem_developer_website.json
shortcuts: All good, I updated it to make sure it match your website structure.

configs/diem_improvement_proposals.json
shortcuts: All good, I updated it to make sure it match your website structure.

shortcuts:
Hi @JoelMarcey,

Thanks for the PR.

I've updated the `lvl0` selector and will send you the credentials once deployed!

Feel free to @ me if there's any issue related to it

JoelMarcey:
@shortcuts Thank you so much for correcting and merging this. I really appreciate it.

Do you have an idea of when we should start seeing results on the site -- for example results from the search box on https://developers.diem.com ?

shortcuts:
@JoelMarcey Crawl task has been started and should already been effective or will be in the next 24 hours max!

I'm sending your credentials (apiKey/indexName) to your mail address so make sure to also update it on your side and also solve your ticket regarding Libra account access

PR #2952, opened 2020-12-01
fix wpgraphql missing results

shortcuts:
# Pull request motivation(s)


jasonbahl:
@shortcuts it looks like more content is showing in search now, but `functions` appear to still be left out. 

For example, searching for `register_graphql_field` is not showing in search even though `/functions/` included in the config and the pages have a `.content` class in the markup. See: https://www.wpgraphql.com/functions/register_graphql_field/

shortcuts:
@jasonbahl I'm on it and will tag you on the new PR once it's done

PR #2973, opened 2020-12-03
[wpgraphql] fix js rendering

shortcuts:
# Pull request motivation(s)


shortcuts:
@jasonbahl it looks like there was some js rendering and it wasn't specified in the config.

It should now be all good, check the `nb_hits` increase :D

PR #2980, opened 2020-12-03
[Issue 2978] Fix search for Apache Pulsar website

### LINKED ISSUE #2978: [Apache Pulsar] Search stopping work for Pulsar 2.6.2 and 2.7.0
http://pulsar.apache.org/
The search results of 2.6.2 and 2.7.0 are empty. how should this be handled and who can help.

Jennifer88huang:
Fixes #2978 
# Pull request motivation(s)
1. The search does not work for Apache Pulsar doc 2.6.2 and 2.7.0 versions on https://pulsar.apache.org/docs/en/standalone/ 

Nothing comes out on the research results.
image

It should be sth like this.
image

1. Do we need to add the version number here for each release?
2. Do we need additional configuration, so it would work better?
3. The search function does not work properly for some key words, how to solve it?

### DIFF
--- a/configs/apache_pulsar.json
+++ b/configs/apache_pulsar.json
@@ -5,8 +5,10 @@
       "url": "https://pulsar.apache.org/docs/(?P<language>.*?)/(?P<version>.*?)/",
       "variables": {
         "version": [
-          "2.6.1",
+          "2.7.0",
           "next",
+          "2.6.2",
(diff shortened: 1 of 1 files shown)


shortcuts:
Thanks for the PR.

Feel free to @ me if there's still an issue with it, you should be able to see the changes in the next 24 hours.

PR #2984, opened 2020-12-03
config: fix Apache Camel documentation levels

zregvart:
We changed the markup of the documentation, breadcrumbs element is no longer a `div` but rather a `nav`. To make the solution more portable with respect to these kinds of changes selector is changed to target by class name rather than by element.

# Pull request motivation(s)

We're receiving `null` values for `lvl0`...`lvl3`.

We should be getting `lvl0`...`lvl3` values corresponding to the breadcrumbs element.

This is a bug, search results are displayed as having _NULL_ headings.

I've contacted support, as I wasn't sure this was an issue we introduced by changing the markup, this should fix it.

### DIFF
--- a/configs/apache_camel.json
+++ b/configs/apache_camel.json
@@ -55,10 +55,10 @@
       "text": "article p, article li"
     },
     "antora-pages": {
-      "lvl0": "div.toolbar ul li:nth-child(1) a",
-      "lvl1": "div.toolbar ul li:nth-child(2) a",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @zregvart 

It looks like this ressource is not public, is that a normal behavior?
https://camel.apache.org/categories/

zregvart:
Thanks! Works wonderfully now!

PR #3005, opened 2020-12-08
[Issue 2978] Fix search for Apache Pulsar website

Jennifer88huang:
# Pull request motivation(s)
The search still does not work for Pulsar 2.7.0.

image

Search should work well for 2.7.0 release.


shortcuts:
Hi,

Thanks for the PR.

The crawl tasks have been executed without any issues on our side.

It looks like your **latest stable version** (2.7.0 in this case) doesn't exists at the URL pattern defined in your config (e.g.: https://pulsar.apache.org/docs/en/2.7.0/) but https://pulsar.apache.org/docs/en/.

I'll do some tests and update it once it's working!

shortcuts:
@Jennifer88huang after a few tests, it looks like the `version` variable is retrieved without it specifying it and still works as an `attributeForFaceting`

You should be able to see the changes in the next 24 hours.

Hope this solves your issue,
Have a nice day

Jennifer88huang:
@shortcuts Thank you very much for your test and detailed explanation.
Now since you remove the versions, so we do not need to update the config file when we have more releases later. Is it correct?

shortcuts:
You shouldn't (waiting for the crawl to be effective to confirm) need to update it anymore but you can still use them as filters!

Jennifer88huang:
@shortcuts That's great. Will check it after 24 hours. 
Thank you very much for your support and help, and have a nice day~

PR #3012, opened 2020-12-09
Update and rename payid.json to paystring.json

Stormtv:
PayID has been renamed to PayString

# Pull request motivation(s)
We have changed our branding from PayID to PayString and thus the URLs have changed also

It is updating based on the old sitemap.

Update based on new sitemap

No

No


Stormtv:
The CI failed when I tried to rename the JSON file to paystring.json. I'm unsure of the proper process of changing the file name or if just updating the contents is sufficient.

shortcuts:
Hi,

Thanks for the PR.

If you would like to change your index name, please send us an email with the email that **requested DocSearch** so we can process.

Have a nice day

PR #3046, opened 2020-12-14
Update jest-v2.json: add missing docusaurus_tag

slorber:
cc @Shipow

### DIFF
--- a/configs/jest-v2.json
+++ b/configs/jest-v2.json
@@ -28,7 +28,8 @@
     "attributesForFaceting": [
       "language",
       "version",
-      "type"
+      "type",
+      "docusaurus_tag"
     ],
     "attributesToRetrieve": [
       "hierarchy",


shortcuts:
Hey @slorber,

I just realized that the `lvl0` selector might be wrong, we use a dynamic one for Docusaurus v2, should I add it?

slorber:
hey @shortcuts 

pinged @Shipow about this on discord, but yeah you can update the config

was just wondering how we ended up with a non up to date docusaurus v2 config, it probably hides another issue in the process (like copying a file from an outdated docusaurus template or something)

shortcuts:
It should be fixed in https://github.com/algolia/docsearch-configs/pull/3048, feel free to @ me if there's any other issue

I'll try to find out why it was generated with the Docusaurus v1 template!

PR #3062, opened 2020-12-17
chore: update tink config

rubda:
# Pull request motivation(s)

We've found that this configuration suits our needs better. Issues with the current config: 

- It doesn't scrape our glossary: https://docs.tink.com/glossary
- It doesn't scrape our changelog: https://docs.tink.com/changelog
- We're seeing repetitive results

Screenshot 2020-12-17 at 11 46 33


shortcuts:
Hi @rubda

Thanks for the PR, feel free to @ me if you still have issue with your config

Issue #3121, opened 2020-12-24
[avo] sitemap is not recognized


tpom6oh:
# Do you want to request a **feature** or report a **bug**?

Probably a bug

### If it is a DocSearch index issue, what is the related `index_name` ?

`index_name`= avo

## What is the **current** behaviour?

Here is our config

https://github.com/algolia/docsearch-configs/blob/master/configs/avo.json

I've run the scrapper locally against it and it only picks up the start url and another url that is present on that page 

```
docsearch-scraper % ./docsearch run avo-config.json          
> DocSearch: https://www.avo.app/docs/ 14 records)
> DocSearch: https://www.avo.app/docs/help/troubleshooting 5 records)

Nb hits: 19
```

## What is the **expected** behaviour?

All the urls from the specified sitemap - https://www.avo.app/docs/sitemap.xml - should be crawled 

#### What have you tried to solve it?

Read the docs and it looks like everything is set up correctly to me 🤔

I've tried to remove the start urls configuration and after that change the local crawler actually checked everything in the sitemap.

The pull request is #3122

shortcuts:
Resolved in https://github.com/algolia/docsearch-configs/pull/3122

PR #3122, opened 2020-12-24
Update avo.json

tpom6oh:
Related to #3121 

I've tried removing `start_urls` and it picks up the sitemap when I run the crawler locally.

### DIFF
--- a/configs/avo.json
+++ b/configs/avo.json
@@ -1,5 +1,6 @@
 {
   "index_name": "avo",
+  "start_urls": [],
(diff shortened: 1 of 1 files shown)


shortcuts:
> Related to #3121 
> 
> 
> 
> I've tried removing `start_urls` and it picks up the sitemap when I run the crawler locally.

As the URLs in the sitemap don't follow the same schema as the one in the start_urls (missing `www`), it looks like it doesn't follow it

PR #3127, opened 2020-12-28
Added 3.2 version for ezplatform

DominikaK:
Added new 3.3 version and removed unmaintained versions for doc.ibexa.co

Also removed configuration for an unmaintained doc project ezplatform-ezpersonalization


DominikaK:
@shortcuts, can you please advise whether this is the wrong way to remove the unused config file? It seems to be failing tests. If so, I'll remove the commit and update versions only.

shortcuts:
@DominikaK Hi! Could you please move the disabled config file inside the `disabled` folder? I'll merge it right after

Thanks!

DominikaK:
Thank you, done.

PR #3177, opened 2021-01-05
update standardnotes to match docusaurus-2

TheodoreChu:
# Pull request motivation(s)

I would like to update [`standardnotes.json`](https://github.com/algolia/docsearch-configs/blob/master/configs/standardnotes.json) to match [`docusaurus-2.json`](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json) to keep up with the latest developments of Docusaurus 2 as discussed in https://github.com/facebook/docusaurus/issues/3996.

The current behavior does not produce results on [standardnotes.org/search](https://docs.standardnotes.org/search/?q=self) even though results appear in the popup that results from searching in the navbar:

image

image

The expected behavior is to have results appear on the /search page as on [Docusaurus 2's website](https://v2.docusaurus.io/search/?q=self).

### DIFF
--- a/configs/standardnotes.json
+++ b/configs/standardnotes.json
@@ -1,31 +1,37 @@
 {
   "index_name": "standardnotes",
-  "start_urls": [
-    "https://docs.standardnotes.org/"
-  ],
-  "sitemap_urls": [
-    "https://docs.standardnotes.org/sitemap.xml"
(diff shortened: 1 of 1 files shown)


TheodoreChu:
Thanks!

PR #3180, opened 2021-01-06
Update patternfly.json

redallen:
Include /html pages in search results.

# Pull request motivation(s)


shortcuts:
Oops, I thought I had it removed!

Good catch.

PR #3219, opened 2021-01-11
Update configcat.json

laliconfigcat:
# Pull request motivation(s)
We'd like to index our blog site too next to our docs.

Only the docs site is indexed.

We'd like both our docs and blog sites to be indexed.

### DIFF
--- a/configs/configcat.json
+++ b/configs/configcat.json
@@ -5,7 +5,8 @@
     { "url": "https://configcat.com/blog/", "tags": [ "blog" ] }
   ],
   "sitemap_urls": [
-    "https://configcat.com/docs/sitemap.xml"
+    "https://configcat.com/docs/sitemap.xml",
+    "https://configcat.com/blog/sitemap.xml"
   ],
   "sitemap_alternate_links": true,
   "stop_urls": [],
--- a/configs/configcat.json
+++ b/configs/configcat.json
@@ -1,7 +1,8 @@
 {
   "index_name": "configcat",
   "start_urls": [
-    "https://configcat.com/docs/"
+    { "url": "https://configcat.com/docs/", "selectors_key": "docsselector", "tags": [ "docs" ] }, 
+    { "url": "https://configcat.com/blog/", "selectors_key": "blogselector", "tags": [ "blog" ] },
   ],
   "sitemap_urls": [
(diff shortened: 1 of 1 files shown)


configs/configcat.json
```diff
@@ -1,7 +1,8 @@
 {
   "index_name": "configcat",
   "start_urls": [
-    "https://configcat.com/docs/"
+    { "url": "https://configcat.com/docs/", "selectors_key": "docsselector", "tags": [ "docs" ] }, 
+    { "url": "https://configcat.com/blog/", "selectors_key": "blogselector", "tags": [ "blog" ] },
```
shortcuts, line 5:
```suggestion
    { "url": "https://configcat.com/docs/", "tags": [ "docs" ] }, 
    { "url": "https://configcat.com/blog/", "tags": [ "blog" ] }
```

If they both have the same structure (same selectors), they don't need a `selectors_key`. But if they are not similarly structured, you should see [this part of the doc](https://docsearch.algolia.com/docs/config-file/#using-custom-selectors-per-page)

configs/configcat.json
```diff
@@ -1,7 +1,8 @@
...
   ],
   "sitemap_urls": [
     "https://configcat.com/docs/sitemap.xml"
```
shortcuts, line 8:
```suggestion
    "https://configcat.com/docs/sitemap.xml",
    "https://configcat.com/blog/sitemap.xml"
```

shortcuts:
In case you want to filter your search based on the page you are on (search docs on /docs, search blog posts on /blog), you can add `"tags"` to [the attributeForFaceting array]( https://github.com/algolia/docsearch-configs/blob/e95960ee53c40fce64aaca3b485dc425d04d9c22/configs/configcat.json#L30-L35)

laliconfigcat:
> In case you want to filter your search based on the page you are on (search docs on /docs, search blog posts on /blog), you can add `"tags"` to [the attributeForFaceting array](https://github.com/algolia/docsearch-configs/blob/e95960ee53c40fce64aaca3b485dc425d04d9c22/configs/configcat.json#L30-L35)

Hey,

Thanks for the review, the fixes and the faceting tip! We'd like our users to be able to search for everything from both sites so we are good for now.

PR #3241, opened 2021-01-15
Update URLs and selectors

jhodges10:
# Pull request motivation(s)
I noticed that our last update still wasn't really picking up all the things we needed it to, so I took another pass at our `config.json` and tested it locally with the dockerized crawler and a non-DocSearch index to confirm the new settings work correctly.

We will likely have another update or two in a week or two once we've completed the next micro-site for this project.

### DIFF
--- a/configs/frame.json
+++ b/configs/frame.json
@@ -19,7 +19,7 @@
     ],
     "stop_urls": [],
     "selectors": {
-        "marketing": {
+        "portal": {
             "lvl0": ".header-wrapper h1",
             "lvl1": ".content-body h2",
             "lvl2": ".content-body h3",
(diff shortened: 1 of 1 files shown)


configs/frame.json
```diff
@@ -1,22 +1,55 @@
...
+            "url": "https://developer.frame.io/portal",
+            "selectors_key": "portal"
+        }
+    ],
+    "selectors_exclude": [
+        "aside"
+    ],
+    "stop_urls": [],
+    "selectors": {
+        "marketing": {
```
shortcuts, line 22:
```suggestion
        "portal": {
```

jhodges10: Good catch!

PR #3345, opened 2021-01-27
Update tixstock.json

wkdcode-ash:
# Pull request motivation(s)
We have 2 environments using the same API key requiring 2 searches that are split out via Tags, we reached out to get a second API key however this would be the better approach.

Currently we only have the sandbox env working with the doc search.

We're now in a position to set the prod env with doc search due to the API going live. we initially requested a second API key however upon speaking with Clement we have been asked to take this approach to solve our issue.

### DIFF
--- a/configs/tixstock.json
+++ b/configs/tixstock.json
@@ -1,26 +1,45 @@
 {
-  "index_name": "tixstock",
-  "start_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi,

This looks all good! 

Let me know if there's any issue, you should see the changes in the next 24 hours

PR #3357, opened 2021-01-30
Update Bulma config to match new website markup

jgthms:
# Pull request motivation(s)

I've recently updated the Bulma website with new markup: https://bulma.io/
This PR updates the config to match this new markup.

The config doesn't match the production website markup.

For the config to match the production website markup.


jgthms:
Thanks for the quick turnaround.

PR #3394, opened 2021-02-03
Update pubnub.json

crimsonred:
# Pull request motivation(s)

We updated our docs URLs and added a new sitemap. So the docs search is not crawling all the pages.

The update also changed the search selectors. With the changes we should be able to search the missing pages.

### DIFF
--- a/configs/pubnub.json
+++ b/configs/pubnub.json
@@ -184,7 +184,7 @@
       "lvl4": ".content .content h4",
       "lvl5": ".content .content h5",
       "lvl6": ".content .content h6",
-      "text": ".wrapper .post p, .wrapper .post li"
+      "text": ".wrapper .post p, .wrapper .post li",
       "sdk": {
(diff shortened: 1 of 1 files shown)


configs/pubnub.json
```diff
@@ -167,51 +167,33 @@
...
-      "lvl5": ".content .content h6",
-      "text": ".content .content p, .content .content li,  .content .content td:last-child",
+      "lvl0": "h1.postHeaderTitle",
+      "lvl1": ".wrapper .post p",
+      "lvl2": ".content .content h2",
+      "lvl3": ".content .content h3",
+      "lvl4": ".content .content h4",
+      "lvl5": ".content .content h5",
+      "lvl6": ".content .content h6",
+      "text": ".wrapper .post p, .wrapper .post li"
```
shortcuts, line 187:
```suggestion
      "text": ".wrapper .post p, .wrapper .post li",
```

PR #3420, opened 2021-02-04
Update index_name and start_urls for notion-formula doc

Devorein:
# Pull request motivation(s)
The documentation site has been transferred to a new URL, that's the reason why the start_url needs to be changed.

The crawler will fail to find a sitemap since the current documentation site will be removed soon.

### DIFF
--- a/configs/notion-formula.json
+++ b/configs/notion-formula.json
@@ -1,7 +1,7 @@
 {
-  "index_name": "notion-formula",
+  "index_name": "nishan",
   "start_urls": [
-    "https://nishans-notion-formula.netlify.app/docs/"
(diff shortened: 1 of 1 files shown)


configs/notion-formula.json
```diff
@@ -1,7 +1,7 @@
 {
-  "index_name": "notion-formula",
+  "index_name": "nishan",
```
shortcuts, line 2: Hi,

If you'd like to change the index name, please send us an email directly, as it requires a few changes. 

In the meantime, you should revert it. 

```suggestion
  "index_name": "notion-formula",
```

shortcuts (changes requested):

PR #3428, opened 2021-02-06
Update photo-sphere-viewer.json

mistic100:
# Pull request motivation(s)

Make all the results from the "API" visible under a single top-level section by using a fake lvl0 declaration and moving all selectors one level down.

### DIFF
--- a/configs/photo-sphere-viewer.json
+++ b/configs/photo-sphere-viewer.json
@@ -34,7 +34,7 @@
     },
     "api": {
       "lvl0": {
-        "selector": "#noop",
+        "selector": "",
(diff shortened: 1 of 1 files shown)


configs/photo-sphere-viewer.json
```diff
@@ -34,12 +34,13 @@
     },
     "api": {
       "lvl0": {
-        "selector": ".main header h2",
+        "selector": "#noop",
```
shortcuts, line 37:
```suggestion
        "selector": "",
```

This is how we usually do

mistic100:
Thank you

PR #3466, opened 2021-02-08
Update wandisco-live-data-migrator.json

benradstone:
Incorrect wording for "tags".

# Pull request motivation(s)


shortcuts:
good catch!

benradstone:
@shortcuts yes, apologies :)

PR #3474, opened 2021-02-09
Revert #3464.

mattstein:
Sorry for the noise—this was a bad idea and made things worse. 🤦‍♂️

### DIFF
--- a/configs/craftcms_commerce.json
+++ b/configs/craftcms_commerce.json
@@ -37,7 +37,7 @@
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": ".theme-default-content .router-link-exact-active",
+    "lvl1": ".theme-default-content h1",
     "lvl2": ".theme-default-content h2",
     "lvl3": ".theme-default-content h3",
     "lvl4": ".theme-default-content h4",


shortcuts:
Hi, 

Could you please give me a detailed example of what should be accomplished?

Thanks

mattstein:
Hi @shortcuts, thanks for merging! I noted what I was attempting to solve in this issue:
https://github.com/pixelandtonic/vuepress-theme-craftdocs/issues/22

I’m not sure if it’s more an issue of UI or indexing, and reverting this change doesn’t solve it but gets us back to a preferable form of problem until I can spend more time on it.

shortcuts:
Hi!

So, for now having `.sidebar-heading.open` as a `lvl1` selector could help but it will display something like `craft\commerce\elements`. 

Best solution would be to have a special selector around the last part (id, class, etc.)

mattstein:
Thanks for the quick response, @shortcuts!

That `.theme-default-content .router-link-exact-active` selector worked perfectly as a `lvl1` selector in terms of what I intended to index—it’s the VuePress DocSearch UI that became uncomfortably cluttered and harder to read.

PR #3477, opened 2021-02-10
update URL for git-gosling project

HonkingGoose:
# Pull request motivation(s)

I've migrated from Netlify hosting to Vercel hosting.
The URL where you can find my project has changed.
This means that the Algolia crawler should stop indexing the Netlify site, and instead index the Vercel site.

I've also transferred the repository under my own namespace (from `RoostingGeese/git-gosling` to `HonkingGoose/git-gosling`).
I don't know if this affect my eligibility for the Algolia DocSearch for Open Source projects, I'm not changing any license/content stuff.

I'm still using Docusaurus v2 on the Vercel hosted URL, so the build system is not changing.

The crawler indexes the Netlify site once every 24 hours.

The Algolia crawler should index the Vercel site instead.

I hope I'm doing this right.
Is there anything else I should know/do to complete the migration?
Can I  keep using the Algolia `APIKey` and `indexName` that were provided when I was using Netlify, or do I need to request new ones?

### DIFF
--- a/configs/git-gosling.json
+++ b/configs/git-gosling.json
@@ -1,10 +1,10 @@
 {
   "index_name": "git-gosling",
   "start_urls": [
-    "https://git-gosling.netlify.app/docs/"
+    "https://git-gosling.vercel.app/docs/"
   ],
   "sitemap_urls": [
-    "https://git-gosling.netlify.app/sitemap.xml"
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi, you can use the same credentials!

HonkingGoose:
@shortcuts That's very handy, thank you for merging! 😄

PR #3526, opened 2021-02-15
Update baslerweb_ja.json to add CSS selector

wilhelmer:
# Pull request motivation(s)

CSS class "order-number" not included in search index (Japanese version).

Added .order-number to "text" selector (Japanese version).

### DIFF
--- a/configs/baslerweb_ja.json
+++ b/configs/baslerweb_ja.json
@@ -14,7 +14,7 @@
     "lvl4": ".md-content h4",
     "lvl5": ".md-content h5",
     "lvl6": ".md-content h6",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi,

We could merge both configs together to support all the domains inside the same one/credentials which will leverage the maintenance, if you are interested let me know so I can proceed.

On your side, you will have to filter your search using [custom tags](https://docsearch.algolia.com/docs/config-file/#using-custom-tags)

wilhelmer:
Hi @shortcuts, good to know. But for now, please keep the configs separate, I don't expect many changes this year.

PR #3529, opened 2021-02-15
vuetifyjs: update selectors

KaelWD:
# Pull request motivation(s)
Adding separate selectors for API pages as they weren't being matched properly

> Please attach also a URL showing that you are a maintainer of the project.

https://vuetifyjs.com/en/about/meet-the-team/#core-team

### DIFF
--- a/configs/vuetifyjs.json
+++ b/configs/vuetifyjs.json
@@ -22,21 +22,29 @@
   "selectors": {
     "default": {
       "lvl0": {
+        "selector": "#default-drawer .v-list > .v-list-group--active > .v-list-group__header",
+        "global": true
+      },
+      "lvl1": {
         "selector": ".markdown-body h1",
         "global": true
       },
-      "lvl1": ".markdown-body h2",
-      "lvl2": ".markdown-body h3",
-      "lvl3": ".markdown-body h4",
+      "lvl2": ".markdown-body h2",
+      "lvl3": ".markdown-body h3",
+      "lvl4": ".markdown-body h4",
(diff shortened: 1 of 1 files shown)


KaelWD:
The documentation says:

> If you're using a multi-level navigation, we recommend that you mark each active level with a custom CSS class. This will make it easier for DocSearch to know where the current page fits in the website hierarchy.

Would that look like this?

```
"lvl0": {
  "selector": ".v-list-group__header.v-list-item--active",
  "global": true
},
"lvl1": {
  "selector": ".markdown-body h1, .v-list-item--active:not(.v-list-group__header)",
  "global": true
},
"lvl2": ".markdown-body h2",
"lvl3": ".markdown-body h3",
"lvl4": ".markdown-body h4",
"text": ".markdown-body p, .markdown-body li"
```

shortcuts:
Hi @KaelWD,

> Would that look like this?

- This `lvl0` selector won't work in your case as multiple drawers/sub drawers can be active which could lead in duplicates/inaccurate results.
- As this `lvl1` is the same as the `.markdown-body h1` it might be easier to keep it this way.

For some reason, I can't push to your PR but [here's an updated](https://gist.github.com/shortcuts/42c3efe2d693ebdc162df9a23c3b1967) of your config using xPath, without adding new classes. I've tested it on a few searches and it works perfectly!

KaelWD:
OK thanks. I'm not really familiar with XPath but the CSS selector `.v-list > .v-list-group--active > .v-list-group__header` should do the same thing.

PR #3535, opened 2021-02-16
Update numl.json "start_urls"

shubham-kaushal:
# Pull request motivation(s)

*Only one URL was available for crawling*

N/A

### DIFF
--- a/configs/numl.json
+++ b/configs/numl.json
@@ -1,10 +1,15 @@
 {
   "index_name": "numl",
   "start_urls": [
-    "https://numl.design/handbook/"
+    "https://numl.design/",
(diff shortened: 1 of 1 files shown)


configs/numl.json
```diff
@@ -1,10 +1,15 @@
 {
   "index_name": "numl",
   "start_urls": [
-    "https://numl.design/handbook/"
+    "https://numl.design/",
+    "https://numl.design/storybook",
+    "https://numl.design/handbook/",
+    "https://numl.design/reference/",
+    "https://numl.design/repl",
```
shubham-kaushal, line 8: Replaced the duplicate URL with numl.design/repl

PR #3578, opened 2021-02-17
[SAP Cloud SDK]: Enable indexing of the <code> tag to highlight the terms in documentation (not related to code blocks)

artemkovalyov:
- The `<code>` tag in Docusaurus is used for inline code snippets and we use it to highlight the terms in our docs.
- We want to see it in the index as those are often searched for
- Larger example code snippets are excluded from the search index

# Pull request motivation(s)

`<code>` tags are not indexed on https://sap.github.io/cloud-sdk/docs/overview/overview-cloud-sdk

The` <code>` tag is used for the inline code highlight which is mostly leveraged for tech terms frequently searched by the customers. This doesn't break the best practice to not index example code snippets as they are not marked with the `<code>` tag but are represented by `<div>` with a dedicated CSS class.

This is a config update

Thanks for the great tool and wonderful support for OS projects.

### DIFF
--- a/configs/sap_cloud-sdk.json
+++ b/configs/sap_cloud-sdk.json
@@ -20,7 +20,7 @@
     "lvl3": "[class^='docItemContainer_'] h3",
     "lvl4": "[class^='docItemContainer_'] h4",
     "lvl5": "[class^='docItemContainer_'] h5",
-    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li"
+    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li, code"
   },
   "selectors_exclude": [
     ".hash-link"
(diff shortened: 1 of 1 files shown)


configs/sap_cloud-sdk.json
```diff
@@ -20,7 +20,7 @@
     "lvl3": "[class^='docItemContainer_'] h3",
     "lvl4": "[class^='docItemContainer_'] h4",
     "lvl5": "[class^='docItemContainer_'] h5",
-    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li"
+    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li, code"
```
artemkovalyov, line 23:
```suggestion
    "text": "[class^='docItemContainer_'] p, [class^='docItemContainer_'] li, [class^='docItemContainer_'] code"
```

PR #3597, opened 2021-02-21
update config, cause of config is not take effect.

chudongvip:
# Pull request motivation(s)


configs/anyrtc.json
```diff
@@ -8,32 +8,21 @@
...
-    "attributesToRetrieve": [
-      "hierarchy",
-      "content",
-      "anchor",
-      "url",
-      "url_without_anchor",
-      "type"
-    ]
-  },
+  "custom_settings": {},
```
shortcuts, line 25:
```suggestion
```

This line can be removed if it's empty

PR #3600, opened 2021-02-22
fix config for dr-dk

krep-dr:
# Pull request motivation(s)
Fix config for scraper

The start url is not correct

### DIFF
--- a/configs/dr-dk.json
+++ b/configs/dr-dk.json
@@ -1,16 +1,40 @@
 {
-  "index_name": "dr-dk",
-  "start_urls": ["https://developer.dr.dk/"],
+  "index_name": "dr-pipeline",
+  "start_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi there,

Is that a private URL?

It seems unreachable/doesn't redirect

krep-dr:
yes, the site is behind an ip filter. But we have whitelisted the ip from here https://www.algolia.com/doc/tools/crawler/troubleshooting/faq/#what-ip-address-can-i-use-for-ip-whitelisting

shortcuts:
@krep-dr Sorry but this is related to the Crawler (https://www.algolia.com/products/crawler/). With DocSearch, your website have to be completely accessible and free to use to meet our policy. https://docsearch.algolia.com/docs/who-can-apply#the-checklist

krep-dr:
ah, sorry I missed that part

PR #3607, opened 2021-02-23
Update vue_test_utils.json

afontcu:
I've added valid content on `/guide` (https://next.vue-test-utils.vuejs.org/guide/), so now tagging `guide` should work :)

### DIFF
--- a/configs/vue_test_utils.json
+++ b/configs/vue_test_utils.json
@@ -12,6 +12,11 @@
       "tags": ["current", "api"]
     },
     {
+      "url": "https://vue-test-utils.vuejs.org/guides",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @afontcu 

I've also removed the `guide` part for the `current` version. Should I add it back?

afontcu:
> Hi @afontcu
> 
> I've also removed the `guide` part for the `current` version. Should I add it back?

Sure, added back! Path is `/guides` and it was already indexed, but I assume it would not be displayed on a separate block from API results

shortcuts:
@afontcu This should be good to go, feel free to @ me if there's any issue!

PR #3629, opened 2021-02-24
add config file for injective

shortcuts:
# Pull request motivation(s)


shortcuts:
@msied Update is live and everything should work as expected

PR #3671, opened 2021-03-02
Update seatsio.json

bverbeken:
added https://docs.seats.io/docs/ to start_urls

# Pull request motivation(s)
We're in the process of moving our old documentation, currently available on https://docs.seats.io, to docusaurus. The new docs site is currently live on https://seatsio.github.io/docs/. 

### DIFF
--- a/configs/seatsio.json
+++ b/configs/seatsio.json
@@ -2,7 +2,7 @@
   "index_name": "seatsio",
   "start_urls": [
     "https://seatsio.github.io/docs/",
-    "https://docs.seats.io/docs/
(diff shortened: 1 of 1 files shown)


configs/seatsio.json
```diff
@@ -1,7 +1,8 @@
 {
   "index_name": "seatsio",
   "start_urls": [
-    "https://seatsio.github.io/docs/"
+    "https://seatsio.github.io/docs/",
+    "https://docs.seats.io/docs/
```
shortcuts, line 5:
```suggestion
    "https://docs.seats.io/docs/"
```

PR #3677, opened 2021-03-02
Update wandisco-live-data-azure.json

benradstone:
# Pull request motivation(s)


benradstone:
All checks passed, this is to get our config inline with our staging/test area.

PR #3761, opened 2021-03-09
Updated configuration

petrdvorak:
We are releasing a new version of the dev portal and we need to adjust the search layout accordingly


configs/wultra_developers.json
```diff
@@ -8,26 +8,43 @@
...
+    "lvl5": "section h4",
     "text": "section p, section li"
   },
   "selectors_exclude": [
     ".sidebar-toc",
     ".sidebar"
   ],
-  "conversation_id": [
-    "1176317311"
-  ],
```
shortcuts, line 25: This is a mandatory field, could you please revert it?

configs/wultra_developers.json
```diff
@@ -8,26 +8,43 @@
...
+      "hierarchy",
+      "content",
+      "anchor",
+      "url",
+      "url_without_anchor",
+      "type"
     ]
-  },
-  "nb_hits": 26160
-}
```
shortcuts, line 33: This one is also mandatory

configs/wultra_developers.json
petrdvorak: Ah... I am sorry - we disabled this for the local debugging...

shortcuts: No worries :D

PR #3763, opened 2021-03-09
Upgrade Jest configurations

slorber:
We just deployed the Jest website using Docusaurus 2.

The old v1 website and older Jest versions are now at https://archive.jestjs.io/

Docsearch config jest => Docsaurus 1 (https://archive.jestjs.io/)
Docsearch config jest-v2 => Docusaurus 2 site (https://jestjs.io/)

### DIFF
--- a/configs/jest.json
+++ b/configs/jest.json
@@ -1,11 +1,12 @@
 {
   "index_name": "jest",
   "start_urls": [
-    "https://archive.jestjs.io/docs/"
+    "https://archive.jestjs.io/"
   ],
   "stop_urls": [
     "/help",
-    "/videos"
+    "/videos",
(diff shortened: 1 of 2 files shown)


slorber:
FYI @simenb

configs/jest.json
```diff
@@ -1,14 +1,15 @@
 {
   "index_name": "jest",
   "start_urls": [
-    "https://jestjs.io/docs/"
+    "https://archive.jestjs.io/"
```
shortcuts, line 4: @slorber I had to remove the `/docs` as it redirects to a 404 the crawl was ending in some cases

shortcuts:
@slorber both updates live

PR #3767, opened 2021-03-10
Add version "1.11" to wpcontentcrawler.json

turgutsaricam:
# Pull request motivation(s)
Adding a new version to the version variable so that the new version can be searched in isolation.

It is not possible to search the docs specific to version 1.11 in isolation.

To search the version 1.11's documentation in isolation.

None of them.

Could you please tell me if it is possible to cover the future versions without specifying each version in the configuration? If it is possible, I do not want to bother you with a pull request each time a new version needs to be added to the configuration.


shortcuts:
Hi,

You could implement something [like this](https://github.com/algolia/docsearch-configs/blob/master/configs/ignite.json#L6-L10) to avoid updating your config manually

turgutsaricam:
Thank you.

PR #3770, opened 2021-03-10
Update secureauth.json

lxbridges:
The start URLs in the current master file point to an old docs website. We want to point to the new site so DocSearch crawls the info there.

Regarding a site map, we do not currently have one because of the way we're building the website. We hope to move to a new build pattern by end of April that will allow us to structure our documentation so that we can build a sitemap.

# Pull request motivation(s)


shortcuts:
hi @lxbridges, I've added a version variable so you could filter you search [like this](https://docsearch.algolia.com/docs/config-file/#using-regular-expressions)

PR #3794, opened 2021-03-13
Ignore one subdir, remove TOC tree wrapper

scaleoutsean:
# Pull request motivation(s)

1) Need to eliminate double results resulting from ToC wrapper in left side-bar as [suggested](https://discourse.algolia.com/t/duplicate-results-in-docsearch-with-a-bootstrap-based-static-site/12206/2)

2) Need to eliminate docs from `docker/*` path because they create unnecessary search hits

Getting duplicate results because of 1) above (I hope; I don't know if I can test this without making this PR first)

Change 2) is just a minor problem as Docker-related results are very rarely required.

Regarding 1), I want to see one unique URL per result.

### DIFF
--- a/configs/trident.json
+++ b/configs/trident.json
@@ -2,7 +2,7 @@
   "index_name": "trident",
   "start_urls": ["https://scaleoutsean.github.io/trident/"],
   "sitemap_urls": ["https://scaleoutsean.github.io/trident/sitemap.xml"],
-  "stop_urls": ['https://scaleoutsean.github.io/trident/docker/*'],
+  "stop_urls": ["https://scaleoutsean.github.io/trident/docker/*"],
   "selectors": {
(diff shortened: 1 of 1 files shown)


configs/trident.json
```diff
@@ -2,7 +2,7 @@
   "index_name": "trident",
   "start_urls": ["https://scaleoutsean.github.io/trident/"],
   "sitemap_urls": ["https://scaleoutsean.github.io/trident/sitemap.xml"],
-  "stop_urls": [],
+  "stop_urls": ['https://scaleoutsean.github.io/trident/docker/*'],
```
shortcuts, line 5:
```suggestion
  "stop_urls": ["https://scaleoutsean.github.io/trident/docker/*"],
```

PR #3796, opened 2021-03-14
Update ignite.json

ydcjeff:
# Pull request motivation(s)

Headings are not shown in the search modal. It also searches into `_modules`, `_sources` directories

Headings should be shown in the search modal. It should not search into `_modules` and `_sources` directories

Current behavior:

Screen_Shot_2021-03-12_at_22 01 30_1

We want to show respective title (h1) in the place of "Documentation" and don't want to show "Source code for ..." results.

So I have included `/_` and `\\?` in `stop_urls`, updated the `lvl0` to `.rst-content h1`. Let me know if something needs to change.

cc: @vfdev-5, one of the maintainers of PyTorch-Ignite

### DIFF
--- a/configs/ignite.json
+++ b/configs/ignite.json
@@ -11,19 +11,19 @@
       }
     }
   ],
-  "stop_urls": [".txt", "^((?!\\.html).)*$"],
+  "stop_urls": ["/_", "\\?"],
   "selectors": {
     "lvl0": {
-      "selector": "",
+      "selector": ".rst-content h1",
       "global": true,
       "default_value": "Documentation"
     },
-    "lvl1": ".rst-content h1",
-    "lvl2": ".rst-content h2",
-    "lvl3": ".rst-content h3",
-    "lvl4": ".rst-content h4",
-    "lvl5": ".rst-content h5",
-    "text": ".rst-content p, .rst-content li"
+    "lvl1": ".rst-content h2",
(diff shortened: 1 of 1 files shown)


ydcjeff:
@vfdev-5 would you like to ignore searching in `quickstart`, `concetps`, `examples`, `faq`, `about`, and `governance` ?

vfdev-5:
@ydcjeff thanks for creating this PR !

> would you like to ignore searching in quickstart, concetps, examples, faq, about, and governance ?

I think we can keep them as we still want to find some info from there, IMO.

shortcuts:
Looks good! Feel free to let me know if there's any other issue

ydcjeff:
Thank you @shortcuts I have one question
What is the default value of `hitsPerPage` ?
We are using this [`@docsearch`](https://www.jsdelivr.com/package/npm/@docsearch/js?version=1.0.0-alpha.28) js and css files. I can't find the default value in the js files.

shortcuts:
The default value is 20, which is defined on the engine side when not sent in the query (you can see it in the `network` tab).

You can overwrite it in your DocSearch instance like this
```
searchParameters: {
    hitsPerPage: 50
}
```

Also, right now, the latest version of DocSearch v3 is the 3.0.0-alpha.34

ydcjeff:
Thank you for your answer @shortcuts

PR #3813, opened 2021-03-16
Update: add wait time

ArthurFlageul:
Certain pages are not indexed even though the selectors seem OK. 

Adding `wait_time` to see if this helps.

### DIFF
--- a/configs/talon.json
+++ b/configs/talon.json
@@ -39,7 +39,7 @@
       "lvl2": ".api-content h2",
       "lvl3": ".api-content h3",
       "lvl4": ".api-content h4",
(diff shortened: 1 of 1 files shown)


configs/talon.json
```diff
@@ -61,6 +61,7 @@
     ]
   },
   "js_render": true,
+  "js_wait": 3,
```
shortcuts, line 64:
```suggestion
  "js_wait": 1,
```

3 seems really long for content to load on an headless mode, we usually only need 1s per pages

ArthurFlageul:
Ok let's try that and see if it works 🤷‍♀️

PR #3817, opened 2021-03-18
Increase wait

ArthurFlageul:
Still no indexing, so let's increase the wait time.


shortcuts:
The `start_urls` order was the issue, fixed in https://github.com/algolia/docsearch-configs/pull/3818

PR #3818, opened 2021-03-18
fix config for talon

shortcuts:
fix **start_urls** order


shortcuts:
@ArthurFlageul as `api` is a subdomain of the root, it was using the default selector, it should be fixed in this PR

ArthurFlageul:
Oh man, thanks a lot 👍

PR #3844, opened 2021-03-21
Update 1host's config

### LINKED ISSUE #3843: [1host] Missing results for query "install"
No crawling seems to have happened, given that no searches work. Our home page only has one link into content, could that be it?

GrahamSH-LLK:
Attempt to copy the docus config because that's what we use.
Resolves #3843

# Pull request motivation(s)

We have no search results
We have search results

Fixes bug

### DIFF
--- a/configs/1host.json
+++ b/configs/1host.json
@@ -5,13 +5,24 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": "article h1",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @GrahamSH-LLK, feel free to ping me if there's still an issue after the update. 

You should see the changes in the 24 hours after the merge

PR #3877, opened 2021-03-28
fix(docusaurus): update domains

shortcuts:
cc @slorber


slorber:
😅 thanks, totally forget that

PR #3893, opened 2021-03-30
Add pinia

posva:
# Pull request motivation(s)

Adding docsearch to https://pinia.esm.dev

I was told the team is busy at the moment but I didn't know the configs were public so I hope this is okay 😅

### DIFF
--- a/configs/pinia.json
+++ b/configs/pinia.json
@@ -1,0 +1,48 @@
+{
+  "index_name": "pinia",
+  "start_urls": [
+    {
+      "url": "https://pinia.esm.dev/introduction.html",
+      "selectors_key": "introduction",
+      "tags": ["Introduction"],
+      "page_rank": 1
+    },
+    {
+      "url": "https://pinia.esm.dev/getting-started.html",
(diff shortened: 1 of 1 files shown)


configs/pinia.json
```diff
@@ -0,0 +1,48 @@
...
+  ],
+  "stop_urls": [],
+  "selectors": {
+    "lvl0": ".content h1",
+    "lvl1": ".content h2",
+    "lvl2": ".content h3",
+    "lvl3": ".content h4",
+    "lvl4": ".content h5",
+    "lvl5": ".content p, .content li",
+    "text": ".content [class^=language-]"
```
posva, line 43: I don't know if adding the code makes sense here but it would be nice to be able to search it if it's possible

shortcuts: as long as `document.querySelectorAll('yourSelector');` returns what you'd except to be returned from the search, it's good! (and in this case, it is)

posva:
@shortcuts No worries, the documentation is really helpful! I will submit it again

PR #3894, opened 2021-03-30
Vue Router: rank guide in front of api

posva:
# Pull request motivation(s)

Right now API results appear before the guide and I would like it to be the other way around

### DIFF
--- a/configs/next_router_vuejs.json
+++ b/configs/next_router_vuejs.json
@@ -4,6 +4,7 @@
     {
       "url": "https://next.router.vuejs.org/guide/",
       "selectors_key": "guide",
+      "page_rank": 10,
       "tags": [
         "guide"
       ]
@@ -11,6 +12,7 @@
     {
       "url": "https://next.router.vuejs.org/api/",
       "selectors_key": "api",
+      "page_rank": 8,
       "tags": [
         "api"
       ]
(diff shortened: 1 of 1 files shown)


shortcuts:
@posva, could you confirm that all the urls are duplicated (with and without `.html`)? so we can exclude one to avoid duplicate results

posva:
normally there are only URLs that end with `/` or with `.html`. The server does redirect those without them at the end, e.g.

- https://next.router.vuejs.org/guide -> https://next.router.vuejs.org/guide/
- https://next.router.vuejs.org/guide/essentials/route-matching-syntax -> https://next.router.vuejs.org/guide/essentials/route-matching-syntax.html


But the valid URLs are the ones with a trailing slash and `.html`.

If the crawler found links that do not follow that format I need to fix them on the docs

shortcuts:
Ok thanks, then only hiding `index.html` should be enough! https://github.com/algolia/docsearch-configs/pull/3895

posva:
Can we do the opposite? Hide all links that do not end with `.html` or `/`? I'm not sure it has an effect on SEO but if we can use only the html and slash-ending links, that would be better.


I noticed other vitepress/vuepress sites have the same issue

posva:
I just read the docs, I will do a PR

posva:
@shortcuts I created https://github.com/algolia/docsearch-configs/pull/3896

PR #3896, opened 2021-03-30
vue router and pinia avoid pages not ending with html or /

posva:
# Pull request motivation(s)

### DIFF
--- a/configs/next_router_vuejs.json
+++ b/configs/next_router_vuejs.json
@@ -19,7 +19,7 @@
     }
   ],
   "stop_urls": [
-    "(?<!\\.html|/)$"
+    "(?:(?<!.html)(?<!/))$"
   ],
   "selectors": {
     "guide": {
@@ -51,5 +51,5 @@
   "conversation_id": [
     "1305035158"
   ],
-  "nb_hits": 924
+  "nb_hits": 1518
 }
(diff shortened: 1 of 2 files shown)


configs/next_router_vuejs.json
```diff
@@ -19,7 +19,7 @@
     }
   ],
   "stop_urls": [
-    ".html"
+    "(?<!\\.html|/)$"
```
posva, line 22: I don't know what is used behind the scenes but it requires a negative look behind and I escaped the `\` so it can escape the `.`. But I also don't know if there is a special syntax to say this is a regex

shortcuts: We use Python so we will need fixed-width look behind, this should do it https://github.com/algolia/docsearch-configs/pull/3896/commits/2c3b5c04ce3268a6ce0546128fb18afd179b73fb

if there's any other indices feel free to open the PR/let me know :D

configs/next_router_vuejs.json
```diff
@@ -19,7 +19,7 @@
     }
   ],
   "stop_urls": [
-    ".html"
+    "(?:(?<!.html)(?<!/))$"
```
posva, line 22: Are you sure we don't need to escape the `.`? This will allow pages like `/rendering-html` and `/rendering-html.html`

shortcuts: oops, fixed in https://github.com/algolia/docsearch-configs/pull/3899

PR #3904, opened 2021-04-01
Update prisma_pan.json

csestito:
Update Prisma for API Docs JS

# Pull request motivation(s)

### DIFF
--- a/configs/prisma_pan.json
+++ b/configs/prisma_pan.json
@@ -1,7 +1,7 @@
 {
   "index_name": "prisma_pan",
   "start_urls": [
-    "https://prisma.pan.dev/docs/"
+    "https://prisma.pan.dev/"
   ],
   "sitemap_urls": [
     "https://prisma.pan.dev/sitemap.xml"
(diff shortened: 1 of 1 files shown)


configs/prisma_pan.json
```diff
@@ -10,22 +10,43 @@
...
+    "lvl5": "article h5, article td:first-child",
+    "lvl6": "article h6",
+    "text": "article p, article li, article td:last-child"
   },
   "selectors_exclude": [
     ".hash-link"
   ],
+  "js_render": true,
+  "use_anchors": true,
+  "js_wait": 5,
```
shortcuts, line 31:
```suggestion
  "js_wait": 1,
```

Usually, waiting 1~2 seconds between pages is enough, have you tried with a lower value?

csestito (commented):
Was able to cut down time, I believe 2 will work now

shortcuts:
> Was able to cut down time, I believe 2 will work now

Cool, thanks for the PR! 

You should see the update in the next 24 hours

PR #3913, opened 2021-04-02
fix(arrow): fix config

shortcuts:
update selectors and fix start/stop urls

### DIFF
--- a/configs/arrow.json
+++ b/configs/arrow.json
@@ -1,30 +1,23 @@
 {
   "index_name": "arrow",
-  "start_urls": [
-    "https://arrow-kt.io"
-  ],
+  "start_urls": ["https://arrow-kt.io"],
   "stop_urls": [
-    "https://arrow-kt.io/docs/0.9",
-    "https://arrow-kt.io/docs/next/core/",
-    "https://arrow-kt.io/docs/next/fx/",
-    "https://arrow-kt.io/docs/next/optics/dsl/",
-    "https://arrow-kt.io/docs/next/aql/intro/",
-    "!index.html"
+    "https://arrow-kt.io/docs/[0-9]",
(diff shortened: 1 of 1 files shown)


shortcuts:
@rachelcarmena some selectors were not working anymore, this selector update should offer better sections!

configs/arrow.json
```diff
@@ -1,30 +1,23 @@
...
-    "https://arrow-kt.io/docs/next/fx/",
-    "https://arrow-kt.io/docs/next/optics/dsl/",
-    "https://arrow-kt.io/docs/next/aql/intro/",
-    "!index.html"
-  ],
-  "sitemap_urls": [
-    "https://arrow-kt.io/sitemap.xml"
+    "https://arrow-kt.io/docs/[0-9]",
+    "https://arrow-kt.io/docs/next",
+    "index.html"
```
rachelcarmena, line 7: Thank you so much @shortcuts !! :pray: 

Please, I just miss `!index.html` instead of a raw `index.html` because we just want the crawler to visit the URLs that have `index.html`. 

For instance, this page is being indexed now:

* https://arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-ior/traverse.html#traverse

and we don't want to index it but only those pages with general data like this one:

* https://arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-ior/index.html

where the link to `https://arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-ior/traverse.html` appears.

Thank you so much again!! :raised_hands:

shortcuts: Fixed in https://github.com/algolia/docsearch-configs/pull/3915, enjoy :D

rachelcarmena: Thanks!! :+1:

PR #3915, opened 2021-04-02
fix(arrow): update config

shortcuts:
exclude any urls that don't ends with `index.html`

cc @rachelcarmena


rachelcarmena:
Thank you so much @shortcuts !! :clap:

PR #3931, opened 2021-04-06
brew_all: update level classes

EricFromCanada:
# Pull request motivation(s)

This improves the config for [docs.brew.sh](https://docs.brew.sh/) & [formulae.brew.sh](https://formulae.brew.sh/) by adding or updating the specific classes of elements to index, and also pluralizes the "Cask" heading.

Not all data was being indexed on certain pages due to the use of `nth-child`. For example, formulae with [aliases](https://formulae.brew.sh/formula/nutcracker) or [old names](https://formulae.brew.sh/formula/brightness) would not have their URLs added to the index.

This PR is dependent on Homebrew/formulae.brew.sh#446 being merged, which implements the class changes.


Rylan12:
Thanks, Eric. I have another suggestion about analytics that I'll put here just to have them written down but probably warrant a separate PR.

It would be nice to have some sort of analytics search ability. There are a two types that I would find helpful:
- Searching for "analytics" to bring up https://formulae.brew.sh/analytics/
- Searching for something like "macOS formula install events" would bring up https://formulae.brew.sh/analytics/install/30d/, https://formulae.brew.sh/analytics/install/90d/, or https://formulae.brew.sh/analytics/install/365d/
    - Essentially, be able to search for the pages linked on https://formulae.brew.sh/analytics/ (which seem to match `https://formulae.brew.sh/analytics(-linux)?/\w+/(30|90|365)d/`)

We wouldn't want to go any farther into the links on those pages, though, as they are already being crawled. I also don't see a need for any of the `/api` urls to be followed (which I think are all already excluded by the `\\.json` `stop_urls` entry)

Is this something that is doable? I've glanced at the documentation but don't see an easy way to stop the crawler from following internal links on those pages (which I think would be necessary). Any thoughts?

For context, the [analytics](https://formulae.brew.sh/analytics/) section of [formulae.brew.sh](https://formulae.brew.sh/) was removed from this configuration in https://github.com/algolia/docsearch-configs/pull/3922 in order to help reduce the number of records. I think part of the reason there were so many records is that each formula and cask page (which is already crawled by the formula and cask parts) are also linked in the `https://formulae.brew.sh/analytics/.../` pages, causing them to be recorded twice.

shortcuts:
Great improvements @EricFromCanada, the new crawl task will be queued right after the merge.

> Searching for "analytics" to bring up https://formulae.brew.sh/analytics/

[gist](https://gist.github.com/shortcuts/0d348d9f77415e85ecbfc120f1e3520d)

Screenshot 2021-04-07 at 17 00 32

> ...causing them to be recorded twice.

Our scraper exclude duplicate URLs but I took an other look at it and the real issue was the number of records per page on the analytics, it seems like it was timing out before the end of the task.

```
> DocSearch: https://formulae.brew.sh/analytics-linux/install/90d/ 19300 records)
> DocSearch: https://formulae.brew.sh/analytics/install/30d/ 30001 records)
> DocSearch: https://formulae.brew.sh/analytics-linux/install/365d/ 30001 records)
> DocSearch: https://formulae.brew.sh/analytics-linux/install/30d/ 13768 records)
> DocSearch: https://formulae.brew.sh/analytics/install/365d/ 30001 records)
```

If the goal is only to link the page, then it's possible to scrape anything else than the table itself.

Rylan12:
Thanks, @shortcuts. Sorry for the late response, I had an unexpectedly busy day yesterday. I'll plan on taking a closer look at this and opening a new PR for the analytics. No need to hold this one up for my suggestions.

shortcuts:
No worries @Rylan12, feel free to tag me or open an issue if needed!

PR #3942, opened 2021-04-08
fix language facet

posva:
# Pull request motivation(s)

I think the config I did wasn't right because it wasn't doing anything, empty results when adding:

```js
      searchParameters: {
		// lang can be 'en-US' or 'zh-CN'
        facetFilters: ['language:' + lang, 'tags:guide,api']
      }),
```

Is config correct with the current setup?

- English docs: https://next.router.vuejs.org/
- Chinese docs: https://next.router.vuejs.org/zh/index.html

### DIFF
--- a/configs/next_router_vuejs.json
+++ b/configs/next_router_vuejs.json
@@ -5,13 +5,19 @@
       "url": "https://next.router.vuejs.org/guide/",
       "selectors_key": "guide",
       "page_rank": 10,
-      "tags": ["guide"]
+      "tags": ["guide"],
+      "extra_attributes": {
+        "language": ["zh"]
+      }
     },
     {
       "url": "https://next.router.vuejs.org/api/",
       "selectors_key": "api",
       "page_rank": 8,
-      "tags": ["api"]
+      "tags": ["api"],
+      "extra_attributes": {
+        "language": ["zh"]
+      }
     }
   ],
   "stop_urls": ["(?:(?<!\\.html)(?<!/))$"],
@@ -26,7 +32,13 @@
       "lvl3": ".content h3",
       "lvl4": ".content h4",
       "lvl5": ".content h5",
-      "text": ".content p, .content li"
+      "text": ".content p, .content li",
+      "language": {
+        "selector": "/html/@lang",
+        "type": "xpath",
+        "global": true,
+        "default_value": "en-US"
+      }
     },
     "api": {
       "lvl0": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @posva 

Sorry I didn't caught the misplaced `language` selector last time.

It also looks like the `lang` doesn't change when on the `zh` version of the doc, so _I think_ the `extra_attributes` will be overridden by the `language` selector

configs/next_router_vuejs.json
```diff
@@ -5,13 +5,19 @@
...
     {
       "url": "https://next.router.vuejs.org/api/",
       "selectors_key": "api",
       "page_rank": 8,
-      "tags": ["api"]
+      "tags": ["api"],
+      "extra_attributes": {
+        "language": ["zh"]
+      }
     }
```
shortcuts: The crawler won't get the `.org/zh/x` endpoint here, you might need to go with something like: https://github.com/algolia/docsearch-configs/pull/3942/commits/04f65854de94d0e422e34b6f3b8cc75e246ec961

but we shouldn't merge the PR before https://github.com/algolia/docsearch-configs/pull/3942#issuecomment-815963322 is fixed

posva:
The docs needs to be fixed when it comes to setting the `lang` attribute on the HTML element but that will be fixed on our side (vitepress). I didn't understand very well what the extra attributes is meant to do, so if it can be removed altogether, then it's even better.

configs/next_router_vuejs.json
posva: that looks good, let me push the change

posva: Oh nvm, I somehow didn't see you pushed it already! 👍

posva: Right now, docs will always have the `en-US` in the lang attribute on the HTML element. Wouldn't that just always show English search results until I fix the lang attribute on my side?

Otherwise, yeah, I will have to push a fix first and deploy the docs. I imagine the search will be broken until this PR is merged. If that's the case what time would be the best to sync with you?

shortcuts: Yes, as of now everything will be under the `en-US` lang so it will show the results for both languages under the `en-US` filter.

I'm on Paris timezone so anything between 9am - 1am is good for me, the crawl will be queued right after the merge

posva: oh perfect, same for timezone for me! Then it sounds like it will behave the same as right now so it should be good

shortcuts: Sorry, what I meant was, right now the `zh` results are not crawled at all, after the merge, they will be but under the `en-US` lang.

In the meantime, we can still merge the fix for the `lang` selector so you'll be able to use the `tags`

posva: Sounds good to me. I guess we could also add a stop_url at `/zh` until I fix the problem of my end so it doesn't crawl the chinese docs under the same facet filter as the english ones

shortcuts: Sure, I'm not on a computer right now so feel free to push and I'll merge!

shortcuts:
Once the bug is fixed, just remove the `zh` filter and it's good to go! @posva

posva:
Thanks a lot, @shortcuts !

PR #3945, opened 2021-04-10
brew_all: re-add analytics pages

Rylan12:
# Pull request motivation(s)

Follow up to #3931 and #3922. See https://github.com/algolia/docsearch-configs/pull/3931#issuecomment-814530447

The analytics pages were intentionally excluded because they were making the crawl timeout. However, it would be helpful for these pages to still show up. The formulae linked on the pages, though, need not appear in the search.

Link showing maintainership: https://github.com/Homebrew/brew#who-we-are

Sitemaps: https://brew.sh/sitemap.xml, https://formulae.brew.sh/sitemap.xml

Searching for "analytics" or "formula install events" don't return the expected pages on https://brew.sh and https://formulae.brew.sh.

Searching for "analytics" would show a result for https://formulae.brew.sh/analytics/. Searching for formula install events would show results for pages like https://formulae.brew.sh/analytics/install/30d/ and https://formulae.brew.sh/analytics-linux/install/30d/.

I'd classify this as a bug

I've copied this change from the [gist](https://gist.github.com/shortcuts/0d348d9f77415e85ecbfc120f1e3520d#file-brew_all-json-L70-L82) that was linked in https://github.com/algolia/docsearch-configs/pull/3931#issuecomment-815016964. I've looked over it to verify that it seems to fit what I'm looking for, but I'm not super familiar with these configuration files so it's possible that I'm misunderstanding what this will do.

One thing I'm not sure about is whether the 30d/90d/365d distinction will be handled by the search? For each analytics category, we have three different time periods (which really is three different pages) so, ideally, each one would show up in the search so the user to navigate directly to the one they want.

Also, I don't think we need to include the `/api/*` entries in the table on https://formulae.brew.sh/analytics/. I think these will be automatically excluded because they are JSON files and should be excluded by the `stop_urls` entry.

### DIFF
--- a/configs/brew_all.json
+++ b/configs/brew_all.json
@@ -53,6 +53,14 @@
       "extra_attributes": {
         "site": ["formulae"]
       }
+    },
+    {
+      "url": "https://formulae.brew.sh/analytics/",
+      "selectors_key": "install_events",
+      "tags": ["analytics", "formula", "install_events"],
+      "extra_attributes": {
+        "site": ["formulae"]
+      }
     }
   ],
   "sitemap_urls": [
@@ -184,6 +192,22 @@
         "global": true,
         "default_value": "en"
       }
+    },
+    "install_events": {
+      "lvl0": {
+        "selector": "",
+        "global": true,
+        "default_value": "Analytics"
+      },
+      "lvl1": "#wrap h2",
+      "lvl2": "#wrap h2",
+      "text": "#wrap h3",
+      "lang": {
+        "selector": "/html/@lang",
+        "type": "xpath",
+        "global": true,
+        "default_value": "en"
+      }
     }
   },
   "custom_settings": {
@@ -191,5 +215,5 @@
     "separatorsToIndex": "_+@/"
   },
   "conversation_id": ["771112570"],
-  "nb_hits": 52725
+  "nb_hits": 36765
 }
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi @Rylan12,

In https://github.com/algolia/docsearch-configs/pull/3945/commits/6819f495a1d2063be3221fbe49dbf39ef7f5184e, I've updated the config so the above pages will show up.

The only issue is, typing something like `formula install events` will pop the pages with the same name for all the endpoints, unless something like `formula 30d` or something like this is typed. (which will show the `code` tag for the JSON api, as this is the only possible distinction)

If possible on your side, adding the time distinction to the `h2` could solve this issue.

Rylan12:
Thanks, @shortcuts.

> If possible on your side, adding the time distinction to the `h2` could solve this issue.

Good idea, I'll open a PR. It probably makes sense for that to be in there, anyway.

shortcuts:
Here's the results with the `h3` selector for now, which won't be needed on updated on your side.

Screenshot 2021-04-12 at 14 13 45

Would you like to merge this version or wait for the new one? @Rylan12

Rylan12:
Thanks, @shortcuts. I don't see any rush here, so may as well wait for https://github.com/Homebrew/formulae.brew.sh/pull/450 to be merged.

Rylan12:
@shortcuts that pull request has now been merged and the changes are live (see e.g. https://formulae.brew.sh/analytics/install/30d/)

shortcuts:
Looks good @Rylan12 ! The new crawl task will be queued after the merge

Rylan12:
Thanks, @shortcuts!

PR #3957, opened 2021-04-13
Vue I18n: update docsearch config

kazupon:
# Pull request motivation(s)
Reason: improvement of vue-i18n docs search UX
https://vue-i18n.intlify.dev/

The search UX is terrible because docsearch configuration is not optimized...

image


configs/vue-i18n.json
```diff
@@ -1,43 +1,89 @@
...
+      "url": "https://vue-i18n.intlify.dev/api/",
+      "selectors_key": "api",
+      "page_rank": 8,
+      "tags": [
+        "api"
+      ]
+    }
+  ],
+  "stop_urls": [
+    "\\.html"
```
shortcuts, line 38: @kazupon 

1. I've removed the `ja` endpoint exclusion, since the `lang` selector works on your site. To only show the results from the selected language, you can add the following `searchParameters` to your DocSearch instance
```js
searchParameters: {
  // language can be 'en-US' or 'ja-JP'
  // tags can be 'guide' or 'api'
  facetFilters: ['language:' + language, 'tags:' + tags]
}
```

2. For some reason, the routing on your site doesn't seems to be the same as on https://next.router.vuejs.org/ for example, which makes the previous RegEx (`"(?:(?<!\\.html)(?<!/))$"`) returns no URLs, in the meantime I've only been able to remove the duplicate results this way.

kazupon: @shortcuts 
Thank you for your follow up

> 1. I've removed the ja endpoint exclusion, since the lang selector works on your site. To only show the results from the selected language, you can add the following searchParameters to your DocSearch instance

Thank you for your advice!
For your `facetFilters` suggestion, the following PRs need to be merged.
https://github.com/vuejs/vitepress/pull/283

> 2. For some reason, the routing on your site doesn't seems to be the same as on https://next.router.vuejs.org/ for example, which makes the previous RegEx ("(?:(?<!\\.html)(?<!/))$") returns no URLs, in the meantime I've only been able to remove the duplicate results this way.

Thanks! no problem! 👍

kazupon:
@shortcuts 
Thanks!

shortcuts:
> @shortcuts 
> 
> Thanks!

No worries, let me know if there's any other issues. Have a great sunday

PR #3971, opened 2021-04-16
try to fix search url by changing h2 selector [https://alexkamaev.github.io]

AlexKamaev:
# Pull request motivation(s)
Attempt to fix search url by changin lvl2 selector from h2 to h2 a

Search results gives the url with id of header as an anchor, however, it's incorrect for my site (it's not a bug, it's just specifics)

Search result url should include if of a.achor-fix element, but not h2.id

No

No


shortcuts:
hi @AlexKamaev, this will indeed work for the URL but not for the text, as the `a` is empty.

AlexKamaev:
I found that the `transformData` methods suits to me perfectly

PR #3973, opened 2021-04-16
Revised Configuration Selectors

jeckenrode-ciq:
Changed the configuration selectors to match changes made in the container structure of our site.

We recently introduced a multi-container architecture to our site, which requires different search selectors. Currently, the search engine is returning results only for a subset of pages; the majority of the site is not included in the search. The new selectors should fix that.

Here's a link to the project: [https://github.com/ChartIQ/docsearch-configs](https://github.com/ChartIQ/docsearch-configs).

Thanks.

### DIFF
--- a/configs/chartiq_doc.json
+++ b/configs/chartiq_doc.json
@@ -12,15 +12,15 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


jeckenrode-ciq:
Thanks.

PR #3975, opened 2021-04-17
Update casthub.json

oyed:
# Pull request motivation(s)
CastHub has moved away from nuxt-content, as such some mark-up has changed


shortcuts:
fixed in https://github.com/algolia/docsearch-configs/pull/3976 since I wasn't able to push to your PR

PR #3986, opened 2021-04-20
fix(temporal): update config

shortcuts:
# Pull request motivation(s)


sw-yx:
thanks so much!!

PR #4008, opened 2021-04-26
Update demisto.json

csestito:
Search Page isn't work in docusaurus

# Pull request motivation(s)


shortcuts:
Hi,

This PR doesn't seem to change anything related to the search.

Looking at the `network` tab when sending a query from https://xsoar.pan.dev/search, it looks like all the `searchParameters` are doubled.

Could you please check the implementation on your side?

If the issue persists, feel free to open an issue!

Thanks!

PR #4010, opened 2021-04-26
Update prisma_pan.json

csestito:
Docusaurus Search Page is broken with old config

Docusaurus searchPage is broken 

# Pull request motivation(s)


shortcuts:
same as in https://github.com/algolia/docsearch-configs/pull/4008#issuecomment-827398839

PR #4015, opened 2021-04-27
adding some language based content ranking for temporal.io search

sw-yx:
# Pull request motivation(s)

adding some language based content ranking for temporal.io search

we were getting a lot of results out of order with less relevant SDKs up top (search "workflows" and you get PHP first, when most people will want Server and Go and Java docs first). 

also had a lot of repeat blog results from the multiple indexes. (so my solution is to use `stop_urls`... i hope it works)

more relevant search results

### DIFF
--- a/configs/temporal.json
+++ b/configs/temporal.json
@@ -113,7 +113,7 @@
   "stop_urls": [
     "https://docs.temporal.io/blog/tags/announcement/",
     "https://docs.temporal.io/blog/tags/case-study/",
-    "https://docs.temporal.io/blog/page/2/",
+    "https://docs.temporal.io/blog/tags/"
   ],
   "selectors": {
     "lvl0": {
--- a/configs/temporal.json
+++ b/configs/temporal.json
@@ -7,26 +7,102 @@
     },
     {
       "url": "https://docs.temporal.io/docs/concepts/workflows",
-      "page_rank": 998
+      "page_rank": 999
     },
     {
       "url": "https://docs.temporal.io/docs/concepts/activities",
-      "page_rank": 997
+      "page_rank": 999
     },
     {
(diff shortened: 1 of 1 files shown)


configs/temporal.json
```diff
@@ -7,34 +7,114 @@
...
+      "page_rank": 996
+    },
+    {
+      "url": "https://docs.temporal.io/docs/php/introduction",
+      "page_rank": 995
+    },
+    {
+      "url": "https://docs.temporal.io/docs/node/introduction",
+      "page_rank": 994
+    },
```
sw-yx, line 95: hopefully this pagerank propagates to the rest of the pages, trying to be lazier here as these will evolve more

configs/temporal.json
```diff
@@ -7,34 +7,114 @@
...
   "sitemap_urls": [
     "https://docs.temporal.io/sitemap.xml"
   ],
   "sitemap_alternate_links": true,
-  "stop_urls": [],
+  "stop_urls": [
+    "https://docs.temporal.io/blog/tags/announcement/",
+    "https://docs.temporal.io/blog/tags/case-study/",
+    "https://docs.temporal.io/blog/page/2/",
+  ],
```
sw-yx, line 117:
```suggestion
  "stop_urls": [
    "https://docs.temporal.io/blog/tags/announcement/",
    "https://docs.temporal.io/blog/tags/case-study/",
    "https://docs.temporal.io/blog/tags/"
  ],
```

spurious tags we dont really need

ideally i'd like Algolia to just not index ANY tags.

configs/temporal.json
shortcuts: All URLs under the given `url` will have the following `page_rank`!

sw-yx: thanks as always shortcuts

PR #4027, opened 2021-04-30
Update transposit.json

tbarn:
# Pull request motivation(s)
Including a new CSS class to exclude the `#` from showing up in search results

### DIFF
--- a/configs/transposit.json
+++ b/configs/transposit.json
@@ -18,6 +18,9 @@
     "lvl4": ".docsearch-content h5",
     "text": ".docsearch-content p, .docsearch-content li"
   },
+  "selectors_exclude": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey @tbarn, you can use this option to remove chars directly from the records: https://docsearch.algolia.com/docs/config-file/#removing-unnecessary-characters

tbarn:
@shortcuts does this work even if the char is in a different class? Like this: 

```html
<h2 id="cloning-a-runbook">Cloning a runbook <a class="direct-link" href="#cloning-a-runbook">#</a></h2>
```

tbarn:
@shortcuts nice. I just added to the root and removed the other one.

PR #4037, opened 2021-05-03
Update koishi.json

Shigma:
Add tags for versioning.

PS: I am not quite sure about `start_urls`. If there is an url matching `https://koishi.js.org/v1/`, will it be tagged with `latest`? I hope not.

PPS: el psy congroo

### DIFF
--- a/configs/koishi.json
+++ b/configs/koishi.json
@@ -1,7 +1,14 @@
 {
   "index_name": "koishi",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey @Shigma, as we follow the given order, anything under v1/ endpoint will be under the v1 tag. 

Sent from my IBM 5100.

PR #4039, opened 2021-05-04
Not used anymore

guillaumemeyer:
# Pull request motivation(s)


h0rhay:
This docsearch listing will be active within the next 28 days.
Our content team has been reviewing the site for the last month.
Please do not disable.

shortcuts:
I've kept your credentials active @h0rhay since you requested it.

Please make sure to have your config compliant before the next run!

h0rhay:
Hi Clement,

Thanks so much for that.
I'm not clear on what needs to be done to keep the config compliant however?

Regards,

George

PR #4042, opened 2021-05-05
Update midwayjs config file for vuepress

czy88840616:
# Pull request motivation(s)

update new website index for midwayjs

### DIFF
--- a/configs/midwayjs.json
+++ b/configs/midwayjs.json
@@ -10,12 +10,12 @@
       "global": true,
       "default_value": "Documentation"
     },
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey,

Is the `start_urls` still good? It looks like the documentation itself has moved to: https://www.yuque.com/midwayjs/

czy88840616:
start_urls is ok, we sync `yuque.com/midwayjs` to `https://midwayjs.org/start_app.html`. if algolia index complete, we will switch main document link to `start_app.html`

PR #4045, opened 2021-05-05
Update boc-group.json

TomPeirs:
We added a site which needs to be indexed.

# Pull request motivation(s)

Everything is working fine, but we added a new website which has the exact same structure on a different domain and we would like to index this too.


TomPeirs:
Oke, all ready to be merged :-)

PR #4052, opened 2021-05-06
Update Lowdefy config

SamTolmay:
# Pull request motivation(s)

We have changed our documentation site and would like to reactivate Docsearch on our site. I have updated the configuration for the new site, but I am not sure that it is correct.

Docsearch is currently deactivated, and the config is for a previous version of the documentation site.

Update the configuration.

Do we need to get a new  API key for the Docsearch widget?


shortcuts:
Hi @SamTolmay,

I've updated the selectors of you config, but we are still not able to retrieve all the available URLs.

Would you be able to add a `sitemap.xml` so we can proceed?

(The new API key will be sent to you directly by email)

SamTolmay:
Hi @shortcuts

Thanks for the update to the config.
I will be able to add a sitemap in the next few days. I will update this PR with the sitemap URL once I have added it.

SamTolmay:
HI @shortcuts 

We have added a sitemap to our docs and I have updated the config with the URL. I see the CI is failing, but I am not sure what is wrong with the config.

PR #4059, opened 2021-05-07
fix(supertokens): update config

shortcuts:
# Pull request motivation(s)


rishabhpoddar:
@shortcuts thanks for the changes to our config. Our site (supertokens.io) uses JS to render the correct content in the docs. However, this PR removes the `js_render` option. Would it still work fine?

PR #4060, opened 2021-05-07
Update signageos.json

lukdanek:
Show only the latest API docs version in the search bar, not the older versions to limit duplicity entries

# Pull request motivation(s)

### DIFF
--- a/configs/signageos.json
+++ b/configs/signageos.json
@@ -4,10 +4,6 @@
       "http://docs.signageos.io/knowledge-base/",
       "https://docs.signageos.io/provisioning/",
       {
-        "url": "http://docs.signageos.io/api/",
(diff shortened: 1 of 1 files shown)


configs/signageos.json
```diff
@@ -1,45 +1,66 @@
...
-    "default": {
-      "lvl0": {
-        "selector": "",
-        "default_value": "Documentation"
+    "index_name": "signageos",
+    "start_urls": [
+      "http://docs.signageos.io/knowledge-base/",
+      "https://docs.signageos.io/provisioning/",
+      {
+        "url": "http://docs.signageos.io/api/",
```
shortcuts, line 7: As this is a global scope, it should be pushed under the versioning block

PR #4067, opened 2021-05-10
fix(adonisjs_next): update config

shortcuts:
# Pull request motivation(s)

### DIFF
--- a/configs/adonisjs_next.json
+++ b/configs/adonisjs_next.json
@@ -28,13 +28,14 @@
       "variables": {
         "zone": "releases"
       }
-    }
+    },
+    "https://docs.adonisjs.com/"
   ],
   "stop_urls": [],
   "selectors": {
     "default": {
       "lvl0": {
-        "selector": "//*nav//*[contains(@class,'active')]/preceding::h3[1]",
+        "selector": "//*[contains(@class,'active')]/preceding::h3[1]",
         "default_value": "Documentation",
         "type": "xpath",
         "global": true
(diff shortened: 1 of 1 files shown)


configs/adonisjs_next.json
```diff
@@ -28,13 +28,14 @@
...
-    }
+    },
+    "https://docs.adonisjs.com/"
   ],
   "stop_urls": [],
   "selectors": {
     "default": {
       "lvl0": {
-        "selector": "//*nav//*[contains(@class,'active')]/preceding::h3[1]",
+        "selector": "//*[contains(@class,'active')]/preceding::h3[1]",
```
thetutlage, line 38: Isn't this going to select any element with the class `active`? I am trying to target just the active elements in the sidebar. Should I mention the `nav` element with some id or class?

thetutlage: Nope. The main concern was it selecting unwanted selectors like the `active` items in the header and the `active` items in the `codegroup tabs`. I don't see any immediate issues, but was mainly concerned with a broad selector

shortcuts: Make sense! As they don't have a preceding `h3`, it won't return anything in the other cases, but going with `//nav//*[contains(@class,'active')]/preceding::h3[1]` is indeed more correct (the issue was that `//*nav` isn't a correct xPath selector).

If you see any issue or to prevent it, feel free to update it, I'll merge right away

thetutlage: Alright. Yes, I am not much aware about the XPath syntax, so its possible to get it wrong.

Lemme play with the docs and send a combined PR for all the required tweaks.

Thanks a ton for your help 🙂

PR #4087, opened 2021-05-18
Add additional urls

Tulean:
# Pull request motivation(s)

We created additional paths to the documentation website, which has a separate repository. The app has separated builds for each version, thus, the `*` in the URL. Please let me know if it is possible to resolve the url issue with versions with this wildcard url strategy. We do not use Docusaurus versioning mechanism unfortunately.
The expected behaviour is that search will function correctly for all the builds which are separated into files.

ex. of our app structure
-v1 build
   -algolia config
-v2 build
   -algolia config
-v3 build
   -algolia config
index.html - which acts like a bridge between each version


configs/gooddata.json
```diff
@@ -2,11 +2,14 @@
...
-    "https://sdk.gooddata.com/gooddata-ui/docs/about_gooddataui.html"
+    "https://sdk.gooddata.com/gooddata-ui/docs/about_gooddataui.html",
+    "https://sdk.gooddata.com/gooddata-ui-apidocs/*/docs/index.html"
+
   ],
   "stop_urls": [],
   "sitemap_urls": [
-    "https://sdk.gooddata.com/gooddata-ui/sitemap.xml"
+    "https://sdk.gooddata.com/gooddata-ui/sitemap.xml",
+    "https://sdk.gooddata.com/gooddata-ui-apidocs/*/sitemap.xml"
```
shortcuts, line 12: Unfortunately, regexes are not supported in the `sitemap_urls`, so you'd have to define all of them manually, or have a sitemap that groups them all!

Tulean:
Hello, @shortcuts  thank you for a fast response to this PR. It is unfortunate that Algolia does not support wildcards since that will mean that we will have to update this config for each version. However, that is not a big issue. I have made changes according to your suggestions. Please take a look. Thank you

configs/gooddata.json
```diff
@@ -1,37 +1,39 @@
...
+            "variables": {
+                "version": ["8.4.0", "Next"]
+            }
+        }
+    ],
+    "stop_urls": [],
+    "sitemap_urls": [
+        "https://sdk.gooddata.com/gooddata-ui/sitemap.xml",
+        "https://sdk.gooddata.com/gooddata-ui-apidocs/v8.4.0/sitemap.xml",
+        "https://sdk.gooddata.com/gooddata-ui-apidocs/vNext/sitemap.xml"
```
shortcuts, line 17:
```suggestion
        "https://sdk.gooddata.com/gooddata-ui-apidocs/Next/sitemap.xml"
```

PR #4089, opened 2021-05-18
Update datachat.json

nategoethel:
# Pull request motivation(s)

On my site, docs.datachat.ai, the search index contains content from the old version of my site (with a different file structure) and the new site, leading to some things being double indexed in the search results. For example, searching for "Slice" provides the following results:
Screen Shot 2021-05-18 at 11 05 54 AM

The purple result goes to a dead link (because the site's structure has changed), while the link above it succeeds. You can also see that these are duplicate results.

Only the content live on the site is returned by the search.

Bug.

### DIFF
--- a/configs/datachat.json
+++ b/configs/datachat.json
@@ -8,13 +8,13 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".main h1",
-    "lvl1": ".main h2",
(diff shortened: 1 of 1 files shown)


shortcuts:
I wasn't able to push to your PR, it's now updated here: https://github.com/algolia/docsearch-configs/pull/4093

PR #4092, opened 2021-05-19
Add new url path to config

Tulean:
# Pull request motivation(s)

Hello, yesterday I have created a new config with additional urls (https://github.com/algolia/docsearch-configs/pull/4087); however, the search is not working as I expected. This is my fault that I have not included more open path for the crawler so I am adding it right now. 

The bigger issue that we are encountering right now is that we have two URLs and we need to separate the crawler so search is not looking into both paths at the same time.

When I am in the `gooddata-ui` path I will be able to search only the content of `gooddata-ui`. Same with `gooddata-ui-apidocs` I will be able to seach only content of `gooddata-ui-apidocs`. I hope we can resolve this somehow in the config, so far I have gone through the documentation and `stop_urls` seems like it won't do the job. Thank you for any suggestions

### DIFF
--- a/configs/gooddata.json
+++ b/configs/gooddata.json
@@ -1,47 +1,53 @@
 {
-    "index_name": "gooddata",
-    "start_urls": [
-        "https://sdk.gooddata.com/gooddata-ui/docs/",
-        "https://sdk.gooddata.com/gooddata-ui/docs/about_gooddataui.html",
-        {
-            "url": "https://sdk.gooddata.com/gooddata-ui-apidocs/v(?P<version>.*?)/",
-            "variables": {
-                "version": ["8.4.0", "Next"]
-            },
-            "tags": ["gooddata-ui-apidocs"]
-        },
-        {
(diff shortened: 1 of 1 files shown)


configs/gooddata.json
```diff
@@ -3,6 +3,12 @@
     "start_urls": [
         "https://sdk.gooddata.com/gooddata-ui/docs/",
         "https://sdk.gooddata.com/gooddata-ui/docs/about_gooddataui.html",
+        {
+            "url": "https://sdk.gooddata.com/gooddata-ui-apidocs/v(?P<version>.*?)",
```
shortcuts, line 7:
```suggestion
            "url": "https://sdk.gooddata.com/gooddata-ui-apidocs/v(?P<version>.*?)/",
```

(to avoid a warning in the crawler from missing / after regexes)

shortcuts:
If you'd like to filter your search based on a URL path, you can use the [`custom_tags`](https://docsearch.algolia.com/docs/config-file/#using-custom-tags) to do so

Tulean:
Hello, I have added the tags; hopefully, this will resolve this issue. Thank you

PR #4094, opened 2021-05-19
Update near.json to include github docs

mehtaphysical:
# Pull request motivation(s)

We want to index our GitHub docs in addition to our docs site. This will help users search better.

### DIFF
--- a/configs/near.json
+++ b/configs/near.json
@@ -1,26 +1,45 @@
 {
   "index_name": "near",
   "start_urls": [
-    "https://docs.near.org/docs/"
+    "https://docs.near.org/docs/",
+    "https://github.com/near/near-sdk-rs/blob/master/HELP.md"
   ],
   "sitemap_urls": [
-    "https://docs.near.org/sitemap.xml"
+    {
+      "url": "https://docs.near.org/sitemap.xml",
+      "selector_key": "docs"
+    },
+    {
+      "url": "https://github.com/near/near-sdk-rs/blob/master/HELP.md",
+      "selector_key": "near-sdk-rs"
+    }
   ],
   "sitemap_alternate_links": true,
   "stop_urls": [],
   "selectors": {
-    "lvl0": {
-      "selector": "//*[contains(@class,'navGroups')]//*[contains(@class,'navListItemActive')]/preceding::h3[1]",
-      "type": "xpath",
-      "global": true,
-      "default_value": "Documentation"
+    "docs": {
+      "lvl0": {
+        "selector": "//*[contains(@class,'navGroups')]//*[contains(@class,'navListItemActive')]/preceding::h3[1]",
+        "type": "xpath",
+        "global": true,
+        "default_value": "Documentation"
+      },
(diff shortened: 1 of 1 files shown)


mikedotexe:
> Hi there,
> 
> Even though it would work, we don't allow unowned domain scraping, would you be able to host this documentation instead?

Hey @shortcuts thanks for the response.
In our case we own all the links in this. Is there a way we can move forward with this?

shortcuts:
Hey @mikedotexe,

Sorry if my answer wan't clear but the issue is that we don't index GitHub website, the documentation needs to be hosted

volovyk-s:
Hey @shortcuts ! Now we are hosting this documentation at https://near-sdk.io. Can we now index it? Should I open a separate PR? How it is usually done?

shortcuts:
Hey @volovyk-s,

Sure, you can open a new PR with the updated config. Also, since you are using Docusaurus v2, make sure to also include this section: https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json#L13-L45

volovyk-s:
@shortcuts looks like we are using `Docusaurus` on our main docs website([repo]( https://github.com/near/docs/ ), [website](https://docs.near.org/)) and `Docusaurus 2` on  the one that I want to add to the index ([repo](https://github.com/near/sdk-docs/), [website](https://near-sdk.io/)). Is that a problem? Will I break the existing search by these changes in the config `selectors` section?

shortcuts:
You could use [`selectors_key`](https://docsearch.algolia.com/docs/config-file/#selectors_key-tailor-your-selectors) to target both domains, one with the selectors for Docusaurus v1, and the other for the v2, and then use [`tags`](https://docsearch.algolia.com/docs/config-file/#using-custom-tags) to filter the search on the frontend between the SDK and docs.

Also, it seems that the `Search by Algolia` is hidden on the documentation, which is part of our policy for DocSearch: https://docsearch.algolia.com/docs/styling/#attribution

amgando:
I know I'm a a little late to the party but we could also just license Algolia directly and customize most of this right?  the only reason some these limitations are in place is because we're using the default free version of DocSearch that ships with Docusaurus

shortcuts:
We don't really enforce having the `Search by Algolia` displayed, but it's the only counterpart of having DocSearch for free, so I was pointing it out!

You can style the DocSearch components as you want, it's something we encourage :D

thisisjoshford:
@shortcuts I would like to upgrade our Algolia search feature to the paid version. Can you point me in the right direction to accomplish this?

shortcuts:
Hey @thisisjoshford, sure, you can contact support@algolia.com and they will help you estimate the cost of you plan based on the number of searches and the number of records you have.

You will still be able to use DocSearch https://docsearch.algolia.com/docs/run-your-own/ if you want, or any Algolia features!

thisisjoshford:
Awesome! I'll reach out. :) Thanks @shortcuts !

PR #4097, opened 2021-05-20
Update apimatic.json

thehappybug:
# Pull request motivation(s)

APIMatic has migrated to Docusaurus 2. I have updated the apimatic.json, adapted from https://github.com/algolia/docsearch-configs/blob/5ca11a9/configs/docusaurus-2.json.

I don't know how accurate my work is. This will need review.

Note that, unlike the Docusaurus team, we do not use versioning and do not have translated docs. We also have a [changelog](https://docs.apimatic.io/changelog) (based on Docusaurus blog) alongside the main docs.

The search config for APIMatic was based on a custom Hugo docs that is no longer up.

Search works on the new Docusaurus site. I have attempted to submit a patch for the config. Once you confirm the patch, I can update my Docusaurus site to add search again.

Update to APIMatic's config.

No.

### DIFF
--- a/configs/apimatic.json
+++ b/configs/apimatic.json
@@ -6,26 +6,41 @@
   "sitemap_urls": [
     "https://docs.apimatic.io/sitemap.xml"
   ],
-  "stop_urls": [
-    "https://docs.apimatic.io$",
-    "api-description",
-    "keep-in-touch"
-  ],
+  "sitemap_alternate_links": true,
+  "stop_urls": [],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Looks good, I'd suggest you to enable search before merging, since the DocSearch plugin on Docusaurus side will add meta tags that we use for search faceting.

Feel free to let me know once updated on your side, and I'll merge the PR, which will trigger a new crawl.

thehappybug:
Hey @shortcuts. We have updated the docs site which should have the search enabled now. You can check out the website here: https://docs.apimatic.io/

The config for this in our docusaurus.config.js file looks like this. Let me know if any settings need to be changed:

```json
    algolia: {
      apiKey: 'REDACTED',
      indexName: 'apimatic',
    },
```

shortcuts:
Should be all good!

PR #4112, opened 2021-05-25
Adding examples to searchable content

jayair:
Hi there, we recently added a new section to the site that needs to be indexed as well. Hopefully adding to the `start_urls` does that.

Let me know if it doesn't!


jayair:
Thank you for providing this service!

shortcuts:
Looks good!

PR #4115, opened 2021-05-25
Update milvus.json

shanghaikid:
[Milvus.io](https://milvus.io) has different versions of docs, we should apply the search for the specific selected version. 

we change attributesForFaceting setting in the json. so that we can use it on frontend.


shortcuts:
Hey, here's how you should update your DocSearch instance on your side, in order to filter the search results by versions: https://docsearch.algolia.com/docs/config-file/#using-regular-expressions

PR #4118, opened 2021-05-25
Add Boosted v5.0 to DocSearch

Nurovek:
Hello, we are on the verge of releasing stable version 5.0.0 of Boosted, which means we need to add it to DocSearch

However, it will need to wait for [Boosted #670](https://github.com/Orange-OpenSource/Orange-Boosted-Bootstrap/pull/670) to be finished.

I will update this PR once this is done.


Nurovek:
V5.0 is available !

shortcuts:
Congrats on the release! :)

PR #4124, opened 2021-05-26
add additional start_urls and tags, increased js_wait time

jscyo:
# Pull request motivation(s)
Added additional start_urls and tags, increased js_wait time

### DIFF
--- a/configs/supertokens.json
+++ b/configs/supertokens.json
@@ -32,21 +32,9 @@
       ]
     },
     {
-      "url": "https://supertokens.io/docs/nodejs/installation",
-      "tags": [
-        "nodejs"
-      ]
-    },
-    {
       "url": "https://supertokens.io/docs/nodejs/",
       "tags": [
         "nodejs"
-      ]
-    },
-    {
-      "url": "https://supertokens.io/docs/ios/installation",
-      "tags": [
-        "ios"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


configs/supertokens.json
```diff
@@ -1,6 +1,18 @@
...
+      "tags": [
+        "community"
+      ]
+    },
+    {
+      "url": "https://supertokens.io/docs/community/",
+      "tags": [
+        "community"
+      ]
+    },
```
shortcuts, line 15:
```suggestion
    {
      "url": "https://supertokens.io/docs/community/",
      "tags": [
        "community"
      ]
    },
```

This should be enough, since it works as a path to match, anything under the endpoint `https://supertokens.io/docs/community/` will be tagged as `community`, same goes for all the other URLs

configs/supertokens.json
```diff
@@ -84,7 +205,7 @@
     ]
   },
   "js_render": true,
-  "js_wait": 2,
+  "js_wait": 8,
```
shortcuts, line 208:
```suggestion
  "js_wait": 5,
```

A high number could lead into some timeouts, we recommend using 5 as a max value but if it didn't worked with 5, we can try it out.

configs/supertokens.json
jscyo: When running the crawler locally we were having issues where, when just giving the https://supertokens.io/docs/community/  URL only the https://supertokens.io/docs/community/introduction would be crawled and the remaining endpoints would not

shortcuts: Adding the root of the domain also helps sometime for missing URLs in case of client-side rendering.

In that case adding `"https://supertokens.io/docs/"` to the end of the `start_urls` seems to work for me locally

jscyo: Alright, cool will update the PR with the change

PR #4131, opened 2021-05-27
update conflux-chain.json migrate conflux-chain to confluxnetwork

yqrashawn:
# Pull request motivation(s)

We are migrating from conflux-chain.org to confluxnetwork.org.

You can check that https://developer.conflux-chain.org is being redirect to https://developer.confluxnetwork.org

I'm not sure if change these two lines will work. Please confirm.

Current behaviour is fine. I'm not sure if further updates will get crawled by the crawler.

### DIFF
--- a/configs/conflux-chain.json
+++ b/configs/conflux-chain.json
@@ -1,10 +1,10 @@
 {
   "index_name": "conflux-chain",
   "start_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Looks good!

PR #4138, opened 2021-05-28
update oasisengine.json

eyworldwide:
# Pull request motivation(s)
Two motivations:
1. Use  `tags`  in  `start_urls`  instead of  `lang` in `attributesForFaceting`  to filter the  the *i18n* content.
2. Replace the old `lvl0` of `selectors` with new value because the crawler can't fetch the html element generated by javascript dynamically.

1. The config `lang` in `attributesForFaceting` is not good for my `i18n` solution.
2. the crawler can't grab the `lvl0` html element.

1.  Can filter the search result by  current language setting.
2. The crawler can fetch the `lvl0` html element.

### DIFF
--- a/configs/oasisengine.json
+++ b/configs/oasisengine.json
@@ -7,7 +7,23 @@
         "version": [
           "0.3"
         ]
-      }
+      },
+      "tags": [
+        "en"
+      ],
+      "page_rank": 2
+    },
+    {
+      "url": "https://oasisengine.cn/(?P<version>.*?)/docs/.+?-cn",
+      "variables": {
+        "version": [
+          "0.3"
+        ]
+      },
+      "tags": [
+        "cn"
+      ],
+      "page_rank": 2
     },
     {
       "url": "https://oasisengine.cn/(?P<version>.*?)/api/",
@@ -15,7 +31,8 @@
         "version": [
           "0.3"
         ]
-      }
+      },
+      "page_rank": 1
     }
   ],
   "sitemap_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
> Replace the old lvl0 of selectors with new value because the crawler can't fetch the html element generated by javascript dynamically.

You can, by adding the [`js_render: true`](https://docsearch.algolia.com/docs/config-file/#javascript-rendering) option, but we recommend to focus on server side rendered content

`tags` should also be added to the `attributesForFaceting` array

eyworldwide:
Thank you for your advice. I'll update the PR soon.

eyworldwide:
And One more question,  I want to search the API content in different language setting environments, So I have to add all language tags OR just omit the `tags` config:

1.

```json
    {
      "url": "https://oasisengine.cn/(?P<version>.*?)/api/",
      "variables": {
        "version": [
          "0.3"
        ]
      },
      "tags": [
        "en",
        "cn"
      ],
      "page_rank": 1
    }
```
or:

2.
```json
    {
      "url": "https://oasisengine.cn/(?P<version>.*?)/api/",
      "variables": {
        "version": [
          "0.3"
        ]
      },
      "page_rank": 1
    }
```

eyworldwide:
@shortcuts Could you give me some advice about the question above? 😃

shortcuts:
Oops I didn't saw it.

Do you mean you'd like the content from `/api` to be returned in any languages? In this case, you can go with `#2` and don't send `facetFilters`

eyworldwide:
@shortcuts Yes. The `#1` also works in any any languages. I've seen the results in my site. Thank you very much!

shortcuts:
Sure, as long as you don't send the required `facetFilters` it will work, the `#2` is just a bit more "correct" :D 

Enjoy!

eyworldwide:
OK. I'll update the config by the "correct" way in the future.

PR #4139, opened 2021-05-29
Use sitemap for docs.aspect.dev

alexeagle:
Prevents the crawler from seeing historical versions of docs (currently there are five copies of rules_nodejs docs for example)

Per https://docsearch.algolia.com/docs/config-file#sitemap_urls-optional this can replace start_urls

### DIFF
--- a/configs/aspect.json
+++ b/configs/aspect.json
@@ -1,9 +1,6 @@
 {
   "index_name": "aspect",
-  "start_urls": [
-    "https://docs.aspect.dev"
-  ],
-  "stop_urls": [],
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, if you'd like the crawler to only follow the URLs in the sitemap, you can remove the `start_urls` option

alexeagle:
I did that in my first commit but the CI says it doesn't conform to the schema

alexeagle:
Oh I bet I just need `start_urls` to be present but an empty list :)

PR #4150, opened 2021-06-02
Update goharbor config

OrlinVasilev:
# Pull request motivation(s)
We need update for the crawing of the website only under /docs
Also we need to properly handle the versions, so to search in only the selected version.

https://github.com/goharbor/community/blob/master/MAINTAINERS.md 

Currently returns results from blogs, docs.. etc we need to limit that to docs only.

To show only results in docs

### DIFF
--- a/configs/goharbor.json
+++ b/configs/goharbor.json
@@ -1,9 +1,12 @@
 {
   "index_name": "goharbor",
   "start_urls": [
-    "https://goharbor.io"
+    "https://goharbor.io/docs"
   ],
   "stop_urls": [],
+  "sitemap_urls": [
+    "https://goharbor.io/sitemap.xml"
+  ],
   "selectors": {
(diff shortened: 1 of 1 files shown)


OrlinVasilev:
@jonasrosland

OrlinVasilev:
@shortcuts can you take a look please

shortcuts:
> Also we need to properly handle the versions, so to search in only the selected version.

You can use the following method to handle versioning on your DocSearch instance https://docsearch.algolia.com/docs/config-file/#using-regular-expressions

OrlinVasilev:
@shortcuts 
We want to implement the same functionality as we have in 
https://github.com/algolia/docsearch-configs/blob/master/configs/velero.json

shortcuts:
Make sense!

PR #4154, opened 2021-06-03
Update cxjs.json

mstijak:
I noticed that Algolia based search is not working properly for [CxJS documentation](https://docs.cxjs.io). We redesigned the documentation and some of the selectors have changed. Can we try this configuration?


shortcuts:
looks good!

PR #4166, opened 2021-06-08
Update projectcontour.io search

jonasrosland:
Signed-off-by: Jonas Rosland <jrosland@vmware.com>

# Pull request motivation(s)
Updating the search crawler for projectcontour.io after a migration from Jekyll to Hugo.

Currently all search for old versions work, but it doesn't scrape for new content.

Current and new content should be scraped and available for search.

No


shortcuts (approved):
Looks good!

PR #4167, opened 2021-06-08
Update rucio-cern.json (now docusaurus v2)

bziemons:
https://rucio.cern.ch/documentation/ was updated to docusaurus v2 lately and the search wasn't working anymore after that. I have taken https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json as a base for my changes. Please have a closer look at the `custom_settings` since I do not know if they apply to all docusaurus v2 websites. The selectors check out afaict.

### DIFF
--- a/configs/rucio-cern.json
+++ b/configs/rucio-cern.json
@@ -44,5 +44,5 @@
   "conversation_id": [
     "1339779250"
   ],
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
This looks good!

PR #4168, opened 2021-06-08
fixing boosted-orange config following v5.0 release

Lausselloic:
update boosted-orange config following v5.0 release and semantic update in documentation fixing incomplete PR #4118

Current search doesn't work on V5 https://boosted.orange.com/docs/5.0/ due to this new semantic


shortcuts (approved):
Looks good!

PR #4188, opened 2021-06-14
update config to make it work on vuepress2

Mister-Hope:
See https://vuepress2.netlify.app/reference/plugin/docsearch.html#get-search-index

image

I think you should also take care of this if other vuepress2 sites are requesting docsearch. I think it wound be nice if it can be corrected on the default config for our sites intead of opening a pr themselevs for everyone using vuepress2

### DIFF
--- a/configs/waline.json
+++ b/configs/waline.json
@@ -22,6 +22,9 @@
       "global": true
     }
   },
+   "custom_settings": {
+    "attributesForFaceting": ["lang"]
+  },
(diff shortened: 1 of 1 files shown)


shortcuts:
I've answered on this issue: https://github.com/vuepress/vuepress-next/issues/217#issuecomment-860644744

Thanks for pointing that out, I'll see where's the issue!

PR #4195, opened 2021-06-14
[Temporal config] add stop url to prevent indexing shared content

sw-yx:
# Pull request motivation(s)

algolia was starting to index a private page that wasn't linked... not sure how

CleanShot 2021-06-15 at 07 42 47

nothing  that is in the /shared folder should show up in algolia.

CleanShot 2021-06-15 at 07 43 07

intended solution is to use a regex in `stop_urls` to prevent the crawl

### DIFF
--- a/configs/temporal.json
+++ b/configs/temporal.json
@@ -113,7 +113,8 @@
   "stop_urls": [
     "https://docs.temporal.io/blog/tags/announcement/",
     "https://docs.temporal.io/blog/tags/case-study/",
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
Looks good! just pushed a quick update which groups everything

sw-yx:
thanks for the fix shortcuts! wasnt sure if that `\/` syntax was needed since i was using regex

shortcuts:
No worries! It does work but it also accepts a matching pattern so in that case it's easier to review :D

PR #4199, opened 2021-06-15
Update "plaiceholder" to use docs-specific sitemap

joe-bell:
# Pull request motivation(s)

The current configuration references the domain's sitemap, but I only want DocSearch to crawl over the `/docs` (which has a specific sitemap)

Sets the correct sitemap URL

N/A


joe-bell:
Thank you 🙏

PR #4201, opened 2021-06-15
Add SendGrid docs

smendes:
# Pull request motivation(s)

We would like to extend the index with SendGrid's documentation placed under the same platform but under a different domain.

Let me know if you have any questions! Thanks!

The current behaviour is to only crawl https://twilio.com/docs

We would like to have results from https://twilio.com/docs as well as https://docs.sendgrid.com


maylonpedroso:
:+1:

shortcuts:
Hey, do you want to have results from both domains under the same search? If that's not the case, I'd suggest to add [these parameters](https://docsearch.algolia.com/docs/config-file/#using-custom-tags) to your DocSearch instance before the merge

maylonpedroso:
Hi @shortcuts! Yes, we want to have the results from both `docs.sendgrid.com` and `www.twilio.com/docs` under the same search.

shortcuts:
Cool! Then you can merge the main branch so I can merge the PR!

PR #4204, opened 2021-06-16
Update taichi.json

rexwangcc:
# Pull request motivation(s)

We recently moved our docs to a new subdomain https://docs.taichi.graphics, after switching to a different static website generator (docusaurus), hence the algolia configuration needs to be updated accordingly :) Note we also made significant changes to the structure so the stop urls are updated too.

I was trying to follow the existing official config of docusaurus [docusaurus-2.json](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json), but still not sure why I shouldn't use:

```
a[contains(@class, 'menu__link navbar__link--active')]
```

in selectors instead of the official's

```
a[contains(@class, 'menu__link menu__link--sublist menu__link--active')]
```

Thank you so much!!!


configs/taichi.json
```diff
@@ -1,48 +1,51 @@
...
   "stop_urls": [
-    "/tutorials/",
-    "/community/",
-    "/events/",
-    "/research/",
-    "/contribution/",
-    "/gallery/",
-    "/faq/"
+    "/community/members/",
+    "/acknowledgments/",
```
rexwangcc, line 12:
```suggestion
    "/acknowledgments/"
```

configs/taichi.json
```diff
@@ -1,48 +1,51 @@
...
-    }
+      "default_value": "Docs"
+    },
+    "lvl1": "header h1",
+    "lvl2": "article h2",
+    "lvl3": "article h3",
+    "lvl4": "article h4",
+    "lvl5": "article h5, article td:first-child",
+    "lvl6": "article h6",
+    "text": "article p, article li, article td:last-child",
```
rexwangcc, line 27:
```suggestion
    "text": "article p, article li, article td:last-child"
```

PR #4207, opened 2021-06-16
chore: update config for sli.dev

antfu:
# Pull request motivation(s)

Update configs for https://sli.dev to provide search for different translations.

- ~~I also renamed `sli` cluster to `slidev` to be more properly referencing to the project's name. I am not sure if this would break anything, if so, I can revert the renaming back for sure.~~
- Is that possible to rename it from `sli` to `slidev`? (not a big deal tho)

Thanks. 


shortcuts:
thanks for the PR @antfu, I've renamed the index/file to `slidev` and grouped the `language` variables together, you'll be able to use like this: 

```js
docsearch({
  […],
  searchParameters: {
    'facetFilters': ["language:en"]
  },
  […],
});
```

shortcuts:
The search API key remains the same, but don't hesitate to @ me if there's any issue, a new crawl will be queued after the merge

antfu:
Thanks a lot!

PR #4208, opened 2021-06-16
fix: update config for vueuse.org

antfu:
# Pull request motivation(s)

Fixing levels query for vueuse.org

image


configs/vueuse.json
```diff
@@ -1,23 +1,60 @@
 {
   "index_name": "vueuse",
   "start_urls": [
-    "https://vueuse.org/guide.html",
-    "https://vueuse.org/functions.html",
-    "https://vueuse.org/add-ons.html",
-    "https://vueuse.org/"
```
shortcuts, line 7: Hey, was this intentional? With the new config, you'll only have the stated pages indexed

shortcuts: this would be the alternative: https://github.com/algolia/docsearch-configs/pull/4208/commits/87153b7101dc7f68408108d0db68f2079c951688

antfu: Yeah, that would be great, thanks! (I as just referring vitepress's config which does not have index crawling so I wasn't sure how to do that)

PR #4232, opened 2021-06-22
use custom tags for fluxcd

dholbach:
Signed-off-by: Daniel Holbach <daniel@weave.works>

As of #4190 Algolia excludes `fluxcd.io/legacy/*` urls. 

As @shortcuts noted in https://github.com/algolia/docsearch-configs/pull/4190#pullrequestreview-683025654 custom tags, we'd like to make use of this feature starting by using this PR.

### DIFF
--- a/configs/fluxcd.json
+++ b/configs/fluxcd.json
@@ -2,12 +2,12 @@
   "index_name": "fluxcd",
   "start_urls": [
     {
-      "url": "https://fluxcd.io/",
-      "tags": ["current"]
-    },
-    {
       "url": "https://fluxcd.io/legacy",
       "tags": ["legacy"]
+    },
+    {
+      "url": "https://fluxcd.io/",
+      "tags": ["current"]
     }
   ],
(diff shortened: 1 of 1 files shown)


configs/fluxcd.json
```diff
@@ -1,14 +1,18 @@
...
   "start_urls": [
-    "https://fluxcd.io/"
+    {
+      "url": "https://fluxcd.io/",
+      "tags": ["current"]
+    },
+    {
+      "url": "https://fluxcd.io/legacy",
+      "tags": ["legacy"]
+    }
```
shortcuts, line 11:
```suggestion
    {
      "url": "https://fluxcd.io/legacy",
      "tags": ["legacy"]
    },
    {
      "url": "https://fluxcd.io/",
      "tags": ["current"]
    }
```

It needs to be the other way around since `legacy` is a subdomain of the root one

shortcuts:
Looks good!

PR #4258, opened 2021-06-25
Inconsistent results for Lowdefy docs indexing

Gervwyk:
Hi, First off, thanks for an awesome product! Docsearch is great 💫

On the Lowdefy docs: https://docs.lowdefy.com, we are getting very inconsistent search results. For example, some days "SQL" searches will show up, others not, it's like the index only catch some data, some of the time, and almost always misses a lot. 

See examples, some days the search work on some pages, other days not:
Screenshot 2021-06-25 at 11 16 26
Screenshot 2021-06-25 at 11 17 09

Not sure how to tweak to config to achieve a better index. Please help?

### DIFF
--- a/configs/lowdefy.json
+++ b/configs/lowdefy.json
@@ -22,6 +22,6 @@
   },
   "conversation_id": ["964484389"],
   "js_render": true,
-  "js_wait": 4,
+  "js_wait": 6,
   "nb_hits": 105
 }


shortcuts:
Hey, are those inconsistent URLs in the sitemap?
Also, if they are rendered client-side, increasing the `js_wait` (as you did) is probably the best option, or maybe state them in the `start_urls` to make sure they are not ignored during the crawl. 

Have you tried running the scraper locally? You can do so by following this link: https://docsearch.algolia.com/docs/run-your-own/

Gervwyk:
Thanks, I'll try and run it locally as well to further improve. Yes, the sitemap is autogenerated so they are present. 

I think it's a loading time thing, since some days the pages do show up in the index and others not, although no-doc update on our side in between. Let's see if the longer wait time resolves this, I know the page often takes a few seconds to load. And it is client-side rendered.

PR #4265, opened 2021-06-25
chore(config): nxdev config updates

bcabanes:
# Pull request motivation(s)
This PR enables the indexing and search for the new nx.dev website release.

### DIFF
--- a/configs/nx.json
+++ b/configs/nx.json
@@ -23,22 +23,20 @@
   "sitemap_alternate_links": true,
   "selectors": {
     "lvl0": {
-      "selector": "#nav h5",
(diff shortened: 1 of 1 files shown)


configs/nx.json
```diff
@@ -23,22 +23,20 @@
   "sitemap_alternate_links": true,
   "selectors": {
     "lvl0": {
-      "selector": "#nav h5",
+      "selector": "//nav[contains(@id, 'nav')]//span[contains(@class, 'text-gray-900')][1]/preceding::h5[1]",
```
shortcuts, line 26: Hey, thanks for the PR.

Due to future changes in our crawler, we recommend users to find a DOM alternative to xPaths selectors.

I'm still merging it for now since it will still work

bcabanes: Thanks for the heads up!

PR #4272, opened 2021-06-27
materialize: update start URL

benesch:
We've moved the docs from materialize.io to materialize.com. Seems like this config needs to be updated like so. Thanks in advance!


shortcuts:
Looks good! Feel free to let me know if there's any issue

PR #4279, opened 2021-06-28
Update dhtmlx.json

Labirintami:
# Pull request motivation(s)


Labirintami:
Hi! We have released the documentation at https://docs.dhtmlx.com/suite

PR #4301, opened 2021-07-02
Updates apostrophe config for v2 and v3 tags (#1)

abea:
# Pull request motivation(s)

The Apostrophe documentation is now in two different websites for versions two and three. We need each indexed separately. @shortcuts helped identify this as the correct strategy to do that.

The config currently indexes a deprecated URL.

We'd like it to index the two live URLs.


shortcuts:
Looks good :)

PR #4323, opened 2021-07-07
Update cardano.json with subdomain to crawl

h0rhay:
We would like to crawl the subdomain: https://developers.cardano.org/ so we can have docsearch active on it too.
I was emailed by support who suggested this change.
Many thanks

# Pull request motivation(s)

### DIFF
--- a/configs/cardano.json
+++ b/configs/cardano.json
@@ -48,6 +48,9 @@
       "text": "[class*='MainContent'] p, [class*='MainContent'] li"
     }
   },
+  "custom_settings": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey @h0rhay, I've pushed an update to your config which makes both websites searchable and with the right selectors. 

Make sure to check the doc to filter your search for both websites: https://docsearch.algolia.com/docs/config-file/#using-custom-tags

Have a nice day!

h0rhay:
Many thanks @shortcuts 
This looks great!

PR #4332, opened 2021-07-08
Update zowe.json

nannanli:
Signed-off-by: nannanli <nannanli@cn.ibm.com>

# Pull request motivation(s)
We just migrated our site https://docs.zowe.org/ to use a new staticsite generator. With the migration, the URLs of the site has some changes.  
- Previously, the versions uses dashses such as "v1-2-0" while now it uses dots "v1.2.0". 
- We archived some doc versions so can disable search for those versions. 

Thus the site might need to be re-indexed. 

Now the search on our beta site returns results from all available versions. What we expect is that when users start a search, it returns result for that single version only. So for example, if they are searching within v1.21.x docs, only return results from https://docs.zowe.org/v1.21.x

### DIFF
--- a/configs/zowe.json
+++ b/configs/zowe.json
@@ -7,32 +7,27 @@
          "version": [
           "stable",
           "active-development",
-          "v1-21-x",
-          "v1-20-x",
-          "v1-19-x",
-          "v1-18-x",
-          "v1-17-x",
-          "v1-16-x",
-          "v1-15-x",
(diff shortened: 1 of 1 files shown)


covalentbond:
@shortcuts I think @nannanli forgot to mention since we migrated our site generator from Vuepress to Docusaurus. We might need to change the whole template as well?

Let me make a PR with updated config.

PR #4339, opened 2021-07-08
Update configs/coralproject.json

wyattjoh:
# Pull request motivation(s)

Update the `configs/coralproject.json` to reflect our new HTML structure.

Showing old search results that are based on an old set of configuration.

Use the new configuration to show the new search results.

**change?**

The PR documentation was helpful!


wyattjoh:
As a note, this was modified to match the `configs/docusaurus-2.json` as we're using the same documentation generator and theme.

PR #4341, opened 2021-07-09
refactor: Update zowe.json

covalentbond:
### Current behaviour
Correcting the changes done in #4332. We forgot to mention that we migrate the site generator from Vuepress to Docusaurus v2 alongside changing the naming convention of versions.

This PR updates the Zowe config as per [Docusaurus v2 template](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json).

##### NB2: Any other feedback / questions ? N/A
Does the `conversation_id` & `nb_hits` needs to be changed?

### DIFF
--- a/configs/zowe.json
+++ b/configs/zowe.json
@@ -1,28 +1,7 @@
 {
   "index_name": "zowe",
   "start_urls": [
-    {
-      "url": "https://docs.zowe.org/(?P<version>.*?)/",
-      "variables": {
-         "version": [
-          "stable",
-          "active-development",
-          "v1.21.x",
-          "v1.20.x",
-          "v1.19.x",
-          "v1.18.x",
-          "v1.17.x",
-          "v1.16.x",
-          "v1.15.x",
-          "v1.14.x",
-          "v1.13.x",
-          "v1.12.x",
-          "v1.11.x",
-          "v1.10.x",
-          "v1.9.x"
-        ]
-      }
-    }
+    "https://docs.zowe.org/"
   ],
   "sitemap_urls": [
     "https://docs.zowe.org/sitemap.xml"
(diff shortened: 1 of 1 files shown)


shortcuts:
Good catch! Looks good :)

covalentbond:
@shortcuts on https://docs.zowe.org/, while searching - the results are shown in multiple version. Is it something, we need to do?

Screenshot 2021-07-09 at 12 53 44 PM

Or Algolia will re-index the site anytime in the day today. That probably will fix this? Thank you!

shortcuts:
The index has been re-indexed 28 minutes ago, I'd suggest you to check the [contextualSearch](https://docusaurus.io/docs/search#contextual-search) option

covalentbond:
I think we have contextualSearch as `true` only. Can you please take a look in [docusaurus.config.js](https://github.com/zowe/docs-site/blob/master/docusaurus.config.js#L168) file.

covalentbond:
Hey, I think I found the problem. We are having two DocSearch components. One in navbar & one on landing page. The DocSearch at navbar seem to work fine.

So I think we will also need to pass the `contextualSearch` as a prop from that [DocSearch component](            <DocSearch) imported in homepage as well?

shortcuts:
Oh I see! `contextualSearch` is a docusaurus option not DocSearch. I'd suggest you to head to their repository for that kind of implementation detail, as they might have a better idea.

Otherwise, since this is the landing page, you can default it to the latest version by adding those options to the DocSearch component

```ts
<DocSearch
  apiKey="59ff39ed48d0820010c7e09fc4b677bf"
  indexName="zowe"
  searchParameters={{
    facetFilters: ["version:YOUR_VERSION", "language:YOUR_LANGUAGE"]
  }}
/>
```

covalentbond:
Hey how to define `language`?

PR #4342, opened 2021-07-09
Attempt to fix scanpy docs

### LINKED ISSUE #4306: AlgoliaSearchError: Index scanpy does not exist
# Do you want to request a **feature** or report a **bug**?

`index_name`= scanpy

Code search doesn't seem to be working, e.g. nothing happens as I type a query. At each keystroke, I get an error in the console:

```
Failed to load resource: https://bh4d9od16a-dsn.algolia.net/1/indexes/*/queries?x-algolia-agent=Algolia%20for%20vanilla%20JavaScript%20(lite)%203.30.0%3Bdocsearch.js%202.6.3&x-algolia-application-id=BH4D9OD16A&x-algolia-api-key=fa4304eb95d2134997e3729553a674b2
 the server responded with a status of 404 (Not Found)
```

This seems similar to https://github.com/algolia/docsearch-configs/issues/4277

I would expect searching to work, and yield results.

#### What have you tried to solve it?

I've mostly googled around for similar issues. Since this was working previously, and I don't think we touched anything around this, I'm not expecting this to be on our side.

#### Any quick clues?

I'm not sure how long it's been broken. I only noticed when we were making the release that had this change and noticed the builds didn't have search working.

ivirshup:
# Pull request motivation(s)

Attempt to fix #4306

I noticed the javascript for getting the versions in the config was receiving version strings like:

```
["↵          latest↵        ", "↵          stable↵        ", "↵          docsearch↵        "] 
```

Which could be causing a problem. So I've `.trim()`-ed those.

I've also implemented the suggestion from @shortcuts (https://github.com/algolia/docsearch-configs/issues/4306#issuecomment-873855656) to modify our `start_urls`

scanpy is getting a 404 when we make API requests

Working search.

### DIFF
--- a/configs/scanpy.json
+++ b/configs/scanpy.json
@@ -9,6 +9,9 @@
           "js": "let versions=[]; for (let b of document.querySelectorAll('.rst-other-versions dl:first-child dd')){versions.push(b.innerText.trim())};return JSON.stringify(versions)"
         }
       }
+    },
+    {
+      "url": "https://scanpy.readthedocs.io/"
     }
   ],
   "sitemap_urls": [
--- a/configs/scanpy.json
+++ b/configs/scanpy.json
@@ -6,7 +6,7 @@
       "variables": {
         "version": {
           "url": "https://scanpy.readthedocs.io/",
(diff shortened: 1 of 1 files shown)


shortcuts:
Looks good, I've removed the `js_render` option since it doesn't seem to change anything

ivirshup:
Looks like they're working now. Thanks!

PR #4343, opened 2021-07-09
Update trantor.json

ssaylo:
Hello, I have modified some trantor settings, please help me merge it.
Appreciate.

### DIFF
--- a/configs/trantor.json
+++ b/configs/trantor.json
@@ -2,9 +2,6 @@
   "index_name": "trantor",
   "start_urls": [
     {
-      "url": "https://trantor-docs-dev.app.terminus.io/"
-    },
-    {
       "url": "https://trantor-docs-dev.app.terminus.io/(?P<version>.*?)/doc",
       "variables": {
         "version": {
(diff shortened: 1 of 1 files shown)


configs/trantor.json
```diff
@@ -2,10 +2,7 @@
   "index_name": "trantor",
   "start_urls": [
     {
-      "url": "https://trantor-docs-dev.app.terminus.io/"
```
shortcuts, line 5: I'd suggest you to keep this URL **at the end** of the `start_urls` array, we usually do that to provide a fallback in case there's an issue with the `version` redirection.

ssaylo: Now the versions is ['v0.14.x', 'v0.17.x'], ant the ```https://trantor-docs-dev.app.terminus.io/```  will be redircted to the default version v0.17.x, it's nothing.
But now I don't have the v0.14.x data because I've changed my website by mistake, can you help me scaper it one times manually?

ssaylo: Just scrapper the website one time manually, because now my website doesn't have any records about v0.14.x, please help me. Appreciate.

shortcuts: I've tried locally and it now works, a new crawl will be queued after the PR merge, you'll see the changes in the next 24 hours!

ssaylo: Appreciate!

ssaylo: By the way, I'm also a HHKB fans...
I have two HHKB, one pro types, one bluetooth.

ssaylo: Hello, the search engine of my website suddenly can't retrieve the data of 0.17.x.
This happened every few several weekends... Is anything woring with it ???

PR #4358, opened 2021-07-12
[Camunda] Add docusaurus_tag to attributesForFaceting

akeller:
# Pull request motivation(s)
Contextual search config yields no results. We would like to modify our search experience to include searching that's limited to the version you are searching in, and without contextual search, we do not get this experience.

https://github.com/camunda-cloud/camunda-cloud-documentation/issues/309

After implementing contextual search config, no search results are returned in the UI

Search results should be returned with contextual search config enabled. 

IMO this is a bug for us because we have a poor search experience at this time with contextual search config enabled.

### DIFF
--- a/configs/camunda.json
+++ b/configs/camunda.json
@@ -8,9 +8,10 @@
   ],
   "sitemap_alternate_links": true,
   "stop_urls": [],
-  "selectors": {
+ "selectors": {
     "lvl0": {
-      "selector": ".menu__link--sublist.menu__link--active",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, I'd also suggest you to copy everything from the [Docusaurus v2 config file](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json) from the `selectors` (included) to the `conversation_id` (excluded) to improve the search results :D

akeller:
@shortcuts double-check me here, I think I did it right 😄

shortcuts:
Looks good!

akeller:
Thanks! You are so quick!

PR #4359, opened 2021-07-12
edit custom_settings

junruchen:
edit custom_settings


configs/mogdb.json
```diff
@@ -1,19 +1,20 @@
 {
   "index_name": "mogdb",
-  "start_urls": [
-    "https://docs.mogdb.io/v2.0.0/overview",
-    "https://docs.mogdb.io/"
```
shortcuts, line 5: I'd suggest you to at least keep the line above in the `start_urls` array

junruchen: In the document: DocSearch will try to read URLs from your sitemap(s) instead of following every link of your starts_urls.
So I remove the start_urls array.
If it useful, I will keep it.

shortcuts: It's indeed optional, but we plan to require the `start_urls` array in a future upgrade of our crawler so it will ease the migration!

junruchen: ok, thanks, I will add it soon

PR #4399, opened 2021-07-23
Update Apollo config to properly omit non-current docset versions

StephenBarlow:


configs/apollodata.json
```diff
@@ -35,14 +35,36 @@
...
+      }
+    },
+    {
+      "url": "https://www.apollographql.com/docs/android/(?P<version>.*?)/",
+      "variables": {
+        "version": [
+          "v3"
+        ]
+      }
+    }
```
shortcuts: Hey @StephenBarlow,

Sorry but this syntax only works in the `start_urls`, in the `stop_urls` only inlined regexes are allowed. For this case, inlining the variable seems enough.

I've updated it here: https://github.com/algolia/docsearch-configs/pull/4399/commits/3f7ecf11cf93ce3631c921cfb0d38bbe8c4b99d3

StephenBarlow: Thanks @shortcuts!

PR #4409, opened 2021-07-27
fix: add new verdaccio docusaurus 2 setting

semoal:
# Pull request motivation(s)

We're migration Verdaccio to Docusaurus 2, and the current config is deprecated

Should work properly

### DIFF
--- a/configs/verdaccio.json
+++ b/configs/verdaccio.json
@@ -1,8 +1,7 @@
 {
   "index_name": "verdaccio",
   "start_urls": [
-    "https://verdaccio.org/docs/",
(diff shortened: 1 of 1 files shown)


configs/verdaccio.json
```diff
@@ -11,29 +10,35 @@
...
+      "anchor",
+      "url",
+      "url_without_anchor",
+      "type"
     ]
-  },
-  "conversation_id": [
-    "1273018287"
-  ],
-  "nb_hits": 6289
```
shortcuts, line 38: these two should still be part of the config

PR #4422, opened 2021-07-30
updated selectors to work with docusaurus v2

jscyo:
# Pull request motivation(s)

We just migrated our docs to docusaurus v2, this resulted in bad search results. This PR is to update the selectors to work with the v2 docs.

The current search gives blank search results

Search results should not give empty results


shortcuts:
Hey, in order to make it work with Docusaurus v2 you need to also add the `custom_settings` of the docusaurus-2.json config

Issue #4364, opened 2021-07-15
[activeui] searching in the docs yields no results.


messaooudi:
Following up on this [issue](https://github.com/algolia/docsearch-configs/issues/4277).

### Current behaviour :
Connect https://activeviam.com/activeui/documentation/5.0.0/, type something in the searchbar.
It will yield empty results.

### Expected behaviour:
The search should yield results.

I am wondering what's the crawl status on 5.0.0 version ?
Any help would be appreciated.

messaooudi:
@shortcuts could you please check if the `5.0.0 version` was crawled ?

shortcuts:
Hey @messaooudi,

The last crawl ended around ~21hours ago but no `5.0.0` version were crawled, only `4.x.x` versions.

I'll try to run the crawl locally to see if it's picked up by the sitemap, worst case would be to add the `5.0.0` sitemap to the `sitemap_urls` in your config.

messaooudi:
Hey @shortcuts, thanks for your reply, 

Our [sitemap_urls](https://github.com/algolia/docsearch-configs/blob/9454e2adbc8acfddae32a8e0bf7e1682254c0a46/configs/activeui.json#L4) contains one sitemap URL, in which we specify a sitemap URL for each version (4.x.x and 5.x.x).

FYI : in the `4.x.x` versions we were using `docusaurus 1` , and in the `5.0.0` version we upgraded to `docusaurus 2`

shortcuts:
> FYI : in the 4.x.x versions we were using docusaurus 1 , and in the 5.0.0 version we upgraded to docusaurus 2

You should update your config to use the [selectors_key](https://docsearch.algolia.com/docs/config-file/#selectors_key-tailor-your-selectors) option then, in order to make the URLs matching the `5.x.x` pattern target the [docusaurus v2 selectors](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json#L11-L25). You can also add all of the [docusaurus v2 settings](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json#L26-L43) to the config (no need to keep the ones from the V1) so that you can take advantage of the [contextualSearch option](https://docusaurus.io/docs/search#contextual-search) on your v5 doc :D

messaooudi:
[...]

PR #4426, opened 2021-07-30
Update ActiveUi config.

messaooudi:
### The problem: 
After upgrading to `Docusaurus 2` from `Docusaurus 1`, the search doesn't yield any results.

### The solution:
Update our [selectors](https://github.com/algolia/docsearch-configs/blob/9454e2adbc8acfddae32a8e0bf7e1682254c0a46/configs/activeui.json#L12-L25) config to match the one from [docusaurus-2](https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json#L11-L25).

See #4364 for more details.

### DIFF
--- a/configs/activeui.json
+++ b/configs/activeui.json
@@ -1,39 +1,29 @@
 {
   "index_name": "activeui",
-  "sitemap_urls": [
-    "https://activeviam.com/activeui/documentation/sitemap.xml"
-  ],
-  "start_urls": [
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, thanks for the PR.

Don't you want to keep the search for the V4 open? If so, I can update the config accordingly.

shortcuts:
The above fix should be enough to make the config works for both v4 and v5

PR #4434, opened 2021-08-02
Update ActiveUi config.

messaooudi:
Following up on #4426

Our doc does not follow the  [directory-structure](https://docusaurus.io/docs/versioning#directory-structure) for versioning, as a result we don't need the **docusaurus_tag** `attributesForFaceting` , we set our custom `activeui-version` meta tag to manage versioning.

I did check this config locally.

### DIFF
--- a/configs/activeui.json
+++ b/configs/activeui.json
@@ -1,10 +1,7 @@
 {
   "index_name": "activeui",
   "sitemap_urls": ["https://activeviam.com/activeui/documentation/sitemap.xml"],
-  "start_urls": [
-    "https://activeviam.com/activeui/",
-    "https://activeviam.com/activeui/documentation/5.0.0/index.html"
-  ],
+  "start_urls": ["https://activeviam.com/activeui/"],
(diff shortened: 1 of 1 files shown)


messaooudi:
@shortcuts any chances you can take a look at this today ? 
Thanks 👍🏻

shortcuts:
Hey @messaooudi, will you use the `contextualSearch` option provided by docusaurus? `docusaurus_tag` will be used if so

messaooudi:
@shortcuts no, we are not using the `contextualSearch`.

messaooudi:
@shortcuts can you re-trigger a crawl ?

shortcuts:
It's triggered after the merge. Note that, due to the size of the index, it could take a few hours.

messaooudi:
@shortcuts ok, thanks. Once the crawl is completed , would you please share the **number of total records** , the **number of records relative to 5.0.0** , and if possible, does the _5.0.0_ records have the `activeui-version` field set to `5.0.0`

PR #4470, opened 2021-08-11
Add custom settings for faceting

tjerman:
# Pull request motivation(s)

We wish to use meta tags to support facet filtering on our documentation page (http://docs.cortezaproject.org/)

Our documentation aggregates most of our versions into a single documentation site.
When searching, the results refer to all versions without any separation between versions.

(the search bar is temporarily removed, but the meta tags are still there)

When I search for something, it should either be categorized by the version or filtered to the current version.
An email exchange suggested we use meta tags and facet filters from our FE.

Feature

### DIFF
--- a/configs/cortezaproject.json
+++ b/configs/cortezaproject.json
@@ -19,5 +19,8 @@
   "conversation_id": [
     "1576826212"
   ],
+  "custom_settings": {
+    "attributesForFaceting": ["version"]
+  },
   "nb_hits": 62880
 }


tjerman:

At what time does your system re-index sites? In the documentation it states every 24h, but what time?
Are we able to request the site to be reindexed manually after such config changes?

configs/cortezaproject.json
```diff
@@ -19,5 +19,8 @@
   "conversation_id": [
     "1576826212"
   ],
+  "custom_settings": {
+    "attributesForFaceting": ["version"]
```
shortcuts, line 23: It seems you also have `language` and `module`, wouldn't you also use them as `attributesForFaceting`?

shortcuts:
> ##### NB2: Any other feedback / questions ?
> At what time does your system re-index sites? In the documentation it states every 24h, but what time?
> Are we able to request the site to be reindexed manually after such config changes?

A crawl is queued after a PR merge

configs/cortezaproject.json
tjerman: I added those extra two because I thought I would also like to have them, but I'll probably remove them for now

shortcuts: Sounds good!

tjerman:
@shortcuts The facet filtering is now working, thanks for your help.

I have another question if you could please:
Our documentation page has multiple guides, one for each role, where some sections may refer to the same part of the system with similar sections.

For example, when searching for testing, I get N results where it may not be obvious what guide that result refers to.
What I would like to see is something in the lines of what Antora does with their docs (https://docs.antora.org/antora/2.3/) where the results are categorized by version. How can I do that but with an arbitrary attribute, such as the guide name?

shortcuts:
@tjerman you could use the [selectors_key](https://docsearch.algolia.com/docs/config-file/#selectors_key-tailor-your-selectors) option to create a new selector for the `/guides` endpoints. In this new selector, you need to make the `lvl0` match the name of the guide (e.g. `.guideTitle h1`).

tjerman:
Alright, I'll see what I can do.
Thanks for your help

PR #4485, opened 2021-08-14
Update rockset.json

meitros:
we moved our docs site to rockset.com/docs

# Pull request motivation(s)


meitros:
The commit should have my work email `mihir@rockset.com` associated with it

PR #4498, opened 2021-08-18
fix(ballerina): index sub library

iamsameeraliyanage:
Revert to default state

# Pull request motivation(s)

### DIFF
--- a/configs/ballerina.json
+++ b/configs/ballerina.json
@@ -1,19 +1,7 @@
 {
   "index_name": "ballerina",
   "start_urls": [
-    {
-      "url": "https://ballerina.io/learn",
-      "page_rank": 1
-    },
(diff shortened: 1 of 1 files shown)


iamsameeraliyanage:
Hi,
I wanted a result, the search text with "Swanlake API" and when it clicks I need to redirect it to "https://lib.ballerina.io/" url, which is a subdomain of this site. can I do that?

shortcuts:
> can I do that?

Yep, this commit https://github.com/algolia/docsearch-configs/pull/4498/commits/80465cad650fd284f4370842e0c04ef6b61a461c should do what you want to achieve!

iamsameeraliyanage:
I think it kind of fixed that.

PR #4503, opened 2021-08-19
fix(configs): delete start_urls

SCAU-zh:

I am the developer of CODING. I add sitemap.xml config so that I  delete start_urls now.

### DIFF
--- a/configs/coding_help.json
+++ b/configs/coding_help.json
@@ -1,9 +1,6 @@
 {
   "index_name": "coding_help",
-  "start_urls": [
-    "https://help.coding.net/docs/start/new.html",
-    "https://help.coding.net/docs/"
-  ],
+  "start_urls": [],
   "stop_urls": [],
   "selectors": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey @SCAU-zh,

You can keep the `start_urls`, as the `sitemap` will always be fetched anyway. Having `start_urls` allows a better "starting point" for our crawlers :D

SCAU-zh:
> Hey @SCAU-zh,
> 
> You can keep the `start_urls`, as the `sitemap` will always be fetched anyway. Having `start_urls` allows a better "starting point" for our crawlers :D

Hello！

I pulled the crawler image and run my own crawler.

Whether to keep this field will affect my result indices.

Keeping `start_urls` will result in outdated content. So I need to delete it.

shortcuts:
Hey, we are getting `14039` records with and without `start_urls`, so it won't affect your results removing it

PR #4508, opened 2021-08-20
Update other domains

iamsameeraliyanage:
Add https://lib.ballerina.io , https://blog.ballerina.io, https://central.ballerina.io, https://play.ballerina.io domains

# Pull request motivation(s)


iamsameeraliyanage:
Hi,
I don't understand why this kind of window showing instead of no result. 
image

I checked network response and it has result. But there are some null value. How i can fix that?

shortcuts:
Hey @iamsameeraliyanage,

When clicking on the results from the above screenshots, you can see that it redirects to the `/learn` endpoint, which might have different selectors to what you've stated in the actual ones. 

You could create a new selector group for this endpoint specifically to solve this issue

configs/ballerina.json
```diff
@@ -1,52 +1,142 @@
...
+            "selectors_key": "ballerina_playground",
+            "tags": [
+                "ballerina_playground"
+            ],
+            "extra_attributes": {
+                "site": [
+                    "ballerina_playground"
+                ]
+            }
+        }
```
shortcuts, line 67: As this is interactive content, it should exists in the config since there's nothing to crawl on this page. Plus, when searching for `playground`, it already redirects to this page: https://ballerina.io/learn/user-guide/getting-started/setting-up-ballerina/#trying-ballerina-in-the-playground

iamsameeraliyanage: @Clément Vannicatte You know what best. Please add that.

iamsameeraliyanage:
@shortcuts can you merge this?

configs/ballerina.json
```diff
@@ -1,52 +1,142 @@
...
+                "selector": "",
+                "default_value": "Ballerina Playground"
+            },
+            "lvl1": "#container h1",
+            "lvl2": "#container h2",
+            "lvl3": "#container h3",
+            "lvl4": "#container h4",
+            "lvl5": "#container h5,#container h6",
+            "text": "#container p, #container li"
+        }
```
shortcuts, line 130:
```suggestion
        }
```

configs/ballerina.json
shortcuts:
```suggestion
```

configs/ballerina.json
```diff
@@ -1,52 +1,117 @@
...
+            "selectors_key": "ballerina_central",
+            "tags": [
+                "ballerina_central"
+            ],
+            "extra_attributes": {
+                "site": [
+                    "ballerina_central"
+                ]
+            }
+        },
```
shortcuts, line 54:
```suggestion
        }
```

PR #4513, opened 2021-08-21
Update mspc.json

VJalili:
The Algolia search on [MSPC website](https://genometric.github.io/MSPC/) does not return any hits (see the screenshot below), in an effort to fix it, updating the search configurations following [algolia docs](https://docsearch.algolia.com/docs/config-file/). 

Screen Shot 2021-08-21 at 1 58 34 PM

### DIFF
--- a/configs/mspc.json
+++ b/configs/mspc.json
@@ -1,15 +1,23 @@
 {
   "index_name": "mspc",
   "start_urls": [
-    "https://genometric.github.io/MSPC/"
+    {
+      "url": "https://genometric.github.io/MSPC/docs/",
+      "selectors_key": "docs"
+    },
+    {
+      "url": "https://genometric.github.io/MSPC/"
+    }
   ],
   "sitemap_urls": [
     "https://genometric.github.io/MSPC/sitemap.xml"
   ],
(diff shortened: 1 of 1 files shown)


configs/mspc.json
```diff
@@ -41,8 +49,5 @@
       "type"
     ]
   },
-  "conversation_id": [
-    "1181088484"
-  ],
```
shortcuts, line 46: Could you please restore the `conversation_id`?

shortcuts (commented):
Let me know if there's still no results after the merge

configs/mspc.json
VJalili: Sure, though I am not sure what is the `conversation_id` and how it is set, is the current value correct? (i.e., `1181088484`)

shortcuts: Hey, yes it is the correct value. We use it internally to match a config to its owner when they apply for DocSearch.

VJalili: Thanks! I reverted that change.

VJalili:
@shortcuts Thank you for merging this; though I am still getting no results.

PR #4526, opened 2021-08-25
Update medusa-commerce.json

kasperkristensen:
# Pull request motivation(s)

Our old config did not provide the expected search results, this should fix it (tested using a local instance of your crawler)


olivermrbl:
Would be great to get this in asap :) @shortcuts We are currently running DocSearch in production, but the indexing are not done correctly atm. Hope you understand :)

PR #4541, opened 2021-08-29
Update zhinst.json

tobiasah:
We really love working with docsearch. Since we updated our documentation we need to update our config a bit:

* Add `variables` for newly added documentations.
* Add a second `start_url`
* Correct a small spelling mistake

### DIFF
--- a/configs/zhinst.json
+++ b/configs/zhinst.json
@@ -5,10 +5,30 @@
       "url": "https://docs.zhinst.com/(?P<project>.*?)/",
       "variables": {
         "project": [
+          "hdawg_user_manual",
+          "hf2_user_manual",
           "labone_programming_manual",
-          "shfqa_user_manual"
+          "mfia_user_manual",
+          "mfli_user_manual",
+          "pqsc_user_manual",
+          "shfqa_user_manual",
+          "shfsg_user_manual",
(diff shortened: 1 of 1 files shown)


configs/zhinst.json
```diff
@@ -23,13 +43,40 @@
...
       },
       "lvl1": ".doc h1",
       "lvl2": ".doc h2",
       "lvl3": ".doc h3",
       "lvl4": ".doc h4",
       "text": ".doc p, .doc td.content, .doc th.tableblock"
+    },
+    "api": {
+      "lvl0": {
+        "selector": "//li[@class='toctree-l1 current active nav-item']/a",
```
shortcuts, line 56: Hey @tobiasah, we plan to deprecate the xPath support and provide an access to a Cheerio instance in the next major update. Do you think you could find a DOM alternative in the meantime? This way, your config will still be close to correct once it has been migrated

tobiasah: Thanks for the input.
I wasn't aware that path will be removed I just used it because I find it a bit more clear. 
I switched the xpath selectors to DOM.

shortcuts: Thanks! We will provide some more informations soon!

PR #4542, opened 2021-08-30
Update medusa-commerce.json

kasperkristensen:
# Pull request motivation(s)

Hi, we faced an issue where selectors were not being used. Only the first specified selector `docs` was used, and not all the expected content that should be crawled as part of the selectors `admin` and `store` was picked up. Changing the order of the start_urls and selectors seems to have fixed this issue when testing with a local instance of your crawler. 

(If simply changing the order is not the best solution, then please let us know!)

### DIFF
--- a/configs/medusa-commerce.json
+++ b/configs/medusa-commerce.json
@@ -2,15 +2,6 @@
   "index_name": "medusa-commerce",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


shortcuts:
Indeed! Sorry, I didn't saw it, it should be fixed after the merge :)

PR #4543, opened 2021-08-30
Update discord.json

typpo:
# Pull request motivation(s)

This selector will capture reference tables that include useful fields and enums :)

Tables like these are not included in search results 
image

Fields and enums are searchable

### DIFF
--- a/configs/discord.json
+++ b/configs/discord.json
@@ -9,7 +9,7 @@
     "lvl3": "[class*='contentWrapperInner'] h3",
     "lvl4": "[class*='contentWrapperInner'] h4",
     "lvl5": "[class*='contentWrapperInner'] h5",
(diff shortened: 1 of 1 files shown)


configs/discord.json
```diff
@@ -9,7 +9,7 @@
     "lvl3": "[class*='contentWrapperInner'] h3",
     "lvl4": "[class*='contentWrapperInner'] h4",
     "lvl5": "[class*='contentWrapperInner'] h5",
-    "text": "[class*='contentWrapperInner'] p, [class*='contentWrapperInner'] li"
+    "text": "[class*='contentWrapperInner'] p, [class*='contentWrapperInner'] li, [class*='contentWrapperInner'] td"
```
shortcuts, line 12: Hey @typpo,

I'd recommend to make the selector more specific, otherwise it will also react records for the `TYPE` for example, which might add a lot of noise to the results.

Limiting it to the `first-child` might be safer.

PR #4545, opened 2021-08-30
Update vuetifyjs.json

nekosaur:
# Pull request motivation(s)
improve indexing results

- switch order of start_urls so that least specific match is last
- add `selectors_exclude`
- add `separatorsToIndex`

### DIFF
--- a/configs/vuetifyjs.json
+++ b/configs/vuetifyjs.json
@@ -2,22 +2,23 @@
   "index_name": "vuetifyjs",
   "start_urls": [
     {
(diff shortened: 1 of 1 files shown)


configs/vuetifyjs.json
```diff
@@ -48,7 +49,8 @@
     }
   },
   "custom_settings": {
-    "attributesForFaceting": ["language"]
+    "attributesForFaceting": ["language"],
```
shortcuts, line 52:
```suggestion
    "attributesForFaceting": ["lang"],
```

It should match the `variables` name

PR #4547, opened 2021-08-31
Update apache_echarts.json

pissang:
Added language faceting

Thanks for the help.

### DIFF
--- a/configs/apache_echarts.json
+++ b/configs/apache_echarts.json
@@ -18,6 +18,9 @@
     "lvl5": ".nuxt-content h6",
     "text": ".nuxt-content p, .nuxt-content li"
   },
+  "custom_settings": {
+    "attributesForFaceting": ["lang"]
+  },
   "conversation_id": [
     "1613538708"
   ],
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, if you'd our scraper to retrieve `language` you need to add [this selector](https://github.com/algolia/docsearch-configs/blob/master/configs/vuepress.json#L33-L37) if available, or directly [from the URLs](https://docsearch.algolia.com/docs/config-file/#using-regular-expressions)

pissang:
Hi, I update it to the "from the URLs" method you suggested.  Could you please help check if it's correct? Thanks a lot!

configs/apache_echarts.json
```diff
@@ -13,9 +18,6 @@
     "lvl5": ".nuxt-content h6",
     "text": ".nuxt-content p, .nuxt-content li"
   },
-  "custom_settings": {
-    "attributesForFaceting": ["language"]
-  },
```
shortcuts: This can be restored with `lang` instead of `language` and it's good!

pissang: Fixed! Thanks!

PR #4548, opened 2021-08-31
Support to search the whole site for `https://skywalking.apache.org/`

Jtrust:
Hello! 
I was wondering if Algolia could support site-wide search after I modify this part of the code.
I would be most grateful if you could give me some help. Thanks!

### DIFF
--- a/configs/skywalking.json
+++ b/configs/skywalking.json
@@ -1,28 +1,66 @@
 {
   "index_name": "skywalking",
   "start_urls": [
+    {
+      "url": "https://skywalking.apache.org/docs/main/(?P<version>.*?)/",
+      "selectors_key": "doc",
+      "variables": {
+        "version": [
+          "latest",
+          "v8.7.0",
+          "v8.6.0",
+          "v8.5.0",
+          "v8.4.0",
+          "v8.3.0",
+          "v8.2.0"
+        ]
+      }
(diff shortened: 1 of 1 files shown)


configs/skywalking.json
```diff
@@ -1,44 +1,13 @@
...
-    },
-    {
-      "url": "https://skywalking.apache.org/docs/skywalking-satellite/(?P<version>.*?)/",
-      "variables": {
-        "version": [
-          "latest",
-          "v0.1.0"
-        ]
-      }
-    }
```
shortcuts, line 34: Hey, what do you mean by `site-wide search`? removing the versions will disable filtering on your website, so when searching you'll have results from all the versions at the same time: [see this section of our documentation](https://docsearch.algolia.com/docs/config-file/#using-regular-expressions)

Jtrust: I mean , at present, only part of the site content is searchable. But not includes `blogs` or  `events` and so on . Could all the content of the site be searched by modify ing the config?

shortcuts: Hey, the fix in https://github.com/algolia/docsearch-configs/pull/4548/commits/4e485efb5398cd48db54d40bc2b310eb5b970954 should do it! Let me know if you find any issues

PR #4549, opened 2021-08-31
update pinia with api

posva:
# Pull request motivation(s)

Adding the API priority

### DIFF
--- a/configs/pinia.json
+++ b/configs/pinia.json
@@ -40,6 +40,14 @@
         "Cookbook"
       ],
       "page_rank": 5
+    },
+    {
+      "url": "https://pinia.esm.dev/api/",
+      "selectors_key": "api",
+      "tags": [
+        "API"
+      ],
+      "page_rank": 4
     }
   ],
   "stop_urls": [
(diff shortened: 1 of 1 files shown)


configs/pinia.json
```diff
@@ -57,11 +65,25 @@
...
+        "default_value": "API"
+      },
+      "lvl1": ".content h1",
+      "lvl2": ".content h2",
+      "lvl3": ".content h3",
+      "lvl4": ".content h4",
+      "lvl5": ".content h5",
+      "lvl6": ".content p, .content li",
+      "text": ".content [class^=language-]"
+    }
```
posva, line 82: Do I need to do this for all selectors? I would like to avoid duplicating all the lvl > 0 selectors since they are the same as the default ones

shortcuts: Hey @posva, yes you need to duplicate them, sorry.

Also, the correct shape should be
```json
{
  "selectors": {
    "default": {
      "lvl0": "",
      // ...
    },
    "api": {
      "lvl0": "",
      // ...
    }
  }  
}

```

posva: Ok, thanks! It would be nice not having to duplicate them in the future. Not sure if that part of the code is open and what language it is to send a PR 😄

shortcuts: It's open source (https://github.com/algolia/docsearch-scraper) and in python, but we are actually working on providing a new infrastructure to DocSearch, stay tuned! :D

posva: oh but can I duplicate it only once and let the default be used for all other selectors? Like the last commit?

shortcuts: Yep exactly, remove the `selectors_key` for everything that will use the `default`, and only specify and duplicate it for the API for example

PR #4555, opened 2021-09-04
Delete brawlupjs.json 

alputv:
# Pull request motivation(s)
I've deleted my guide for https://brawlup.js.org/ and made a new guide. I want to delete brawlupjs index from Algolia and apply for a new one. Thanks for understanding!


shortcuts:
Hey! If it's still related to the same project, no need to delete/re-apply, you can just edit your config

alputv:
> Hey! If it's still related to the same project, no need to delete/re-apply, you can just edit your config

Not the same project :(

PR #4564, opened 2021-09-08
near-sdk-rs docs added to the main index

volovyk-s:
We want to index both https://docs.near.org/ and https://www.near-sdk.io/ on our main documentation website.
The first one is using `Docusaurus` and the second one `Docusaurus 2`. That is why I have added `selectors_key` and some parameters adviced by @shortcuts  (thanks @shortcuts  :)).

### DIFF
--- a/configs/near.json
+++ b/configs/near.json
@@ -1,34 +1,71 @@
 {
   "index_name": "near",
   "start_urls": [
-    "https://docs.near.org/docs/"
+    {
+      "url": "https://docs.near.org/docs/",
+      "selectors_key": "docs"
+    },
+    {
+      "url": "https://near-sdk.io",
+      "selectors_key": "sdk-docs"
+    }
   ],
   "sitemap_urls": [
-    "https://docs.near.org/sitemap.xml"
+    "https://docs.near.org/sitemap.xml",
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
looks good!

volovyk-s:
Hey @shortcuts ! This PR was merged > 24 hours ago, but it looks like https://www.near-sdk.io/ is still not indexed. Should I wait for more, or maybe there is some problem in the config?

shortcuts:
Hey, the crawl has been finished 21 hours ago and went well! Keep in mind that the Algolia DocSearch plugin must be enabled in order to use the `contextualSearch` option on Docusaurus

volovyk-s:
@shortcuts all we have for Algolia is here: https://github.com/near/docs/blob/39dd5baf97459b90c9eb3de64578167c1f6f36f3/website/siteConfig.js#L22
Should we change this config somehow?
Or maybe we should change something in`near-sdk.io` repo [here](https://github.com/near/sdk-docs/)?

shortcuts:
Hey @volovyk-s 

I just realized the URL wasn't correct so our scraper ignored it: https://github.com/algolia/docsearch-configs/pull/4580

This should make the SDK searchable!

PR #4586, opened 2021-09-14
chore: update electronjs configuration

molant:
# Pull request motivation(s)

The main electronjs.org site is migrating to Docusuarus. This PR updates the
configuration to match more closely the beta-electronjs index one so it takes
advantage of the sitemap and new markup.

### DIFF
--- a/configs/electronjs.json
+++ b/configs/electronjs.json
@@ -1,87 +1,47 @@
 {
   "index_name": "electronjs",
   "start_urls": [
-    {
-      "url": "https://www.electronjs.org/docs/api/",
-      "tags": [
-        "api"
-      ],
-      "selectors_key": "api"
-    },
-    {
-      "url": "https://www.electronjs.org/docs/",
-      "tags": [
-        "docs"
-      ],
-      "selectors_key": "docs"
-    },
-    {
-      "url": "https://www.electronjs.org/blog/",
-      "tags": [
-        "blog"
-      ],
-      "selectors_key": "blog"
-    }
+    "https://www.electronjs.org/docs/latest/"
   ],
-  "stop_urls": [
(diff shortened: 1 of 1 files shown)


erickzhao (approved):

configs/electronjs.json
```diff
@@ -1,87 +1,47 @@
...
-      "selectors_key": "docs"
-    },
-    {
-      "url": "https://www.electronjs.org/blog/",
-      "tags": [
-        "blog"
-      ],
-      "selectors_key": "blog"
-    }
+    "https://www.electronjs.org/docs/latest/"
```
shortcuts, line 4: Is there a reason to only target the latest? Are the old versions available?

(Sorry I can't check myself, I'm on mobile)

erickzhao: We're migrating our docs over to Docusaurus. `docs/latest/` is the path for Docusaurus and `/docs/` is currently the path for the old docs. Neither are versioned as of yet. :)

shortcuts: Hey @erickzhao 

Thanks, that's what I thought!

If you'd like to keep the old version searchable, it's possible:
- Use the `tags` to make groups that will be used as `facetFilters` in the frontend, see: https://docsearch.algolia.com/docs/config-file/
- Couple that with specific `selector_keys` to target the right DOM elements, see: https://docsearch.algolia.com/docs/config-file/

molant: Hi @shortcuts!

Not really looking at making the old version searcheable for now. We want to add docs versioning but we are not there yet so indexing `/latest` is what we want 😊

molant:
Thanks @shortcuts!

PR #4598, opened 2021-09-16
Update with new documents

thetis-apps:
Includes all localized documents
Language code added to filename

# Pull request motivation(s)


thetis-apps:
Tried to change name of configuration file.

shortcuts:
Hi, sorry it's a bit more complicated to update file names and index names!

PR #4607, opened 2021-09-16
chore: update nuxtjs.org settings [wip]

danielroe:
# Pull request motivation(s)

We'd like to update the config for https://nuxtjs.org, which is currently not showing any results in search. (I am [a core team member](https://nuxtjs.org/teams/).)

**Note**: I'm currently not able to test this owing to not having a writeable API key (email sent separately to request this).

PR #4619, opened 2021-09-17
fix: update language tags for nuxtjs.org

danielroe:
# Pull request motivation(s)

Follow-up on https://github.com/algolia/docsearch-configs/pull/4607#issuecomment-921628399 

Language codes are coming from https://github.com/nuxt/nuxtjs.org/blob/a2fdf7dd45860e06085715a2ed543a41850f163a/nuxt.config.ts#L151-L168

I've suggested tagging with both iso & code in case we change the frontend behaviour in future. (cc: @Tahul)

### DIFF
--- a/configs/nuxtjs.json
+++ b/configs/nuxtjs.json
@@ -2,12 +2,26 @@
   "index_name": "nuxtjs",
   "start_urls": [
     {
-      "url": "https://nuxtjs.org/(?P<lang>.*?)/",
+      "url": "https://nuxtjs.org/fr/",
       "variables": {
         "lang": [
-          "en",
-          "fr",
-          "ja"
+          "fr-FR"
+        ]
+      }
+    },
+    {
+      "url": "https://nuxtjs.org/ja/",
+      "variables": {
+        "lang": [
+          "ja-JP"
+        ]
+      }
+    },
+    {
+      "url": "https://nuxtjs.org/en/",
+      "variables": {
+        "lang": [
+          "en-US"
         ]
       }
     },
@@ -15,7 +29,7 @@
       "url": "https://nuxtjs.org/",
       "extra_attributes": {
         "lang": [
-          "en"
+          "en-US"
         ]
       }
     }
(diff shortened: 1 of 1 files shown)


configs/nuxtjs.json
```diff
@@ -2,19 +2,37 @@
   "index_name": "nuxtjs",
   "start_urls": [
     {
-      "url": "https://nuxtjs.org/(?P<lang>.*?)/",
+      "url": "https://nuxtjs.org/fr/",
       "variables": {
         "lang": [
```
shortcuts, line 7: I'd suggest using `lang` instead of `language` so that you don't have any changes to do on the nuxtjs doc!

danielroe: thanks :heart:

shortcuts: @danielroe https://github.com/algolia/docsearch-configs/pull/4619/commits/62ca71b9a9af23bde72e6e5f6cc457f1658acd74 should do it! 

For future updates, our scraper also supports `docsearch:` meta tags: https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags

Or selecting the `lang` from the `html` element ([like this](https://github.com/algolia/docsearch-configs/blob/master/configs/vuejs-v3.json#L22-L26)) but it seems to be broken on https://nuxtjs.org/fr/ right now, not sure if you're aware of it!

danielroe: Amazing - so that sounds like it should be possible to omit the list of start_urls + extra_attributes entirely and simply search via language code instead (for example, `facetFilters: ["lang:en"]`)?

shortcuts: Exactly, you'd only have `https://nuxtjs.org/` in the `start_urls` and the i18n would be handled by our scraper either by the meta tags or the `lang` selector!

To avoid any changes in your documentation, [this selector](https://github.com/algolia/docsearch-configs/blob/master/configs/vuejs-v3.json#L22) can be called `language` so you'd keep the actual `facetFilters: ["language:en"]`

PR #4628, opened 2021-09-20
Exclude sdk/php pages from index

obaybakova:
# Pull request motivation(s)

We need to exclude pages under /sdk/php from the search results.

When we search for "php sdk", we see pages in the section in the search results.

When we search for "php sdk", we find nothing.

### DIFF
--- a/configs/xsolla.json
+++ b/configs/xsolla.json
@@ -82,7 +82,7 @@
   "stop_urls": [
     "/search",
     "/changelog/tags",
-    "https://developers.xsolla.com(/[a-z][a-z])?/sdk/php/*",
+    "/sdk/php/",
     "https://developers.xsolla.com/changelog/$",
     "https://developers.xsolla.com/ru/changelog/$",
     "https://developers.xsolla.com/de/changelog/$",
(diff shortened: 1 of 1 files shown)


configs/xsolla.json
```diff
@@ -82,6 +82,7 @@
   "stop_urls": [
     "/search",
     "/changelog/tags",
+    "https://developers.xsolla.com(/[a-z][a-z])?/sdk/php/*",
```
shortcuts, line 85:
```suggestion
    "/sdk/php/",
```

If the goal is to remove any pages under the `/sdk/php/`, this will do it!

obaybakova: We also have a locale identifier right after the domain like 
https://developers.xsolla.com/de/sdk/php/
https://developers.xsolla.com/ja/sdk/php/
https://developers.xsolla.com/ko/sdk/php/

Should it be `(/[a-z][a-z])?/sdk/php/` then?

shortcuts: `stop_urls` checks matching patterns, meaning as long as `/sdk/php/` is in the URL, it will be ignored

obaybakova: Got it, thanks!

PR #4631, opened 2021-09-20
fix/revolut: update config

ismayilmusayev:
# Pull request motivation(s)

Search does not include Redocusaurus plugin content


shortcuts:
hey @ismayilmusayev, it seems that client-side rendering is required in order to crawl the redocusaurus plugin, this commit should fix it https://github.com/algolia/docsearch-configs/pull/4631/commits/dbf79261756faa02c22948e9f506d8ddb9503a5b

PR #4636, opened 2021-09-21
fix(httpie): only main docs page should be parsed

bartekjacak:
# Pull request motivation(s)

Crawler was parsing all the https://httpie.io/docs/* pages because they were linked inside the main page.

### What is the current and expected behaviour?

It should only do that for the main `https://httpie.io/docs` page.

### Question:

We also experience the following issue:

image

Whatever we search for, all the suggestions are always the same. What's interesting, links seem to be correct:

CleanShot 2021-09-21 at 12 40 56

CleanShot 2021-09-21 at 12 41 31

📸   Live demo at https://httpie.io/.

### DIFF
--- a/configs/httpie.json
+++ b/configs/httpie.json
@@ -1,11 +1,6 @@
 {
   "index_name": "httpie",
-  "start_urls": ["https://httpie.io/docs"],
-  "stop_urls": [
-    {
-      "url": "https://httpie.io/docs/.+"
-    }
-  ],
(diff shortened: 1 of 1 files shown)


shortcuts:
hey @bartekjacak, please let me know if my last commit makes the result better once merged :D

bartekjacak:
It works right now – thank you @shortcuts!

PR #4641, opened 2021-09-22
px4.json: add sitemap.xml

hamishwillee:
This adds a sitemap.xml for the whole site, which I hope will fix our inconsistent search result issue.

If this makes a difference I will do the work to make this more modular. Note, this only specifies URLs and update frequency.

Also tries to add selector for our parameter docs which are often in tables. 

As advised by ClementV. Thanks!

EDIT: NOTE, moved to draft. The auto-generated sitemap is a bit crap. let me have another go.

### DIFF
--- a/configs/px4.json
+++ b/configs/px4.json
@@ -1,5 +1,6 @@
 {
   "index_name": "px4",
+  "sitemap_urls": ["https://docs.px4.io/sitemap.xml"],
   "start_urls": [
     {
       "url": "https://docs.px4.io/master/",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey @hamishwillee, feel free to ping me when it's ready!

hamishwillee:
Thanks @shortcuts - I "think" we're good to go :-)
Is it _likely_ that if we merge now, in 12 hours I will be able to test the effects?

shortcuts:
> Is it likely that if we merge now, in 12 hours I will be able to test the effects?

The crawl is queued after the merge, but px4 is a pretty big index so it should be! Feel free to ping me if not, I'll tell you how far the process is

hamishwillee:
Thanks very much - really appreciate it. I'm hoping this just make it work. Search is simultaneously one of the very most important things in the site, and something I never want to think about :-)

PR #4654, opened 2021-09-26
chore: update ignite config

ydcjeff:
# Pull request motivation(s)

PyTorch-Ignite now has own website: https://pytorch-ignite.ai which has how to guides and tutorials. We want Docsearch to work on that site.

Search only applied to https://pytorch.org/ignite

Search should show results from https://pytorch-ignite.ai

Feature

Not sure the updated config is correct. If you could check, that would be great. Thanks!

### DIFF
--- a/configs/ignite.json
+++ b/configs/ignite.json
@@ -9,22 +9,42 @@
           "js": "let versions=[]; for (let b of document.querySelectorAll('.rst-other-versions dd')){versions.push(b.innerText)};return JSON.stringify(versions)"
         }
       }
+    },
+    {
+      "url": "https://pytorch-ignite.ai/",
+      "selectors_key": "pytorch-ignite-ai"
     }
   ],
   "stop_urls": ["/_", "\\?", "index.html"],
   "selectors": {
-    "lvl0": {
-      "selector": ".rst-content h1",
-      "global": true,
-      "default_value": "Documentation"
+    "default": {
+      "lvl0": {
+        "selector": ".rst-content h1",
+        "global": true,
+        "default_value": "Documentation"
+      },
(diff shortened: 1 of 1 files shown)


ydcjeff:
@shortcuts Can you please review this? Thanks!

configs/ignite.json
```diff
@@ -9,22 +9,42 @@
           "js": "let versions=[]; for (let b of document.querySelectorAll('.rst-other-versions dd')){versions.push(b.innerText)};return JSON.stringify(versions)"
         }
       }
+    },
+    {
+      "url": "https://pytorch-ignite.ai/",
+      "selectors_key": "pytorch-ignite-ai"
     }
```
shortcuts, line 16: You will see results from both websites on both search instances with this setup. If you'd like to limit the search to its website, you can use [`tags`](https://docsearch.algolia.com/docs/config-file/#using-custom-tags) for each URL group, and then in your frontend have:

on https://pytorch.org/ignite/
```jsx
<DocSearch
  // ... your credentials
  searchParameters={{
    facetFilters: ['tags:ignite']
  }}
/>
```

and on https://pytorch-ignite.ai/
```jsx
<DocSearch
  // ... your credentials
  searchParameters={{
    facetFilters: ['tags:ignite-ai']
  }}
/>
```

shortcuts: concrete example

config: https://gist.github.com/shortcuts/b4b4e5ae8219a5e2496eb5fab2d2e1f6
live demo: https://codesandbox.io/s/bold-kate-g35nd?file=/src/App.js

ydcjeff:
Thanks @shortcuts . I have updated the config.

PR #4659, opened 2021-09-28
Update ignite.json

ydcjeff:
# Pull request motivation(s)

Search results on https://pytorch-ignite.ai shows docsearch hit title as the h1 extracted title.
What we want is if the search results are under the sidebar items, we want to show sidebar title like `Tutorials` and `How-to guides`. If the search results are not under the sidebar items, we want to show the h1 title instead.

For examples, searching `how to` should show `How-to guides` hit title and search results containing `how-to`s.
And searching `ecosystem` should show `Ecosystem` hit title and ecosystem search results.

H1 title is shown.

Sidebar title to be shown.

Feature

I am not sure if the config is correct. Let me know for additional required changes.

Sorry for multiple frequent PRs. Thanks!

@shortcuts

### DIFF
--- a/configs/ignite.json
+++ b/configs/ignite.json
@@ -35,7 +35,7 @@
     },
     "pytorch-ignite-ai": {
       "lvl0": {
-        "selector": ".prose h1",
+        "selector": "nav > h5, .prose h1",
         "global": true,
         "default_value": "Documentation"
       },
--- a/configs/ignite.json
+++ b/configs/ignite.json
@@ -35,7 +35,7 @@
     },
     "pytorch-ignite-ai": {
       "lvl0": {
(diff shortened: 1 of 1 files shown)


configs/ignite.json
```diff
@@ -35,7 +35,7 @@
     },
     "pytorch-ignite-ai": {
       "lvl0": {
-        "selector": ".prose h1",
+        "selector": "nav > h5, .prose h1",
```
shortcuts, line 38: Our scraper acts as `document.querySelectorAll(...)`, which in this case will select multiple `h1`s.

The best would be to add an `.active` class (or something similar) when a children is selected so we can target them directly

PR #4691, opened 2021-10-04
Update thespaghettidetective.json

dipuzyrev:
Update selectors to categorize blog posts in the search box

### DIFF
--- a/configs/thespaghettidetective.json
+++ b/configs/thespaghettidetective.json
@@ -12,7 +12,7 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, we are actually migrating DocSearch infrastructure which will deprecate xPath selectors. Existing selectors will be translated but new one skipped.

I'd suggest your to use DOM selectors to retrieve the blog categories, and target the `/blog` endpoint to its own [`selectors_key`](https://docsearch.algolia.com/docs/config-file/#selectors_key-tailor-your-selectors)

dipuzyrev:
Thank you very much for your help! In the new update `/blog` endpoint has it's own `selectors_key` with DOM selector instead of xPath.

PR #4696, opened 2021-10-05
Update domain for Nanoc index

ddfreyne:
# Pull request motivation(s)

Nanoc has moved from https://nanoc.ws to https://nanoc.app.

The search on https://nanoc.app is not functioning at the moment, likely because this config points to the old https://nanoc.ws instead.

Search is working as intended.


ddfreyne:
Thanks for the quick response @shortcuts !

PR #4702, opened 2021-10-06
Update backstage.json

ntgussoni:
We decided to index the pages individually to have more control over the page rank instead of using the sitemap.

# Pull request motivation(s)

### DIFF
--- a/configs/backstage.json
+++ b/configs/backstage.json
@@ -2,14 +2,22 @@
   "index_name": "backstage",
   "start_urls": [
     {
+      "url": "https://backstage.io/docs/reference/",
+      "page_rank": 4
+    },
+     {
       "url": "https://backstage.io/docs/",
-      "page_rank": 5
+      "page_rank": 10
+    },
+     {
+      "url": "https://backstage.io/blog/",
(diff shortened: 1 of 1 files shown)


configs/backstage.json
```diff
@@ -1,12 +1,23 @@
 {
   "index_name": "backstage",
   "start_urls": [
-    "https://backstage.io/docs/"
-  ],
-  "sitemap_urls": [
-    "https://backstage.io/sitemap.xml"
```
shortcuts, line 7: The sitemap should still be present if available, it helps the scraper to find URLs but follows the `start_urls` structure

ntgussoni: Reading the docs it wasn't very clear, https://docsearch.algolia.com/docs/config-file/#sitemap_urls-optional
> DocSearch will try to read URLs from your sitemap(s) instead of following every link of your starts_urls.
I'll add it back

shortcuts: Thanks for the feedback, a new documentation is on the way https://github.com/algolia/docsearch/pull/1048

You can submit an alternative wording if you want!

PR #4703, opened 2021-10-06
Update lvl0 selector for Galasa

ind1go:
# Pull request motivation(s)

We're in the process of adding DocSearch to galasa.dev. User testing shows that the default 'Documentation' category is being selected for everything, plus the React component then limits to 5 results from a single category. So this PR adds a better lvl0 selector.

Every results comes under the 'Documentation' section, with no distinction between them, and max 5 results.

The user gets an awareness of the hierarchy of a result in the search modal, with up to 5 results from each of several categories.

### DIFF
--- a/configs/galasa.json
+++ b/configs/galasa.json
@@ -9,7 +9,9 @@
   ],
   "selectors": {
     "lvl0": {
-      "selector": "",
+      "selector": "(//nav[@id='sidebar']//div[contains(concat(' ', normalize-space(@class), ' '), ' markerInNavPath ') and not(contains(concat(' ', normalize-space(@class), ' '), ' selected '))]/a/text() | //nav[@id='headerNav']//a[contains(concat(' ', normalize-space(@class), ' '), ' selected ')]/text())[last()]",
+      "type": "xpath",
+      "global": true,
       "default_value": "Documentation"
     },
     "lvl1": "[class*='docContent'] h1",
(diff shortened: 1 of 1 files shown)


configs/galasa.json
```diff
@@ -9,10 +9,12 @@
...
     "lvl0": {
-      "selector": "(//nav[@id='sidebar']//div[contains(concat(' ', normalize-space(@class), ' '), ' markerInNavPath ') and not(contains(concat(' ', normalize-space(@class), ' '), ' selected '))]/a/text() | //nav[@id='headerNav']//a[contains(concat(' ', normalize-space(@class), ' '), ' selected ')]/text())[last()]",
-      "type": "xpath",
+      "selectors": [
+        "#sidebar div.markerInNavPath:not(.selected)",
+        "#headerNav a.selected"
+      ],
       "global": true,
-      "default_value": "Documentation"
+      "default_value": "Docs"
```
shortcuts: Sorry if it was not clear syntax will only work after the new infra. For now you should only have one DOM selector (https://docsearch.algolia.com/docs/config-file/#setting-a-default-value):

```
"selectors": "#sidebar div.markerInNavPath:not(.selected)",
"default_value": "Docs"
```

ind1go: Got it! Fixed.

shortcuts: Sorry for the confusion :D

PR #4705, opened 2021-10-06
Update radar.json

andlum:
# Pull request motivation(s)
Updating the radar.json to better resemble the Docusaurus 2 config: https://github.com/facebook/docusaurus/issues/4555#issuecomment-936527991

- Anchor links (the `#` that accompany headers) are appearing in search results 
- TOC links are appearing in search results

Search results should omit the anchor links symbol and TOC links should not appear in search results

### DIFF
--- a/configs/radar.json
+++ b/configs/radar.json
@@ -16,6 +16,24 @@
     "lvl5": "h6",
     "text": "p, li"
   },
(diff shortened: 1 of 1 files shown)


shortcuts:
You can also copy the selectors to provide better results

Issue #4699, opened 2021-10-06
[mdanalysis] no new search results after adding new sitemaps to sitemapindex


orbeckst:
# Bug: no search results after adding new sitemaps

### If it is a DocSearch index issue, what is the related `index_name` ?

* `index_name`= mdanalysis
* index file [mdanalysis.json](https://github.com/algolia/docsearch-configs/blob/master/configs/mdanalysis.json)

## What is the **current** behaviour?

We added two new sitemaps to our siteindex (see _Any other feedback_ below for details) for "[distopia](https://www.mdanalysis.org/distopia)" and "[pytng](https://www.mdanalysis.org/pytng/)". Searching for new unique content does _not_ give any results:

*If the current behaviour is a bug, please provide all the steps to reproduce and screenshots with context.*

To perform search, go to https://www.mdanalysis.org/ and use the search box as show in the screen shot:

![algolia search example on mdanalysis.org](https://user-images.githubusercontent.com/237980/136123442-c9b11beb-de1c-4e62-835d-8cca51b16156.png)

The screen shot shows that the unique term _CalcBondsOrtho_ (for distopia) is not found, as explained in more detail below:

#### distopia failed example text
1. search for "distopia" 
2. no exact results (only a fuzzy match in a blog post, which is not a correct match)
    - should have found https://www.mdanalysis.org/distopia/ and https://www.mdanalysis.org/distopia/api/distopia.html

#### distopia failed example API docs
Note that this example probably fails because content is in a dl (definition list) with dt/dd tags:

1. search for "CalcBondsOrtho"
2. no results: 
    - should have found https://www.mdanalysis.org/distopia/api/distopia.html?highlight=calcbondsortho#_CPPv4I0E14CalcBondsOrthovPK1TPK1TPK1TNSt6size_tEP1T

#### pytng failed example text
1. search ' "TNG API" '
2. No results found in pytng (only a blog post)
   - should have found https://www.mdanalysis.org/pytng/

#### pytng failed example API docs
Note that this example probably fails to find the API doc because content is in a dl (definition list) with dt/dd tags, the text use should have been found

1. search "TNGFileIterator"
2. No results found
   - should have found text use https://www.mdanalysis.org/pytng/documentation_pages/Examples.html
   - should have found API doc  https://www.mdanalysis.org/pytng/documentation_pages/API.html?highlight=tngfileiterator#pytng.TNGFileIterator

## What is the **expected** behaviour?

[...]

PR #4712, opened 2021-10-08
fix(mdanalysis): update config

shortcuts:
**Summary**

Fix typo in selectors


orbeckst:
Thanks for fixing the trailing comma. I noticed it the other day but didn't know if that was an error or not (was too lazy to check json specs...).

It hasn't fixed #4699 but this might be more related to the issues that you commented on in #4699 — will look into it in more detail.

PR #4743, opened 2021-10-18
fix(figma): update config

shortcuts:
**Summary**

handle `widget-docs` pages

PR #4745, opened 2021-10-18
configs (figma): exclude blog/ too

ymichael:
# Pull request motivation(s)

This was suggested here by @shortcuts  https://github.com/algolia/docsearch-configs/pull/4743#discussion_r730862507 and I asked to remove it 😢 - testing it out, I think we actually want this.

We return the listing page as a result:

image

Don't index the listing page at all!

### DIFF
--- a/configs/figma.json
+++ b/configs/figma.json
@@ -20,7 +20,10 @@
       "tags": ["plugin-docs"]
     }
   ],
-  "stop_urls": ["/blog/page"],
+  "stop_urls": [
+    "/blog/",
+    "/blog/page"
+  ],
   "selectors": {
     "default": {
       "lvl0": {


ymichael:
Oh ok will check it out thanks!

shortcuts:
> Oh ok will check it out thanks!

No problem! On your side, you'd only have to update the credentials of your frontend implementation, but the config is already parsed and the index populated

ymichael:
Hey I'm so sorry, I'm having a lot of trouble with the UI - I can't figure out how to make the changes in this PR via the UI 

Also, I don't think I fully understand what changing the credentials means - is there documentation for this you can link me to?

ymichael:
OH I see https://crawler.algolia.com/ - let me try that

shortcuts:
Don't hesitate to provide any feedback on the migration process, the features or anything! It's still an early stage for us so we're trying to improve everything we can

PR #4801, opened 2021-11-02
Update openeo.org and openeo.cloud config

m-mohr:
# Pull request motivation(s)

Site crawling is incomplete, see https://github.com/openEOPlatform/documentation/issues/26

Question: Can I set js_render and js_wait per start URL? Most sites are pre-rendered, but only some need js_render (and rarely js_wait) to be set. It would probably speed up crawling if I could do that, otherwise the PR is good as it is.

Some search terms don't return results.

All valid search terms return results.

### DIFF
--- a/configs/openeo-cloud.json
+++ b/configs/openeo-cloud.json
@@ -17,10 +17,6 @@
       "page_rank": 0
     },
     {
-      "url": "https://docs.openeo.cloud",
-      "page_rank": 1
-    },
-    {
       "url": "https://docs.openeo.cloud/file-formats/",
       "page_rank": 1,
       "selectors_key": "file-formats"
@@ -39,6 +35,10 @@
       "url": "https://docs.openeo.cloud/api/",
       "selectors_key": "api",
       "page_rank": -1
+    },
+    {
+      "url": "https://docs.openeo.cloud",
+      "page_rank": 1
     }
   ],
   "stop_urls": [],
--- a/configs/openeo-cloud.json
+++ b/configs/openeo-cloud.json
@@ -1,30 +1,147 @@
 {
   "index_name": "openeo-cloud",
   "start_urls": [
-    "https://docs.openeo.cloud"
+    {
+      "url": "https://open-eo.github.io/openeo-js-client/latest/",
+      "selectors_key": "js-client",
+      "page_rank": 0
+    },
+    {
+      "url": "https://open-eo.github.io/openeo-r-client/",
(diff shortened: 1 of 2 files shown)


shortcuts:
> Can I set js_render and js_wait per start URL? Most sites are pre-rendered, but only some need js_render (and rarely js_wait) to be set. It would probably speed up crawling if I could do that, otherwise the PR is good as it is.

Sorry, it's not possible (yet), users will be able to do so in our new infra! Read more: https://docsearch.algolia.com/docs/migrating-from-legacy

configs/openeo-cloud.json
```diff
@@ -1,30 +1,147 @@
...
+    },
+    {
+      "url": "https://open-eo.github.io/openeo-python-client/",
+      "selectors_key": "python-client",
+      "page_rank": 0
+    },
+    {
+      "url": "https://docs.openeo.cloud",
+      "page_rank": 1
+    },
```
shortcuts, line 22: Broader paths should be at the end of the `start_urls` array otherwise they will get picked up before deeper ones

configs/openeo-cloud.json
```diff
@@ -1,30 +1,147 @@
...
+    {
+      "url": "https://open-eo.github.io/openeo-r-client/",
+      "selectors_key": "r-client",
+      "page_rank": 0
+    },
+    {
+      "url": "https://open-eo.github.io/openeo-python-client/",
+      "selectors_key": "python-client",
+      "page_rank": 0
```
shortcuts, line 17: Default is 0 in case you want to clean those lines, but it's not necessary

configs/openeo.json
```diff
@@ -1,30 +1,88 @@
 {
```
shortcuts, line 1: Same comments about `page_rank: 0` and `start_urls` order

shortcuts (changes requested):

configs/openeo-cloud.json
m-mohr: Thanks, I like it more if the order prio is directly clear so left them in. :-)

configs/openeo.json
m-mohr: Fixed

configs/openeo-cloud.json
m-mohr: That makes a lot of sense, fixed! Thanks.

shortcuts (approved):
Good :D

PR #4810, opened 2021-11-04
Update openeo

m-mohr:
So I've drafted #4801 to include more sources, but I'm facing some issues:
- The openeo-cloud index seems empty now, I can't find anything. My searches are always returning empty result sets.
- The openeo index still works (strange, same selectors, both VuePress) but it doesn't include any of the new pages added.

Any clue what's going wrong here? I changed the config a bit to improve the search experience overall (e.g. _ should not split words), but I'm not sure how I can fix openeo-cloud. I tried running the docker image and run from source, but that seems targeted towards Linux and I'm on Windows, so that failed for me (cat/jq and pipenv issues).

### DIFF
--- a/configs/openeo-cloud.json
+++ b/configs/openeo-cloud.json
@@ -147,6 +147,23 @@
   "custom_settings": {
     "attributesForFaceting": [
       "lang"
+    ],
+    "separatorsToIndex": "_",
+    "synonyms": [
+      [
+        "cube",
+        "cubes",
+        "datacube",
+        "datacubes"
+      ],
+      [
+        "js",
+        "javascript"
+      ],
+      [
+        "gtiff",
+        "geotiff"
+      ]
     ]
   },
   "conversation_id": [
--- a/configs/openeo-cloud.json
+++ b/configs/openeo-cloud.json
@@ -45,20 +45,20 @@
   "selectors": {
     "default": {
       "lvl0": {
-        "selector": "p.sidebar-heading.open",
+        "selector": ".theme-default-content h1",
         "global": true,
         "default_value": "openEO Platform"
       },
-      "lvl1": ".theme-default-content h1",
-      "lvl2": ".theme-default-content h2",
-      "lvl3": ".theme-default-content h3",
-      "lvl4": ".theme-default-content h4",
-      "lvl5": ".theme-default-content h5",
+      "lvl1": ".theme-default-content h2",
+      "lvl2": ".theme-default-content h3",
+      "lvl3": ".theme-default-content h4",
+      "lvl4": ".theme-default-content h5",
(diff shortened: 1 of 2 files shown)


shortcuts:
Hey, indexing seems to be really slow due to the `js_wait` and `use_anchor` option, it creates a lot of record. Is there a reason for having those two?

shortcuts:
The `use_anchor` option won't be support in [our new infra](https://docsearch.algolia.com/docs/migrating-from-legacy) since it's directly handled in the indexing. I'll update both config to make them work properly!

m-mohr:
Hey @shortcuts , thanks for taking the time to check the issues.

Maybe I misunderstood the use_anchor option, but I had hoped that enabling it would allow that results jump directly to the right place on very long pages, e.g. here https://openeo.org/documentation/1.0/processes.html#sgn
If that's not the case, then use_anchor can be removed for sure.

js_wait was not included before so that can't be an issue with the previous PR. I just added it now because pages were not included and some pages are a bit slower, e.g. https://docs.openeo.cloud/data-collections/ and https://docs.openeo.cloud/processes/ - I'm not sure whether they are fast enough to be properly handled by js_render only?

So is the issue right now that indexing has not finished yet for the last PR? That would explain the issues indeed. Is there a way to see the index somehow to (1) see the status of the crawling and (2) check whether optimizations to the selectors could help?

shortcuts:
I was not able to push to the PR, so I created a new one https://github.com/algolia/docsearch-configs/pull/4812

> Maybe I misunderstood the use_anchor option, but I had hoped that enabling it would allow that results jump directly to the right place on very long pages, e.g. here https://openeo.org/documentation/1.0/processes.html#sgn
If that's not the case, then use_anchor can be removed for sure.

This should be handled by default, `use_anchor` will consider `anchors` as pages, and create multiple records which is not usually optimal

> js_wait was not included before so that can't be an issue with the previous PR. I just added it now because pages were not included(? - if indexing did not finish then screw that) and some pages are a bit slower, e.g. https://docs.openeo.cloud/data-collections/ and https://docs.openeo.cloud/processes/ - I'm not sure whether they are fast enough to be properly handled by js_render only?

Missing pages can be solved by adding a `sitemap.xml` file to your website or adding multiple entry points to the `start_urls` to help our scraper retrieve URLs from multiple pages, but it's indeed useful for bigger pages so I left it at `js_wait: 1`

> Is there a way to see the index somehow to (1) see the status of the crawling and (2) check whether optimizations to the selectors could help?

Indexing did not finish indeed. It will be possible to track the status of the crawl with the new infra (see https://docsearch.algolia.com/docs/migrating-from-legacy), but as of now it's not possible.

m-mohr:
Thanks a lot!

PR #4812, opened 2021-11-04
fix(config): fix openeo indexing

shortcuts:
follow up of https://github.com/algolia/docsearch-configs/pull/4810

### DIFF
--- a/configs/openeo-cloud.json
+++ b/configs/openeo-cloud.json
@@ -164,12 +164,11 @@
       ]
     ]
   },
+  "scrape_start_urls": false,
+  "js_render": true,
+  "js_wait": 1,
   "conversation_id": [
     "1611749633"
   ],
-  "scrape_start_urls": false,
-  "nb_hits": 297,
-  "js_render": true,
-  "js_wait": 3,
-  "use_anchors": true
+  "nb_hits": 13676
 }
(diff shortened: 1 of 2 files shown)


m-mohr:
@shortcuts Sorry for disturbing you again, but what's the state of the recrawl for both pages? I don't get any results yet and want to confirm it's still crawling and not another issue that may need to be solved. Thanks!

shortcuts:
Both have been crawled daily since the merge, last one finished ~20h ago

m-mohr:
@shortcuts Thanks. Hmm, then the configs are faulty, but I have no clue why that is... Is there anything indexed at all? I can't get any results at all.

shortcuts:
Definitely indexed! 
Screenshot 2021-11-09 at 14 09 46

It seems there's no `lang` selector in the config but the filter is used in the frontend, you need to add it to the selectors: [see this config for reference](https://github.com/algolia/docsearch-configs/blob/master/configs/next_router_vuejs.json#L72-L77)

m-mohr:
@shortcuts Oh, I see. Thanks!
I guess I could also remove it from the frontend, but if not, does the lang need to be provided per selector "group" or once for the whole config?

shortcuts:
for each selector group yes

PR #4817, opened 2021-11-07
Create config for bootstrap v3

Hkr32:
copied and adapted from this config
https://github.com/algolia/docsearch-configs/blob/master/configs/bootstrap-v3.json

will the key be the same? or do you need to write an application for a new one?

### DIFF
--- a/configs/bootstrap-su-v3.json
+++ b/configs/bootstrap-su-v3.json
@@ -1,0 +1,29 @@
+{
+  "index_name": "bootstrap-su-v3",
+  "start_urls": [
+    "https://getbootstrap.su/docs/3.4/"
+  ],
+  "sitemap_urls": [
+    "https://getbootstrap.su/docs/3.4/sitemap.xml"
+  ],
+  "stop_urls": [
+    "docs/3.4/examples/"
+  ],
+  "selectors": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hi, please also re-apply so we can proceed

Hkr32:
My main app https://github.com/algolia/docsearch-configs/blob/master/configs/bootstrap-su.json
Submitted a separate application for the third version on https://docsearch.algolia.com/apply/

shortcuts:
> Is it possible to place all this in one file? My main app https://github.com/algolia/docsearch-configs/blob/master/configs/bootstrap-su.json

Hey! Sure, in that case, you can just update your main file to support the v3 version

Hkr32:
And how to do it correctly? I do not understand very well how to do this correctly (((

shortcuts:
You can follow what you did here: https://github.com/Hkr32/docsearch-configs/blob/e4718c0d6d4f275ff052cf21da7f91b14b6952ca/configs/bootstrap-su.json#L27-L36

And have a specific version for `3.x` with a different `selectors_key`, see doc: https://docsearch.algolia.com/docs/legacy/config-file/#selectors_key-tailor-your-selectors

PR #4821, opened 2021-11-09
fix(jovo): update config

shortcuts:
**Summary**

Support v4 endpoint

### DIFF
--- a/configs/jovo.json
+++ b/configs/jovo.json
@@ -4,37 +4,31 @@
     {
       "url": "https://www.jovo.tech/docs/reference/",
       "page_rank": -1,
-      "tags": [
-        "reference"
-      ],
+      "tags": ["reference"],
       "selectors_key": "reference"
     },
     {
       "url": "https://www.jovo.tech/tutorials/",
       "page_rank": 1,
-      "tags": [
-        "tutorials"
-      ],
+      "tags": ["tutorials"],
       "selectors_key": "tutorials"
     },
     {
       "url": "https://www.jovo.tech/docs/",
       "page_rank": 2,
-      "tags": [
-        "docs"
-      ]
+      "tags": ["docs"]
+    },
+    {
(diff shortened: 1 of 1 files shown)


jankoenig:
Thank you. Is it possible to use our v3 and our v4 website separately?

We're going to launch our v4 publicly next week and are going to run our v3 website (right now https://www.jovo.tech/docs/) at v3.jovo.tech for a while.

Since the two versions v3 and v4 are completely different, it would be great to have separate versions so that the search results aren't mixed up.

If that's not possible, we'd probably remove the search from our v3 website and just use the v4 index.

shortcuts:
Hey @jankoenig, sure, you can make use of the `tags` https://github.com/algolia/docsearch-configs/blob/a5186aeaae2d928990d1343c25a68dc040838339/configs/jovo.json#L7 to [filter your search in the frontend](https://docsearch.algolia.com/docs/legacy/config-file/#using-custom-tags)

jankoenig:
Thank you @shortcuts!

Our v4 documentation is spread across /docs and a /marketplace, could the latter also be added to the config? https://v4.jovo.tech/marketplace

shortcuts:
Let me know when it's ready on your side and I'll merge!

configs/jovo.json
```diff
@@ -2,29 +2,52 @@
   "index_name": "jovo",
   "start_urls": [
     {
-      "url": "https://www.jovo.tech/docs/reference/",
+      "url": "https://v4.jovo.tech/docs/reference/",
```
jankoenig, line 5: Our v4 page doesn't hav a reference. Actually, our v3 one doesn't have one either, could `reference` be removed completely?

shortcuts: Done!

jankoenig:
Great, thanks @shortcuts! This is good to go from my side 🎉

Issue #4842, opened 2021-11-15
[infracost] Missing results for query in table entries, e.g. "aws_dms_replication_instance"


alikhajeh1:
I noticed that the table entries in the following 3 pages aren't showing-up in the search results, any idea if there's something we can do to fix this? the markdown code for the pages are [here](https://github.com/infracost/docs/tree/master/docs/supported_resources).
https://www.infracost.io/docs/supported_resources/aws
https://www.infracost.io/docs/supported_resources/azure
https://www.infracost.io/docs/supported_resources/google

shortcuts:
Hey @alikhajeh1,

Only the first and last child of the table are indexed in your config: https://github.com/algolia/docsearch-configs/blob/master/configs/infracost.json

If you'd like to update it, feel free to open a pull request

alikhajeh1:
Thanks @shortcuts! I should have noticed that sooner 😓 I sent a [PR](https://github.com/algolia/docsearch-configs/pull/4845) just now

PR #4845, opened 2021-11-16
Include table in search index

alikhajeh1:
# Pull request motivation(s)
Fixes https://github.com/algolia/docsearch-configs/issues/4842
Also adds the [Infracost blog](https://www.infracost.io/blog/) in the index as we publish monthly product updates and it'll be useful for users to search them.

### DIFF
--- a/configs/infracost.json
+++ b/configs/infracost.json
@@ -1,8 +1,19 @@
 {
   "index_name": "infracost",
   "start_urls": [
-    "https://www.infracost.io/docs/"
-  ],
+    {
+      "url": "https://www.infracost.io/docs/",
+      "tags": [
+        "docs"
+      ]
+    },
+    {
+      "url": "https://www.infracost.io/blog/",
+      "tags": [
+        "blog"
+      ]
+    }
+  ],  
   "sitemap_urls": [
     "https://www.infracost.io/sitemap.xml"
   ],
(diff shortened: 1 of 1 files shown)


configs/infracost.json
```diff
@@ -19,8 +30,8 @@
     "lvl2": "article h2",
     "lvl3": "article h3",
     "lvl4": "article h4",
-    "lvl5": "article h5, article td:first-child",
-    "text": "article p, article li, article td:last-child"
+    "lvl5": "article h5, article",
+    "text": "article p, article li, article"
```
shortcuts, line 34:
```suggestion
    "lvl5": "article h5",
    "text": "article p, article li, article td"
```

I assume this is what you'd achieve

alikhajeh1: I think so, we have `td` that have `code` too, e.g.:
```
<tr><td>Bastion</td><td><code>azurerm_bastion_host</code></td><td></td></tr>
```

Should I include that too? `article p, article li, article td, article td code`

shortcuts: It should be enough like that, but feel free to re-open a PR if not

shortcuts (approved):

configs/infracost.json
alikhajeh1: Thanks!

PR #4556, opened 2021-09-06
Update oasisengine.json [new]

eyworldwide:
Fix the old PR can't merge master.

Old PR: https://github.com/algolia/docsearch-configs/pull/4551

@shortcuts Try to merge this PR.

PR #4850, opened 2021-11-18
Oasis engine site: upgrade version to 0.6

eyworldwide:
# Pull request motivation(s)

Upgrade version to 0.6.  Just add a new version.


shortcuts (approved):

eyworldwide:
@shortcuts I need your help,  because it did not work after you merge the request. 

I've added a new version **0.6** to the `versions` field, but I can't search anything in the **0.6** version of the site(https://oasisengine.cn/0.6/docs/install):

image

However, I still can search everything in the **0.5** version of the site (https://oasisengine.cn/0.5/docs/install) :
image

And I‘ve checked the `sitemap.xml`(https://oasisengine.cn/sitemap.xml), which have been updated to `0.6`:
image

shortcuts:
Hey @eyworldwide, 

Most of the previous versions (0.3, 0.4, 0.5) now redirects to 404, is that expected?

Also, it seems your selectors does not output records for the 0.6 version, they might need to be fixed!

You can handle multiple selectors group using the `selectors_key` option: https://docsearch.algolia.com/docs/legacy/config-file/#selectors_key-tailor-your-selectors

eyworldwide:
@shortcuts Thank for your questions and advices.

> Most of the previous versions (0.3, 0.4, 0.5) now redirects to 404, is that expected?

Before 0.5 version, we have no English version of docs, so these versions(0.3, 0.4) redirects to 404.

> Also, it seems your selectors does not output records for the 0.6 version, they might need to be fixed!
 
I've checked the outputs in html source, `0.6` version has `article h1` tags just as `"lvl1": "article h1"` set in the config file.

image

- config file:
image

I feel confused because I did the same thing in the PR of `0.5` upgrade(https://github.com/algolia/docsearch-configs/pull/4556/files), and it works. But I fail now when upgrade to `0.6` version.

> You can handle multiple selectors group using the selectors_key option: https://docsearch.algolia.com/docs/legacy/config-file/#selectors_key-tailor-your-selectors

Thank you for your advice of using `selectors_key`, which is useful for multiple selectors group, and I will try it in the future.

PR #4852, opened 2021-11-18
Update Jovo config (Jovo v4 launch 🚀 )

jankoenig:
Hi! We're launching Jovo `v4` today, which is quite a big launch for us 🎉 

For this, we had some changes to our websites:
- `jovo.tech` is now at `v3.jovo.tech`
- `v4.jovo.tech` is now at `jovo.tech`

Would be amazing to have this merged soon so that our users can experience the great search on our `v4` website :rocket: 

Thanks a lot!!


shortcuts (approved):
Congrats!

PR #4857, opened 2021-11-19
Update objectiv.json

ivarpruijn:
Update `text` selector for our search config. We will deploy an update to our docs to include the `ignore-for-search` class to filter on.

URL: https://objectiv.io/docs/ (search not enabled yet)

# Pull request motivation(s)
The footer in our documentation currently gets indexed, resulting in search results that list the items in the footer multiple times.

Search for e.g. 'dataframe' and see 'Edit This Page' multiple times:
image

The search results do not show the 'Edit This Page' item multiple times. The links in the footer of our docs are filtered out of the search results / not indexed.

### DIFF
--- a/configs/objectiv.json
+++ b/configs/objectiv.json
@@ -20,8 +20,9 @@
     "lvl3": "article h3",
     "lvl4": "article h4",
     "lvl5": "article h5, article td:first-child",
-    "text": "article p, article li:not(.ignore-for-search), article td:last-child"
+    "text": "article p, article li, article td:last-child"
   },
+  "selectors_exclude": ["footer"],
   "strip_chars": " .,;:#",
   "custom_settings": {
     "separatorsToIndex": "_",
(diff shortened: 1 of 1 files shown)


configs/objectiv.json
```diff
@@ -20,7 +20,7 @@
     "lvl3": "article h3",
     "lvl4": "article h4",
     "lvl5": "article h5, article td:first-child",
-    "text": "article p, article li, article td:last-child"
+    "text": "article p, article li:not(.ignore-for-search), article td:last-child"
```
shortcuts, line 23: We have the `selectors_exclude` option so you can better read what you index/what you exclude: https://docsearch.algolia.com/docs/legacy/config-file/#selectors_exclude-optional

shortcuts (commented):
Good! Let me know when I should merge the PR as it will instantly trigger a crawl

configs/objectiv.json
ivarpruijn: That looks like a great solution for us, thank you! Will update this PR and let you know.

ivarpruijn: PR updated. Would be great if you could merge and trigger a crawl. Thanks!

shortcuts (approved):

PR #4865, opened 2021-11-22
feat: [arco] add vue site and language config

yuhan0709:
# Pull request motivation(s)

I want to add two configurations:
1. Distinguish the usage documentation of `react` and `vue`.
2. Distinguish between `zh-CN` and `en-US` document languages

I changed `start_urls`, but these two URLs have overlapping scenes, I'm not sure if this is possible.

If not, is there any good solution?

Looking forward to your reply， THX.

### DIFF
--- a/configs/arco.json
+++ b/configs/arco.json
@@ -1,25 +1,43 @@
 {
   "index_name": "arco",
   "start_urls": [
-    "https://arco.design/",
-    "https://arco.design/react/docs/start"
+    {
+      "url": "https://arco.design/vue/*",
+      "tags": ["vue"],
+      "selectors_key": "vue"
+    },
+    {
+      "url": "https://arco.design/react/*",
+      "tags": ["react"]
+    },
+    {
+      "url": "https://arco.design.+?(en-US)",
+      "tags": ["en-US"]
+    }
   ],
-  "sitemap_urls": [
-    "https://arco.design/sitemap.xml"
-  ],
+  "sitemap_urls": ["https://arco.design/sitemap.xml"],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".ac-content h1",
-    "lvl1": ".ac-content h2",
-    "lvl2": ".ac-content h3",
-    "lvl3": ".ac-content h4",
-    "lvl4": ".ac-content h5",
-    "lvl5": ".ac-content h6",
-    "text": ".ac-content p, .ac-content li"
+    "default": {
+      "lvl0": ".ac-content h1",
+      "lvl1": ".ac-content h2",
+      "lvl2": ".ac-content h3",
+      "lvl3": ".ac-content h4",
+      "lvl4": ".ac-content h5",
+      "lvl5": ".ac-content h6",
+      "text": ".ac-content p, .ac-content li"
+    },
+    "vue": {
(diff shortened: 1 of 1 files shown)


configs/arco.json
```diff
@@ -1,25 +1,43 @@
...
+      "selectors_key": "vue"
+    },
+    {
+      "url": "https://arco.design/react/*",
+      "tags": ["react"]
+    },
+    {
+      "url": "https://arco.design.+?(en-US)",
+      "tags": ["en-US"]
+    }
```
shortcuts, line 16: Matching patterns also work for `start_urls` so you don't need wildcards or variables when there's only one option

yuhan0709: I have changed it, but in the case of en-US, we still need to use regular to match

yuhan0709:
@shortcuts hello, is this pr still being checked?

shortcuts (approved):

yuhan0709:
I want to know how long it will take effect after the modification? I currently see that I still can’t find what I want.

yuhan0709:
I want to manually trigger the crawler update but why can't log in to the crawler :( 
image

yuhan0709:
I login in the crawlers, But it’s empty, I can’t manually trigger the crawler :(
image

shortcuts:
> I want to know how long it will take effect after the modification? I currently see that I still can’t find what I want.

Hey, crawls are triggered after a merge, so you should already see the changes

> I login in the crawlers, But it’s empty, I can’t manually trigger the crawler :(

Your website has not been migrated to the Crawler yet, see https://docsearch.algolia.com/docs/migrating-from-legacy

yuhan0709:
> Hey, crawls are triggered after a merge, so you should already see the changes
> 
I have tagged different startUrls, but I cannot find the content even after adding the searchParameters, Is there a way I can see the captured data.

> Your website has not been migrated to the Crawler yet, see https://docsearch.algolia.com/docs/migrating-from-legacy

For this migration, maybe I should receive a migration email?  Looking forward :)

shortcuts:
> I have tagged different startUrls, but I cannot find the content even after adding the searchParameters, Is there a way I can see the captured data.

Sorry we don't have a way to provide access to DocSearch users yet, if URLs/records are missing the best bet would be to add a `sitemap.xml` to your website and/or include client-side rendering options if necessary: https://docsearch.algolia.com/docs/legacy/config-file#js_render-optional

> For this migration, maybe I should receive a migration email? Looking forward :)

Yep exactly, you'll receive one once the migration is done for your index!

yuhan0709:
I’m here again~ I really don’t know why I added `facetFilters: ['tags:en-US'],` but I can’t find anything...😭
This is our website https://arco.design/react/en-US/docs/start

shortcuts:
`react` will be picked before `en-US` https://github.com/algolia/docsearch-configs/pull/4865/files#diff-19df1e2cd29582f371fddacc3cd6eab952f1ffb228324bca255da1c7ba6c09b6R10

`start_urls` are picked hierarchically

yuhan0709:
So if my path is `/react/en-US/docs/start`, it actually hits the `tag` of `react`? And why is there also nothing when I search in `tags:vue`. 👉https://arco.design/vue/en-US/docs/start

PR #4866, opened 2021-11-22
Update objectiv.json

ivarpruijn:
The pages of the Modeling section in our docs (https://objectiv.io/docs/modeling/) load content dynamically. 

This PR adds the `js_render` and `js_wait` options to our config to enable crawling the content on those pages. We set `js_wait` to 2 seconds to be on the safe side, though the pages generally load in a few hundred milliseconds; we hope the impact on crawling speed is not too high.

### DIFF
--- a/configs/objectiv.json
+++ b/configs/objectiv.json
@@ -23,6 +23,8 @@
     "text": "article p, article li, article td:last-child"
   },
   "selectors_exclude": ["footer"],
+  "js_render": true,
+  "js_wait": 2,
   "strip_chars": " .,;:#",
   "custom_settings": {
     "separatorsToIndex": "_",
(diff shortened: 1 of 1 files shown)


configs/objectiv.json
```diff
@@ -23,6 +23,8 @@
     "text": "article p, article li, article td:last-child"
   },
   "selectors_exclude": ["footer"],
+  "js_render": true,
+  "js_wait": 2,
```
shortcuts, line 27: I'd suggest to start with `1` which should be enough and increase if necessary, since this `wait` time is applied on **every** pages

ivarpruijn: Sounds good, will update. Thanks.

ivarpruijn: Updated

shortcuts (approved):
Hope this solves the issue!

ivarpruijn:
Thanks again @shortcuts !

shortcuts:
No problem :D

PR #4869, opened 2021-11-24
Oasis engine site: Remove version  variables

eyworldwide:
# Pull request motivation(s)
As happened in https://github.com/algolia/docsearch-configs/pull/4850, it's troublesome to pull a request every time  when upgrade to a version of the site. 

So I remove the version variables in `start_urls`, and the crawler still can fetch the urls from `sitemap.xml`.

Any advice welcome: )


shortcuts:
Hey, I think what's best is adding the `docsearch` meta tags in your DOM so you only have to update it on your side and the crawler will retrieve those values automatically: https://docsearch.algolia.com/docs/required-configuration/#introduce-global-information-as-meta-tags

eyworldwide:
@shortcuts Thank you for advice. I've updated the codes and I've add a version meta in DOM
(see: view-source:https://oasisengine.cn/0.6/docs/install/ ):
image

If there's no problem, pls merge the request. Thank you again.

shortcuts (approved):

eyworldwide:
@shortcuts It didn't work after merging the request. I think maybe there's some cache in server of docsearch. because I can search everything in 0.5 version (https://oasisengine.cn/0.5/docs/install ), but still can't search anything in 0.6 (https://oasisengine.cn/0.6/docs/install ).

shortcuts:
> @shortcuts It didn't work after merging the request. I think maybe there's some cache in server of docsearch. because I can search everything in 0.5 version (https://oasisengine.cn/0.5/docs/install ), but still can't search anything in 0.6 (https://oasisengine.cn/0.6/docs/install ). 

Hey, I'll try locally and let you know

How do you send filters on your frontend?

configs/oasisengine.json
```diff
@@ -2,28 +2,7 @@
...
-      "variables": {
-        "version": [
-          "0.3",
-          "0.4",
-          "0.5",
-          "0.6"
-        ]
-      },
-      "page_rank": 1
+      "url": "https://oasisengine.cn/(?P<version>.*?)/"
```
shortcuts, line 5: I did not saw the `version` was still present there... that's the issue

shortcuts:
this should work: https://github.com/algolia/docsearch-configs/pull/4877

configs/oasisengine.json
eyworldwide: It works, thank you! @shortcuts I've seen the PR (https://github.com/algolia/docsearch-configs/pull/4877).

DocSearch is a wonderful product but it's difficult to debug.

shortcuts: Don't worry, all of these process are fixed with our new infra: https://docsearch.algolia.com/docs/migrating-from-legacy

Keep an eye out on your emails :D

PR #4872, opened 2021-11-25
fix: use the lang attribute of <html/>

yuhan0709:
# Pull request motivation(s)

The `en-US` and other tags may overlap

use the lang attribute of `<html/>`

### DIFF
--- a/configs/arco.json
+++ b/configs/arco.json
@@ -9,10 +9,6 @@
     {
       "url": "https://arco.design/react",
       "tags": ["react"]
-    },
-    {
-      "url": "https://arco.design.+?(en-US)",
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
Looks good!

yuhan0709:
It's very collapsed, but it still doesn't work. The result is empty :(

yuhan0709:
I want to know if the facet I added takes effect, does the crawler really get the data corresponding to the tag?

shortcuts:
Your website requires client-side rendering, fixed it here: https://github.com/algolia/docsearch-configs/pull/4878

PR #4886, opened 2021-11-30
Modify Workbench config

brad2k:
Hello! 

The crawler is not being populated with helpful data for the page that I would primarily like to surface during a search, so I am modifying a few values to hopefully get better data.

I have added a global classname to my primary body content: `DocSearch-content`

I also added a short `js_wait` as an experiment to see if perhaps header content is not yet available when the crawler begins.

As you can see in the screenshot below, I am trying to search for our Accordion component, which lives [here](https://workbench.gusto.com/components/accordion/). The first category of results show only as "#" and lists various anchors on the page, but none of the header text is being displayed, which is confusing. When searching for a component, I always want those results to appear first.

localhost_8000_components_skeleton_

Please let me know if I'm doing something wrong. Thank you for your help!

### DIFF
--- a/configs/workbench.json
+++ b/configs/workbench.json
@@ -4,21 +4,19 @@
     "https://workbench.gusto.com/",
     "https://workbench.gusto.com/getting-started/engineers/"
   ],
-  "stop_urls": [
-    "\\/\\?"
-  ],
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
Looks good :D

PR #4889, opened 2021-11-30
chore: Update `gatsbyjs.json` `sitemap_urls`

fk:
We (more or less) recently changed the location of our sitemap on gatsbyjs.com.
This PR adjusts the gatsbyjs.com docsearch config accordingly.

~cc @LekoArts @DSchau please double check the new URL.~ 🙏

### DIFF
--- a/configs/gatsbyjs.json
+++ b/configs/gatsbyjs.json
@@ -34,7 +34,7 @@
     }
   ],
   "sitemap_urls": [
-    "https://www.gatsbyjs.com/sitemap.xml"
+    "https://www.gatsbyjs.com/sitemap/sitemap-0.xml"
   ],
   "stop_urls": [
     "/packages/",


shortcuts:
Hey @fk, the `gatsbyjs` index have been migrated to [our new infra](https://docsearch.algolia.com/docs/migrating-from-legacy), you should be able to make the changes directly from the web interface and trigger a crawl :D

Let me know if you need help

DSchau:
Hey @shortcuts  -- could you let us know the account holder of the instance of Algolia? I just tried my:

- github (dschau)
- Gatsby Email (dustin at gatsbyjs dot com)

and no dice! The latter prompted me to register for an account. Could you add my account (either the Github or the Gatsby Email) as an admin on the instance?

shortcuts:
Hey @DSchau, 

> could you let us know the account holder of the instance of Algolia?

There's 6 owners of the gatsbyjs index. I've sent an invite to dustin@gatsbyjs.com (you'll need to create an Algolia account) so you can manage the owners on your side!

fk:
Thank you, @shortcuts! 🤗 🙏

shortcuts:
No problem, please don't hesitate to ping me if there's any issue :D

DSchau:
@shortcuts I think we have two Algolia instances, one of which is actually handling the search volume from gatsbyjs.com.

Specifically, that application is called "Docsearch" and the index is gatsbyjs, and its ID is `BH4D9OD16A`. What makes me believe that we're using that one and not the one we're invited to is that the one we're invited to doesn't have any searches, whereas that one (the one that ends in 16A) has thousands of searches!

Could I get an invite to that one? As a general practice, would you prefer that we migrate to the one that I was invited to above?

cc @fk

shortcuts:
> I think we have two Algolia instances, one of which is actually handling the search volume from gatsbyjs.com.

As of now, not all the users have migrated to our new infra, so you can still see our legacy app (`BH4D9OD16A`) in your dashboard, but you should refer to the new one (` HQ84ML84ER ` in your case). You can read [this small paragraph](https://docsearch.algolia.com/docs/migrating-from-legacy#why-do-i-see-two-algolia-apps-in-my-dashboard) with a bit more details in it

> Could I get an invite to that one?

The invite I sent you was to the new infra/app (` HQ84ML84ER `), would you also like an invite to the old one? (You'll only see the `analytics` tab of the dashboard).

>  As a general practice, would you prefer that we migrate to the one that I was invited to above?

As it provides a web interface to trigger/schedule crawls, edit your config, GitHub action and much more, I'd definitely recommend you to migrate ASAP :D

DSchau:
Yes, please -- invite would be great!

fk:
> You can read this small paragraph with a bit more details in it

Argh :facepalm:, again. Apologies for not being a thorough reader, @shortcut – thanks for holding our hands!

shortcuts:
> Yes, please -- invite would be great!

Done!

> Argh 🤦, again. Apologies for not being a thorough reader, @shortcut – thanks for holding our hands!

No problem :D

Issue #4893, opened 2021-12-01
[avax] Missing results for query "avax"


yulin-dong:
No search result for query of anything such as `avax` on https://docs-beta.avax.network/. Your help is greatly appreciated.

shortcuts:
Hey,

To use the `contextualSearch` option of Docusaurus, your DocSearch config needs to follow the Docusaurus v2 config template, you can read more here: https://docusaurus.io/docs/2.0.0-beta.8/search#contextual-search

PR #4894, opened 2021-12-01
fix(avax): update config

shortcuts:
Use Docusaurus v2 config template

Fixes https://github.com/algolia/docsearch-configs/issues/4893

### DIFF
--- a/configs/avax.json
+++ b/configs/avax.json
@@ -2,16 +2,12 @@
   "index_name": "avax",
   "start_urls": [
     {
-      "url": "https://docs.avax.network",
-      "tags": [
-        "prod"
-      ]
+      "url": "https://docs.avax.network/",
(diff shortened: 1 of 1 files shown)


configs/avax.json
```diff
@@ -2,26 +2,48 @@
   "index_name": "avax",
   "start_urls": [
     {
-      "url": "https://docs.avax.network",
+      "url": "https://docs.avax.network/",
       "tags": ["prod"]
     },
     {
       "url": "https://docs-beta.avax.network",
```
yulin-dong, line 9: should `/` be added at the end of this line? or does it matter?

shortcuts: Oops I forgot this one indeed, it will work anyway but you can add it if you want!

PR #4896, opened 2021-12-02
px4.json: Add a sitemap for the master branch

hamishwillee:
The master branch now has its own, independent sitemap.

### DIFF
--- a/configs/px4.json
+++ b/configs/px4.json
@@ -1,6 +1,6 @@
 {
   "index_name": "px4",
-  "sitemap_urls": ["https://docs.px4.io/sitemap.xml"],
+  "sitemap_urls": ["https://docs.px4.io/sitemap.xml","https://docs.px4.io/master/sitemap.xml"],
   "start_urls": [
     {
       "url": "https://docs.px4.io/master/",


shortcuts:
Hey @hamishwillee, your index has been migrated to [our new infra](https://docsearch.algolia.com/docs/migrating-from-legacy), you can submit your change directly on our web interface live editor and trigger a crawl on your side!

hamishwillee:
@shortcuts Thanks. Unfortunately (?) you set this up for me so I have no direct login access to the web interface. I created a personal login with my hamishwillee at gmail dot com that you could perhaps link to this to allow me to do so? Or perhaps given this is an open source organisation site, it should be created in a new PX4 organisation login?

Please advise.

shortcuts:
Hey, I can see that your email should have receive an invite and the migration email 🤔 

I've sent a new invite, but you can also look for emails from `docsearch@algolia.com` and `accounts.docsearch@algolia.com` :D

hamishwillee:
Thanks very much. Closing!

BTW, some things aren't getting crawled due to extractor too many records and memory limits. These are big pages of important links. Is there anything I can do to fix that? 
A specific page is https://docs.px4.io/master/en/advanced_config/parameter_reference.html

shortcuts:
Hey, sorry for the late reply!

I can see the `autotune.html` URL getting crawled successfully now, does it show up in your search? 

One of the reason you can see few results, is that all the records are grouped under the `documentation` of your `lvl0`, I'd suggest you to make sure the `selector` for this level works so that you have blocks of heading (which also leads in more results (e.g. https://docsearch.algolia.com/ where we select the active sidebar item))

> BTW, some things aren't getting crawled due to extractor too many records and memory limits.

We've bumped the limit to 750 records per pages, but will work on a better fix in a near future to try to reduce the number of records outputted, which should help.

However, big (**big**) pages might not be handled and are also not recommended

hamishwillee:
Thanks @shortcuts .
Its not showing up, nor is http://docs.px4.io/master/en/concept/events_interface.html if I search for events.

Thing is, these are the "new" docs, and everything else seems to be found when I enter the title heading - e.g. PX4 Architecture, Getting Started etc. So I'm suspicious that something more fundamental is going on.

shortcuts:
> Was I supposed, perhaps, to update an API key?

You should indeed update the frontend credentials with your new `appId` and `apiKey` (they can be both found in the migration email). You can also visit the Algolia dashboard and try searching for the terms here, since there's no pagination/limitation you'll make sure to view all records matching your query and see if it's indexed.

Looking at the monitoring of the Crawler, I can definitely see records being indexed for those pages, but due to this:

> One of the reason you can see few results, is that all the records are grouped under the documentation of your lvl0, I'd suggest you to make sure the selector for this level works so that you have blocks of heading (which also leads in more results (e.g. https://docsearch.algolia.com/ where we select the active sidebar item))

You might not see it in the frontend

hamishwillee:
Thanks @shortcuts - sorry for not getting back to you - took me a bit to get back to this.

So the problem appears to have fixed itself - both autotune and events are now showing up in search. Don't know "why" though. I was using the wrong API key and I have now updated that too.

PR #4905, opened 2021-12-03
feat: start_urls add en-us

yuhan0709:
# Pull request motivation(s)

`facetFilters` add `lang:en` not working

`facetFilters` add `lang:en` will search in the en-US site

### DIFF
--- a/configs/arco.json
+++ b/configs/arco.json
@@ -11,7 +11,9 @@
       "tags": ["react"]
     },
     "https://arco.design/",
-    "https://arco.design/vue/docs/start"
+    "https://arco.design/vue/docs/start",
+    "https://arco.design/react/en-US/docs/start",
+    "https://arco.design/vue/en-US/docs/start"
   ],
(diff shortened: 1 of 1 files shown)


configs/arco.json
```diff
@@ -11,7 +11,9 @@
       "tags": ["react"]
     },
     "https://arco.design/",
-    "https://arco.design/vue/docs/start"
+    "https://arco.design/vue/docs/start",
+    "https://arco.design/react/en-US/docs/start",
+    "https://arco.design/vue/en-US/docs/start"
```
shortcuts, line 16: Hey, to enable filtering in your search, you can follow those small tutorials: 
- [using regular expressions](Closing since applications have been answered/deployed)
- [using custom tags](https://docsearch.algolia.com/docs/legacy/config-file/#using-custom-tags)

shortcuts (changes requested):

PR #4915, opened 2021-12-07
Update documentation paths for the "loadsterperformance" index.

azhawkes:
We recently reorganized the documentation content on our site, and
the Loadster manual is now located under /manual instead of
/documentation/manual. We've also added a changelog.

https://github.com/loadster


azhawkes:
Also, thanks very much for this awesome service!

azhawkes:
Never mind, we've just switched over to the Algolia Crawler and the new configuration seems to be working there. Thanks anyway!

PR #4916, opened 2021-12-07
[TypeScript] Remove old handbook pages

orta:
# Pull request motivation(s)

Removes the old handbook pages which are marked as deprecated from the search index

We don't want to break URLs, so the pages still exist but there's a big "you want to go here instead" marker for links, it's probably about time we remove them from the search index.

N/A

N/A

### DIFF
--- a/configs/typescriptlang.json
+++ b/configs/typescriptlang.json
@@ -24,7 +24,15 @@
   "stop_urls": [
     "/global-plugin/",
     "declaration-files/templates/",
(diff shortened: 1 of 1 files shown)


shortcuts (approved):
Looks good!

(Side note: you should receive an email soon with details about our new infra and your new Algolia app, you can read more [here](https://docsearch.algolia.com/docs/migrating-from-legacy) :D)

PR #4951, opened 2021-12-15
Update genome.json

cuongld2:
Change the base url

# Pull request motivation(s)


cuongld2:
Please help to merge this.

PR #4960, opened 2021-12-16
Update sap_spartacus.js

gladius-mtl:
Remove `"https://sap.github.io/",` from the list of `startUrls` since this URL should never be indexed for the `spartacus-docs` site. Unless I'm mistaken and `"https://sap.github.io/",` should really be in the list of `startUrls` ? 🙂


shortcuts:
Hey, this repo only hosts the config for a future reference (and mostly inform users), the config should be edited in the interface, see: https://docsearch.algolia.com/docs/manage-your-crawls#update-your-config

PR #4969, opened 2021-12-22
Update Workbench config

brad2k:
The crawler was not able to find the content on the previous L0 selector for some reason. Trying again with a different one. 

Also removing sitemap in favor of explicitly-ranked start_urls since not all pages in the sitemap were being indexed for some reason.


configs/workbench.json
```diff
@@ -1,11 +1,38 @@
...
+    },
+    {
+      "url": "https://workbench.gusto.com/resources/"
+    },
+    {
+      "url": "https://workbench.gusto.com/updates/"
+    }
+  ],
   "stop_urls": ["\\/\\?"],
-  "sitemap_urls": ["https://workbench.gusto.com/sitemap/sitemap-0.xml"],
```
shortcuts, line 5: Specifying the sitemap is always a good point, we do not chose one or the other but it complements URLs found from start_urls, I'd recommend to keep it

configs/workbench.json
```diff
@@ -1,11 +1,38 @@
...
   "selectors": {
     "lvl0": {
-      "selector": ".DocSearch-active-menu-item",
+      "selector": ".DocSearch-l0",
```
shortcuts, line 35: Selector seems valid, is there any client-side rendered content?

shortcuts (changes requested):

configs/workbench.json
brad2k: Thank you for pointing that out. I added it back. 👍

configs/workbench.json
brad2k: The previous selector (`.DocSearch-active-menu-item`) was part of our sidenav, which is partially rendered in a React `useEffect` so maybe there was a delay and I needed to add a `js_wait`. Unfortunately I cannot test it, so I would be making a random guess. 

The new selector (`.DocSearch-l0`) is a hidden `span` on each page with a string specifying its L0 value. It is rendered the same as all the other text content on the page, so I hope that it will extract properly.

shortcuts: Perfect!

The new infra will allow testing crawls etc. so keep an eye on your emails!

shortcuts (approved):

PR #4970, opened 2021-12-22
Update based on docusaurus-2.json selectors

philipp985:
https://github.com/algolia/docsearch-configs/blob/master/configs/docusaurus-2.json

# Pull request motivation(s)


shortcuts:
Hey, your index have been migrated to our new infra (https://docsearch.algolia.com/docs/migrating-from-legacy/), the changes should be done via the web interface!

PR #4971, opened 2021-12-22
nus-cs2103-ay1920s1.json: Update site location

damithc:
Reason: This index is for an course website. The config file needs to update because the website has moved to a new location, for the new semester.


damithc:
Gentle reminder about this... I tried using the migrated config instead but the online editor does not allow me to change the site location. 
image

PR #5007, opened 2022-01-05
Update pan.json

sserrata:
Switching to use canonical URL

Noticed that we are no longer indexing content from panos.pan.dev after move to Firebase and migration to new DocSearch.

# Pull request motivation(s)

No search results for facet filter `tags:strata`.

Should be seeing search results.

### DIFF
--- a/configs/pan.json
+++ b/configs/pan.json
@@ -20,7 +20,7 @@
       ]
     },
     {
-      "url": "https://strata.pan.dev/",
+      "url": "https://panos.pan.dev/",
       "tags": [
         "strata"
       ]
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, your website has been migrated to the new DocSearch infrastructure, you can edit your config via our live editor without waiting for an approval!

Read more here: https://docsearch.algolia.com/docs/migrating-from-legacy

sserrata:
Thanks @shortcuts! I've also had The following Algolia case opened concerning the same issue:

https://support.algolia.com/hc/en-us/requests/473938

I appreciate any help in the matter.

PR #5022, opened 2022-01-12
discord.json: index h6

typpo:
There are useful `h6` tags that we should index, e.g. on https://discord.com/developers/docs/reference

# Pull request motivation(s)


shortcuts:
Hey, the discord index have been migrated to our new infra (read more here: https://docsearch.algolia.com/docs/migrating-from-legacy). You can now do the edits directly from our web interface and trigger and also trigger new crawls.

You should have receive an email from `docsearch@algolia.com` about it

PR #5029, opened 2022-01-13
activepivot documentation: Docusaurus1 and Docusaurus2 sub-sites

gguimezanes:
add new (default) selectors for docusaurus2, keeping old (docusaurus1) selectors for previous versions

# Pull request motivation(s)

The search does not work for the new versions of our documentation, since we migrated to Docusaurus 2. For example, https://activeviam.com/activepivot/5.11.0/

The search should work on both the older versions of the documentation (Docusaurus 1) and the newer (Docusaurus 2).
For example, search for "Copper" on both https://activeviam.com/activepivot/5.10.0/ and
https://activeviam.com/activepivot/5.11.0/

### DIFF
--- a/configs/activepivot.json
+++ b/configs/activepivot.json
@@ -1,7 +1,63 @@
 {
-  "index_name": "activepivot",
+"index_name": "activepivot",
   "start_urls": [
-    "https://activeviam.com/activepivot/"
+    "https://activeviam.com/activepivot/",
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, this index has been migrated to our new infra, you can read more about it here: https://docsearch.algolia.com/docs/migrating-from-legacy

Changes can be done via our web interface, which also allow you to trigger new crawls!

PR #5031, opened 2022-01-14
tweak lvl0 and lvl1 selectors for transferwise

dozortsev:
# Pull request motivation(s)

Want to fix empty results

Screen Shot 2022-01-14 at 1 49 09 PM

Should similar to https://formatjs.io/

Screen Shot 2022-01-14 at 1 50 26 PM

### DIFF
--- a/configs/transferwise.json
+++ b/configs/transferwise.json
@@ -5,12 +5,16 @@
   ],
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".Content h1",
-    "lvl1": ".Content h2",
-    "lvl2": ".Content h3",
-    "lvl3": ".Content h4",
-    "lvl4": ".Content h5",
-    "lvl5": ".Content h6",
+    "lvl0": {
+      "selector": "a.Nav__Link",
+      "global": true
+    },
+    "lvl1": ".Content h1",
+    "lvl2": ".Content h2",
+    "lvl3": ".Content h3",
+    "lvl4": ".Content h4",
+    "lvl5": ".Content h5",
+    "lvl6": ".Content h6",
     "text": ".Content p, .Content li"
   },
   "conversation_id": [
--- a/configs/transferwise.json
+++ b/configs/transferwise.json
@@ -6,7 +6,7 @@
   "stop_urls": [],
   "selectors": {
     "lvl0": {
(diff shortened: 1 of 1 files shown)


dozortsev:
Hey @shortcuts may I get a review for small changes in our config please?

configs/transferwise.json
```diff
@@ -5,12 +5,16 @@
...
   "stop_urls": [],
   "selectors": {
-    "lvl0": ".Content h1",
-    "lvl1": ".Content h2",
-    "lvl2": ".Content h3",
-    "lvl3": ".Content h4",
-    "lvl4": ".Content h5",
-    "lvl5": ".Content h6",
+    "lvl0": {
+      "selector": "a.Nav__Link",
```
shortcuts, line 9: This will select all links, not only the active one

shortcuts (commented):
Looking at your screenshot, it seems the issue is not the `lvl0` but an anchor link out of the heading tag, `lvl1` is empty but an `anchor` is found so that's why you get an empty record

configs/transferwise.json
dozortsev: fixed https://github.com/algolia/docsearch-configs/pull/5031/commits/735f5905f2eb3e79d4d3e767b354fd6fd399ec54

dozortsev:
Hey @shortcuts 

thanks for review!

I'm not sure I understand what do you mean, do you mind maybe explain via screenshot of DOM from our [website](https://transferwise.github.io/neptune-web/about/Home)? :)

> ..it seems the issue is not the `lvl0` but an anchor link out of the heading tag, `lvl1` is empty but an `anchor` is found so that's why you get an empty record

In anycase I'm okey to try current changes as it's so feel free to merge this PR.

shortcuts:
> I'm not sure I understand what do you mean, do you mind maybe explain via screenshot of DOM from our website? :)

Example on https://transferwise.github.io/neptune-web/about/Setup#Setup

You have an empty record when searching for `setup`, this is due to the `h1` having an `id` but not being wrapped in an anchor link, our scraper consider it as an `anchor` but can't find a link for it. The reason it does not occur on other heading, is because `h2`, etc. are wrapped in a `a`.

You can solve it by either wrapping it in an `a` or removing the `id` from it.

PR #5038, opened 2022-01-16
[hanabi-conventions] Ignore convention reasons

DarthGandalf:
@Zamiell 

# Pull request motivation(s)
Enables submission of https://github.com/hanabi/hanabi.github.io/pull/820 because we don't want this new page to appear in search

The PR linked is not merged yet, so the new page doesn't appear in search. But if it were submitted, they would.

The new page shouldn't appear in the search.

### DIFF
--- a/configs/hanabi-conventions.json
+++ b/configs/hanabi-conventions.json
@@ -7,7 +7,9 @@
     "https://hanabi.github.io/sitemap.xml"
   ],
   "sitemap_alternate_links": true,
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey there, this index has been migrated to our new infra (read more here: https://docsearch.algolia.com/docs/migrating-from-legacy)

Changes should be done via our web editor, which also allow you to trigger crawls, please check your emails to see your application details!

Have a great day

PR #5064, opened 2022-01-28
Adding pipedream docs v2 index

ctrlaltdylan:
# Pull request motivation(s)

We're adding a new version of our docs 🙌 

This new version is running a brand new site with a new subpath. We'd like to make a separate index much like the Bootstrap layout.

We have one index currently `pipedream`. 

We'd like to add a new separate index called `pipedream-v2`

None, thanks for providing this service, it's great for us and our users!

### DIFF
--- a/configs/pipedream-v2.json
+++ b/configs/pipedream-v2.json
@@ -1,0 +1,41 @@
+{
+  "index_name": "pipedream-v2",
+  "start_urls": [
+    "https://pipedream.com/docs-v2/"
+  ],
+  "stop_urls": [],
+  "selectors": {
+    "lvl0": {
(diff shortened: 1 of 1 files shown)


shortcuts:
Hey, this index has been migrated to [our new infra](https://docsearch.algolia.com/docs/migrating-from-legacy), changes can now be done directly from [our web interface](https://crawler.algolia.com/) :D

Please check your emails for detailed informations regarding your new application.

Note: versions can be handled from the same config file, by adding a field to retrieve it in the `recordExtractor` https://docsearch.algolia.com/docs/record-extractor/#with-custom-variables

